PR #398, opened 2014-08-17
Feature/beets interface

RazielleS:
As stated a few weeks before in the chat, I worked on a interface between Ampache and beets.
New Catalogs:
Beets: This is the local one. Works if on the same machine. The web Server must have access to the Music on the file system and the beets Database.

Beetsremote: Uses the API from the web plugin from beets. Should work with remote machines also, but I couldn't test this. The Music will be streamed through ampache, the Listener may not have access on the beets Web Server.

Any comments and improvements are highly appreciated :-)

Edit says: Would be nice if someone could test the remote part with ogg and mp3 because I have none of these in my Library.

### DIFF
--- a/modules/Beets/Catalog.php
+++ b/modules/Beets/Catalog.php
@@ -33,8 +33,8 @@ use Song;
  *
  * @author raziel
  */
-abstract class Catalog extends \Catalog {
-
+abstract class Catalog extends \Catalog
+{
     /**
      * Added Songs counter
      * @var integer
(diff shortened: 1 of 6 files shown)


Afterster:
@Razrael I don't use your beets work so I don't know if it is still working with latest Beets version?
Someone reported an issue on Google Groups, see https://groups.google.com/forum/#!topic/ampache/XT_4BPfMc8k

Thanks.

PR #476, opened 2014-09-23
A very first implementation of UPnP localplay module

SeregaPru:
Can play and stop selected song, shows song's metainfo.
ControlPoint URL should be added as instance address (for example http://192.168.1.106:49154/AVTransport/ctrl)

Coming soon: playlist management, next/prev songs selection.


SeregaPru:
Hello, Afterster.

Sorry for long silence.

Now I'm almost ready to commit full implementation of UPNP play.
May commit joust now..
But... there is still one question, for fully fishing the work.

Now I can play only selected song in playlist, 
but I don't understand how to start play next song in playlist, when previous finished.
Which event should be triggered? Who should trigger it?

How it works for another localplay implementations?
Could you point me where to look.

Thank you.

SeregaPru:
Hello, Afterster.

Sorry for long silence.

Now I'm almost ready to commit full implementation of UPNP play.
May commit just now..
But... there is still one question, for fully fishing the work.

Now I can play only selected song in playlist, 
but I don't understand how to start play next song in playlist, when previous finished.
Which event should be triggered? Who should trigger it?

How it works for another localplay implementations?
Could you point me where to look.

Thank you.

Parfenov Sergey

Afterster:
Great !
About your playlist question: with others localplay modules the playlist is sent to the player which handle the move from current song to the next one automatically. This is not an event triggered by Ampache when the song ended.

SeregaPru:
Hello.

I'm still trying to make something for auto playing next song in list, when prefious finished.

First idea was to use UPNP subsribtion, but it works strange for me, and works different in different renderers. Some renderers even doesn't support it.

Then I've looked how it's done for example in FooBar2000.
When it plays to  UPNP renderer, it sends periodically (every second) request - GetTransportState, and processed response state.
Looks good.
But I need your help and advice - how it can be done in Ampache.
How can I make some background, or periodically called task for checking renderer's transport state.

Now I have such idea:
To make special script, launched from command line, for example.
It will:
* periodically send UPD notify packets 
* (maybe) - listen incoming UDP discover packets and send requests
* and control, if something is playing to UPNP renderer, check its transport state and when song finished, send next song from list to renderer.

But not good is that user should lauch this script to play to UPNP.

Do you have any ideas?

Sergey.

Afterster:
Any reason to close this whereas it is not merged?
My opinion is that the script could be called from php to run in background. It will requires additional setup on server access rights but this can be a prerequisite as it is for channels to work. What do you think?

SeregaPru:
I havу greatly reworked all classes and decide to close this request and open the new one.
Will do it soon.
Now it's worked and played well, still one issue with stop playing, now when stop it begin to play next song.

PR #643, opened 2015-01-02
add placeholder text to search box

brownl:
Another very minor aesthetic change, but I think it makes the purpose of the field a bit more obvious to new users.
Before:
Before
After:
After


SUTJael:
Yes thanks but you need to think about other languages: ``<?php echo T_('Search...'); ?>``

brownl:
oh, sorry, I completely forgot... I'll sort that out - what happens if the string hasn't been translated? does it just echo nothing?

brownl:
I sorted it

Afterster:
Thanks for your PR, merged ;)
If the string hasn't been translated, it echo the default string (=English).

PR #649, opened 2015-01-03
loop over directories the "correct" way

brownl:
see http://php.net/manual/en/function.readdir.php - specifically:

    /* This is the correct way to loop over the directory. */
    while (false !== ($entry = readdir($handle))) {
        echo "$entry\n";
    }

    /* This is the WRONG way to loop over the directory. */
    while ($entry = readdir($handle)) {
        echo "$entry\n";
    }

### DIFF
--- a/lib/class/art.class.php
+++ b/lib/class/art.class.php
@@ -1031,7 +1031,7 @@ class Art extends database_object
             $processed[$dir] = true;
 
             // Recurse through this dir and create the files array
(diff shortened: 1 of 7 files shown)


Afterster:
Thanks for this PR. You should also suggest the fix on /modules/Auth/OpenID/FileStore.php to https://github.com/openid/php-openid project.

brownl:
Thanks, I'll make a PR there.

Issue #469, opened 2014-09-19
Multiple genres not read even after catalog wipe


anettles:
I had problems with multiple genres in the past & have mostly removed them.  But this says it was fixed:

https://github.com/ampache/ampache/issues/171

So here's what I tried - In MP3tag I went to extended tags and edited GENRE to be "genre1\\genre2" which is supposed to separate them with a null.  I saved it in ID3v2.3 UTF-16.  It appears correctly in Picard as far as I can tell.  I tried updating the catalog and the 2nd tag doesn't show either on the album's genre or in the tag cloud.  I tried deleting the catalog and re-adding it but it still doesn't show in either place.

Related to this - would it be possible to have the tags update with a catalog update instead of having to wipe & start over?

I added a new catalog with a single album as a test - tagged it with 2 genres the same way.  It still only shows up with the first genre.  Here's the log from that:

2014-09-19 20:14:34 [User] (ajax.server.php) -> Called for page: {index} 
2014-09-19 20:14:34 [User] (session) -> Session created: 2239b224641fa77b8f2b74fe9dc32a41 
2014-09-19 20:14:39 [User] (session) -> 61tpejkpjoi7p8cj1h94j0kdo2 has been extended to Fri, 19 Sep 2014 22:14:39 +0000 extension length 1411164879 
2014-09-19 20:14:39 [User] (session) -> Session created: fd9486999a6097d3f0068404f25d0b39 
2014-09-19 20:14:39 [User] (autoload) -> 'OAuthException' not found! 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin 7digital is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Amazon is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Bit.ly is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin ChartLyrics is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Flickr does not support get_avatar_url, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Gravatar is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Growl is not installed, skipping 
2014-09-19 20:14:39 [User] (Plugins) -> Plugin Headphones is not installed, skipping 
[...]

PR #655, opened 2015-01-05
[FEATURE] Create and update splitted tags

Razrael:
Hey there

I discovered some odds with the tagging system, so I quickly fixed the most important (for me) with this Pull Request.

First: Some tagging softwares (like beets' LastGenre plugin) can add multiple genres to a song. But they are written in one single tag divided by a separator.
So I divide them if there is only one genre tag by some common separators ( , and ; )

Second: There is apparently nothing that updates the song genre when the catalog gets updated.
At this time, I have only updated the tags from a song.

This could be a first step for #469, I have not tested it with real multiple tags though.

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -1421,6 +1421,11 @@ abstract class Catalog extends database_object
         $new_song->replaygain_track_peak = floatval($results['replaygain_track_peak']);
         $new_song->replaygain_album_gain = floatval($results['replaygain_album_gain']);
         $new_song->replaygain_album_peak = floatval($results['replaygain_album_peak']);
+        $tags            = Tag::get_object_tags('song', $song->id);
+        foreach($tags as $tag) {
+            $song->tags[] = $tag['name'];
+        }
+        $new_song->tags        = $results['genre'];
         $artist                = $results['artist'];
         $artist_mbid           = $results['mb_artistid'];
         $albumartist           = $results['albumartist'] ?: $results['band'];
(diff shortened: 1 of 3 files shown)


Razrael:
Just a short update
Probably I can add the update for artist and album tags this evening, I had a little epiphany.

Can someone tell me what this checks are checking exactly? Only my part of code or the whole file?
I am not sure if it fails on my changes or if something different in the files is bad.

Afterster:
Thank you for this PR @Razrael.
I believe changes in lib/class/vainfo.class.php should be removed as it will be handled with #657 in a more flexible way. But other changes are indeed good fixes.
I will wait for lib/class/vainfo.class.php revert and artist/album updates too. Thank you! :smile: 

Travis check php syntax and codestyle. The check failed on your changes, see https://travis-ci.org/ampache/ampache/jobs/46007397 output.

Razrael:
weird...
I ran 'php-cs-fixer.phar fix -v --fixers=indentation,linefeed,trailing_spaces,short_tag,braces,controls_spaces,eof_ending,visibility catalog.class.php' before commit 125d203. But this thing is still complaining about this file :-(

brownl:
@Razrael your code failed on `lib/class/catalog.class.php (trailing_spaces)` - check, do you have trailing spaces?

Razrael:
Looks better now, thanks.
The test is done with a 'scripts/tests/codestyle.sh'.
Can I get this thing somewhere? Would be nice because the php-cs-fixer aparently fails with some things.

Afterster:
This script is on the repository ;). It also calls php-cs-fixer behind.
But I toke a closer look on your changes and adding `if (true ||` will be problematic. With your changes all files will be updated on catalog update, this doesn't looks like a clean solution.

Razrael:
You are of course right. This statements were never intended to reach the repository, I just don't wanted to rewrite every time my song-tags after testing and forgot about these two...
Thanks for your good checking, for eyes are always better than two :-)

SUTJael:
Any chance to use separators described into #657?

Afterster:
@SUTJael this is also already merged, quicker to directly commit then ;). You're talking about beets tag processing right?

PR #657, opened 2015-01-06
Added support for additional ID3v2 genre delimiters.

lotan:
This is the first stab at adding support for additional ID3v2 genre delimiters.

I've seen in other places in the code that config checking (in this case using isset and is_array) is not done - I don't know if this is done somewhere else. So this code could be slimmed down a bit.

I've used the following files to test the code:
http://www.filedropper.com/additionalid3v2genredelimitertest

Unfortunately, I don't know how representative that test is.

I'd really appreciate, if someone with that issue could test it.

### DIFF
--- a/lib/class/vainfo.class.php
+++ b/lib/class/vainfo.class.php
@@ -804,7 +804,20 @@ class vainfo
 
             switch ($tag) {
                 case 'genre':
-                    // Pass the array through
+                    // read additional id3v2 delimiters from config
+                    $delimiters = (array) AmpConfig::get('additional_id3v2_genre_delimiters');
+                    if (isset($data) && is_array($data) &&
+                        isset($delimiters) && is_array($delimiters)) {
+                        // if there is only one element in the array,
+                        // iterate through the delimiters
+                        // until the genre string splits into several elements
+                        $i = 0;
+                        while (count($data) === 1 && $i < count($delimiters)) {
+                            $data = explode($delimiters[$i], $data[0]);
+                            $i++;
+                        }
+                    }
+                    // assign genre array
                     $parsed['genre'] = $data;
                 break;
                 case 'part_of_a_set':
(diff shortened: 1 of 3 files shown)


Afterster:
Maybe **,** and **|** should also be added in default delimiter?

SUTJael:
We should also upgrade the config file version no?

Afterster:
It is already merged, and I updated the config version.

lib/class/vainfo.class.php
```diff
@@ -804,7 +804,20 @@ private function _cleanup_id3v2($tags)
...
-                    // Pass the array through
+                    // read additional id3v2 delimiters from config
+                    $delimiters = (array) AmpConfig::get('additional_id3v2_genre_delimiters');
+                    if (isset($data) && is_array($data) &&
+                        isset($delimiters) && is_array($delimiters)) {
+                        // if there is only one element in the array,
+                        // iterate through the delimiters
+                        // until the genre string splits into several elements
+                        $i = 0;
+                        while (count($data) === 1 && $i < count($delimiters)) {
```
Razrael: Maybe we could replace this whole while thing with a simple preg_split() after checking $data?
There we could also filter the whitespace from the genre elements (multiple genre in beets are like this: "genre1, genre2, ...")
This would also fix the problem that it cannot handle comma. Of course then the delimiters should not be parsed to an array.

lotan:
@Razrael: That would indeed solve the comma problem and probably be a bit quicker. Do you want to conribute the code?

Razrael:
Sure, probably I can do it this evening.

Razrael:
@Afterster Can I change the content of the setting in the config file with an update?
It would need another syntax and also include the comma.
I did not used this yet so I'm not used to the magic.

lotan:
@Razrael: You could use ```catalog_file_pattern``` as a template. That would allow multi-character delimiters like ```//```:
https://github.com/ampache/ampache/blob/develop/config/ampache.cfg.php.dist#L133
https://github.com/ampache/ampache/blob/develop/lib/class/catalog.class.php#L319

Issue #656, opened 2015-01-05
Social feature broken in recent development tree


florihupf:
Hi there,

I am using latest and greatest of the development tree.  I love the "social feature".  However, as of recently when I click the comment bubble the text entry field has gone missing and I can't post any comments on a song anymore.

brownl:
could you give a rough estimation of when this became an issue?

florihupf:
It was working in the last officially released version (I think 3.7.0).
When I updated to the dev version about a month ago it wasn't anymore.

Thank you for taking a peek!

~ Florian

brownl:
okay, the issue is in /modules/jquery-mediaTable/jquery.mediaTable.css, line 69:

    (68) .activeMediaTable th, .activeMediaTable td {
	(69)    display: none;
    (70) }

With that rule disabled, the input box renders fine. does anyone know why it is here?

brownl:
Created Pull Request #659

florihupf:
Thank you!

PR #659, opened 2015-01-06
Fix shoutbox not displaying (#656)

brownl:
This addresses issue #656.

jQuery-mediaTables (for some reason) was stopping the shoutbox from showing. It  turns out that mediaTables makes no difference to the shoutbox, so the simple solution is not to use mediaTables for it.


Afterster:
Good catch, thanks!

Issue #654, opened 2015-01-05
Tags show up as single letters instead of words.


florihupf:
Hi there,

I think this used to work and now is broken.   Some of my tagged mp3s have tags in them.

For example one has: "Dance & Noir".

However ampache reads them as:

'&', 'D', 'N', 'a', 'c', 'e', 'i', 'n', 'o', 'r'.

Is this something I do wrong on my end when tagging? Or is this a problem of the ID3 library?

Thank you!

SUTJael:
Any chance to retrieve the raw data to see how your tag is encoded? Which software do you use? I think it's related to recent update on tag processing.

florihupf:
I am using TagScanner 5.1 on Windows.

It used to work great in version 3.7.0 of Ampache. So I agree, it must be a
recent regression.

Is there another way I can help?

~ Florian

SUTJael:
Would be great to give us a sample file :)

Razrael:
@florihupf I will look into this issue in #662 
Would be great if you could give us a file where the problem occurs, so we can reproduce and fix this.

Afterster:
Must be an encoding issue again. Could be related to #628 and https://github.com/JamesHeinrich/getID3/pull/39 pull request.

Razrael:
Possible. If we can get a file with an error we can simply do a regresson test.

florihupf:
Will send a file today.

florihupf:
Here is a file: http://www.nid-nad-nud.org/seafile/f/9de42c3d60/

The genre is encoded wrongly.

florihupf:
The genre comes out as:

/, A, b, c, E, i, l, M, n, p

Where is should be:

Abstract/Minimal/Experimental

Razrael:
Thanks for the file.
I tested this on my dev server. On latest development branch I got no weird behavior, though the tags were parsed as one single string, including the delimiter.
When I test it on the branch from Pullrequest #662 the tags are splitted properly.
I assume you checked the tags in the tag cloud and on the song page in Ampache? Just to avoid misunderstandings :-)

I would suggest you to get the latest development branch again and update your music library. Support for updating tags is included since a few days. Maybe a other developer can double check this, could possibly 

After that we could continue with your Ampache settings file, just to be sure.

florihupf:
Ok, I think I am on latest greatest, but I will update just to make sure
and fully rebuild my catalog.

Yes, the broken output is visible in the tag cloud and in the song page in
[...]

PR #662, opened 2015-01-07
Feature/improve genre splitting

Razrael:
As discussed in a earlier pullrequest I changed the genre splitting to a regex-based function.
I included it at all places where the cleanup happens, not only id3v2, so I renamed also the config tag.

@lotan I think it is the same way you mentioned, the catalog patterns are also regex, but they have not a complicated pattern so they are easier to understand.

Side note: I always have problems to determine if my code has a valid syntax because I usually use another syntax at work. With php-cs-fixer I get errors on a lot of files, so I just fix the whole file...But this leads to ugly commits. Is it possible to do a code cleanup to psr2 for most of the project?

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -295,10 +295,11 @@ metadata_order = "getID3,filename"
 ; DEFAULT: filename getID3
 metadata_order_video = "filename,getID3"
 
-; Some taggers use delimiters other than \0 for ID3v2 fields
+; Some taggers use delimiters other than \0 for fields
 ; This list specifies possible delimiters additional to \0
+; This setting takes a regex pattern.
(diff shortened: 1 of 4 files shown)


SUTJael:
Thanks, good work! Before merging, could you take a look to issue #654? We should fix that issue at the same time.
About the code cleanup, we'll do that ;)

Razrael:
Yes I can. But it would be great if we could get a sample file for this one.
I cannot imagine how this could happen, and it never happend to me, so I need some stuff to work with.

Razrael:
The bug has solved itselve, so we can probably continue with this one.
I already raised the config version number. Anything else to do?

Afterster:
Looks fine to me, good work as usual. Thanks again ;)

Issue #280, opened 2014-06-20
Add support for arbitrary information, text.


Xenakis:
User may want to add arbitrary information about the album/artist/etc.

Source text file may include artist info, i.e., bio; and may contain album info, i.e., credits; perhaps more.

Text file should also be added to DB for search.

Afterster:
I will move this issue to 3.8.1 so 3.8.0 will not have any improvement and we can focus on test and bug fix for a release. Let me know if it is a problem for you ;)

Razrael:
For me it is no problem, currently I have almost no time to work on it anyway...

Afterster:
Razrael work was merged few weeks ago. I'm going to close this then.

PR #664, opened 2015-01-10
#280 Add custom Metadata

Razrael:
Finally I managed to create the things needed for custom metadata.

So now you can just update your library to get all metadata.
- The metadata is shown in the song view
- Metadata can be changed over usual update song function
- Theoretically the changed Metadata can be written back to files if configured. I couldn't test this because I cannot write anything back to files. Have to track this bug separately.
- Metadata can be searched through Song search

Technical details
- I introduced Database repository and Model, so the DB can be used more abstract
- changed the autoloader to support namespaces outside of modules
- There are some stubs in the repository, where I can implement some magic to improve speed.

I did not validate and fix some of the files to not messing up the Pullrequest, so you can actually see what changed. I can do this later, unless you want to do this yourself in the whole repository, then I can rebase this one.

Questions and suggestions are welcome :-)


Afterster:
That's a lot of work, thanks!
I reviewed your code it looks great but I'm not completely able to test it:
 - Why did you change NetBeans PHP version to 5.4? Ampache should be PHP 5.3 compatible for now. PHP 5.3 and Apache 2.2 compatibility will be drooped one day, but this is to early for that (Debian just moved recently to PHP 5.4 in stable version).
 - You introduced a PHP 5.6 only feature (I believe it's why you bumped to PHP 5.4 ; unfortunately it's not even a PHP 5.4 feature but a PHP 5.6) with const array declaration `const aliases = array(` on song.class.php.
 - I'm also asking myself what would be the behavior for users with lot of metadata junk. Is there anyway they can avoid this additional metadata field creation in database if they don't want to have it?

Razrael:
Hey, thanks for your comment.

**PHP 5.4**
I used some 5.4 features. Some of them because they are handier, some, like traits to making the structure not even messier as it is already.
I looked trough all major shared hostings I know and found (at least in switzerland), and all of them offer at least PHP 5.4, and some offer also the possibility to use 5.5 or 5.6.
The big hostings don't support PHP 5.3 anymore. There are also some widely used CMS that have dropped 5.3 in their beta (and as far as I know we are working on the beta for Ampache 3.8 :-) )
For example Drupal 8 goes with PHP 5.4 and typo3 has only suppport for PHP 5.5 and upwards.
I think the meaning of a beta is to **go a step forward**. People that don't have the possibility to use PHP 5.4 can still use Ampache 3.7. What do you think?

**Feature PHP 5.6**
Using a PHP 5.6 only feature was not intended. I usually stick to the feature checking from Netbeans so I can focus on coding only and it didn't throw a warning at me...Or maybe I missed it somehow. Will change that.
Looks like I should compile some old PHP versions on my machine so I can test the syntax.

**Disabling the Metadata Feature**
As of now there is no way to disable this feature. Why should anyone have junk metadata?
Sure I can add an additional line in the config file, but I'm not sure why anyone would disable this. They could even use it to cleanup something.

I will also do some additions to keep the performance up. There are too many sql querys for inserting.
Additional commits will hit the PullRequest probably today or tomorrow.

Afterster:
Dropping PHP 5.3 for Ampache 3.8 could be envisaged indeed, moving forward as you said. @SUTJael what do you think?

About disabling the feature, I know several Ampache users with bad tagged music files and they are happy with it because they don't see the mess behind neither they want to spend time on cleanup. Just thinking but what about an option to 'limit' imported metadata to a specific list?

Afterster:
Just a reminder, a warning from another issue argument in favor to drop 5.3 compatibility.
See http://php.net/manual/fr/function.session-set-save-handler.php, lib/class/session.class.php should be updated to this more object oriented approach since 5.4.

SUTJael:
Dropping PHP 5.3 is also good to me.

About the metadata feature I think there's two important things to check before choosing:
1. Is there any performance fall due to multiples insertions?
2. What about the database growing size?

In both cases, if the anwser is that we loose some performances, it would be great to be able to disable that feature. Otherwise, I don't see any trouble by keeping all metadata and allows the user to cleanup (or not) all tags.

Razrael:
Beeing able to disable the feature is fine for me, I can add this soon.
Of course there is a performance dropdown while inserting and probably also while updating. Of course only if there is a lot of additional data.
I will try to get the performance up especially for inserting, because there is currently a lot of potential.

I was thinking since the beginning about the possibility to hide some tags: The admin would be able to configure what tags are visible (and maybe dont import them on new songs). I prepared some stuff for this addition (e.g. the field in mysql is already there), but there is no tool to do this yet. I may update this here if you think this would be a goood solution.

@SUTJael what do you mean with "Is there any performance fall due to multiples insertions?"
What multiple insertions?

For the moment it is hard for me to test the performance, my catalog has a bit less than 5000 songs and ampache is running on a cubieboard 2, so not that much performance. But it's all working without lot of waiting.

Afterster:
I believe @SUTJael was thinking about 1 tag = 1 sql insert request. I don't think your changes has major performances issues, the only missing thing that I see like I said is the admin part to 'hide' or 'ignore' some tags. Let me know when you update this PR and I will test more.
Great work anyway ;)

Razrael:
This is exactly what it currently does and I want to optimize this anyway, so there will be less queries.
I will add the admin part later. Currently I'm low on time, but I hope I can provide this soon.

Afterster:
This is still highly interesting.
3.8.0 was released so I'm wondering if we could integrate your work on next release @Razrael? Thanks.

Razrael:
Hey

I have this still on my list. But first I have to repair my Server,  
the whole system crashed and I have to set it up again.
After that I can care about implementing the admin controlls.

I keep write an update as soon as I get somewhere :-)

ciroiriarte:
Hi!, any update on this?. I have an use case that would allow to stress the solution and validate scalability (around 30k files generated daily).

I'm also wondering what file formats will be supported. For example, my understanding is that MP3 don't have arbitrary tags, on my tests I'm adding info grouped as "comment" but OGG and FLAC provide separated custom fields, which I assume would be easier to parse.

Really hoping to see this feature released.

ciroiriarte:
Sorry, *MP3 doesn't have arbitrary tags*. Although only tested encoding & tagging with lame.

Razrael:
Hi there

Finally I found some time to update the PR with the ability to deactivate the custom metadata functions.

@ciroiriarte It should support at least FLAC and OGG. I'm not sure about the abilities of MP3 tough. It just uses an already built in library, so I did not code the part of parsing the metadata out of files myself.

I'm not sure about scalability, I used it on a fairly slow server, so I can't compare it.
But there are possibilities to improve the performance of the whole thing. But I'd like to add this in a separate PR because this one is quite hard to maintain when there were updates.

Razrael:
@Afterster Could you please take another look at this?
About two weeks ago I added some changes, but you probably got not informed about it.

Afterster:
Yes I'm currently in the process of reviewing your PR, should be merged soon.
Basically your changes is really great and I'm glad you did it, thanks again.
But I noticed some changes that doesn't respect the current code standard (it is what it is but more important is to have a consistent code base) so I'm changing it. Also, your Model/Repository and namespace logic is good but I'm a bit worried that the only objects to use it will be your metadata classes, again for code consistence. I believe we will deal with it that way right now but theorically all classes should be moved to this model. I have to think about it again ^^.

Razrael:
I'm glad to hear that :-)

If you want to move to my introduced model I suggest to wait a bit. I have some improvements in my mind to get better performance. Its about persisting the Database, which will happen now every time you save the object. But I want to have a persistence class which can take care about this process so you can persist multiple objects at once. Could be quite a performance improvement for the metadata saving and possibly also other parts.

But this would be a breaking API change, so its easier when its not used that much yet.

Of course I would do that in another PR. But I think I can't do that before my vacations at the end of December.

Afterster:
Ok merged!
I will not move to your introduced model right now but whereas I merged it, it introduced a code inconsistence because of this model/repository logic. I believe we should move to a cleaner database layer like this one but this will take some time, or at worst undo this model to have the same logic everywhere. If you attempt to do modifications on this like you said (persistence, ...) it would be great. I will also hope you will be motivated to update other objects to this model so I can focus myself on other issues. And thanks again for this feature and beginning of code improvement ;)

Razrael:
In your previous message you wrote about code standard which was not met. Looks like you changed this things in a commit. But shouldn't the build system complain about the code style? I used the provided script to check and fix my mistakes. But obviously there were a few occasions where it didn't work. Can we change that?
I rely on this script because I'm used to another code style, so I make plenty of mistakes here.

About the future changes: I intend do to the persistence layer and I think I could also try to refactor some of the current code. Need to take a look first how hard it is, but I think it should be doable.

There are also tons of things I'd like to change and fix. For example the UI code is pretty much a big mess in my opinion. But I have to find some time for it...

PR #666, opened 2015-01-12
Fix Content-Type of asx

thinca:
reason: http://support.microsoft.com/kb/288102/en-gb

Windows Media Player is going to play a playlist as wmv file, and fail.


Afterster:
Indeed, thanks.

PR #674, opened 2015-01-19
add videos in playlist 

SurvivalHive:
dont sure if i´m right here - dont find forums or something else for ampache...

i have many music tracks as video but i cant add them into a playlist.
nothing happends -  not important which video file it is...

it will be awesome if you can add videos in a playlist so u can play the complete list automaticly.
at the moment it is not possible - only step by step...


Afterster:
I think you created this PR my mistake, I'm going to close it.
Please create an issue instead. I also believe you're talking about master branch (3.7.0 version)?

PR #696, opened 2015-01-29
Modified get_image_preview function to include a check that 'transcode_input' is defined

wagnered:
And added it to $command to transcode.


Afterster:
Mh indeed I missed that when adding the argument separation. Thanks.

Issue #624, opened 2014-12-15
Option to hide footer statistics


Rouzax:
I think I remember that I enabled these statistics somewere.
![image](https://cloud.githubusercontent.com/assets/4103090/5432940/35d225e4-8442-11e4-81e8-dcacdbe25e86.png)
However for the life of me I can't find it anywere in the GUI or config file.

Rouzax:
Thanks

PR #702, opened 2015-01-31
Option to show/hide footer statistics

brownl:
This addresses issue #624.

### DIFF
--- a/templates/footer.inc.php
+++ b/templates/footer.inc.php
@@ -19,10 +19,11 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  *
  */
-
+/*
 $ampache_path = dirname(__FILE__);
 $prefix = realpath($ampache_path . "/../");
 require_once $prefix . '/lib/init.php';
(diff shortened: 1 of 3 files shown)


templates/footer.inc.php
```diff
@@ -19,6 +19,11 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  *
  */
+
+$ampache_path = dirname(__FILE__);
+$prefix = realpath($ampache_path . "/../");
+require_once $prefix . '/lib/init.php';
```
Afterster: Why do you have these lines in footer.inc.php? Does this was for test and get committed by error?

brownl:
without them I get errors about AmpConfig not being defined. Or am I doing something wrong?

Afterster:
I don't think you get errors without these lines, are you sure?
Could you also clean your commits now for the PR? Several commits with revert inside the PR itself is too much noise :) Thanks.

brownl:
Okay, opened #703. Thanks.

PR #703, opened 2015-01-31
(Cleaned) add option to show/hide footer statistics

### LINKED ISSUE #624: Option to hide footer statistics

brownl:
Cleaned version of #702. This fixes #624.

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -7,7 +7,7 @@
 ; if this config file is up to date
 ; this is compared against a value hard-coded
 ; into the init script
-config_version = 25
+config_version = 26
 
 ;###################
 ; Path Vars        #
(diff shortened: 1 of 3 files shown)


Afterster:
Yep, merged thanks (you could also have cleaned your previous PR simply with some git commands fyi).
I just enabled it by default and replaced you `: endif` with braces to respect project code style.

brownl:
Thanks, sorry I'm not very experienced with git!---- On Sat, 31 Jan 2015 22:34:45 +0000  Afterster<notifications@github.com> wrote ----Yep, merged thanks (you could also have cleaned your previous PR simply with some git commands fyi).I just enabled it by default and replaced you : endif with braces to respect project code style.—Reply to this email directly or view it on GitHub.

PR #707, opened 2015-02-02
Set http status code from the curl request

nicklan:
Need to pass through the http status code from the underlying request in order to properly answer with a 206 when skipping bytes.  This needs to go here because this is where the response body starts, so setting it later won't work.

This will fix issues with gstreamer where it reported that ampache couldn't seek because of the 200 response code that was previously being returned.  I've also tested that seeking still works in VLC with this.


Afterster:
Thanks, merged.

Issue #496, opened 2014-10-07
Enable QRCode based authentication


RyanCopley:
A QR code should be created and displayed so that mobile users can utilize it to login to their Ampache server from a mobile client. My personal suggestion would be:

```
ampache://username:password@hostname
```
or (and?)
```
ampache://apiKey@hostname
```

Afterster:
Ok currently done on account details page, for API key only. I updated the wiki to make official this Ampache scheme (https://github.com/ampache/ampache/wiki/XML-API#ampache-sheme).

RyanCopley:
:+1: you rock

RyanCopley:
This might sound dumb, but how do I get to this? I did manage to find it, but I have *no* idea how I did

RyanCopley:
Nevermind... you click your username in the header :)

RyanCopley:
Hey @Afterster -- The QR code isn't working, we've tried multiple phones and multiple apps and none of them are picking up the QR codes at all

RyanCopley:
I set the width and height to 256 and it seems to work-- very odd.

Afterster:
Indeed 256 is better for recognition and indeed it looks like the generated qrcode has bad recognition rate. I'm using https://github.com/jeromeetienne/jquery-qrcode/ here, it doesn't have any known issue like that so maybe it is a misconfiguration...

RyanCopley:
I just opened a PR for a larger image size 

https://github.com/ampache/ampache/pull/721

RyanCopley:
Afterster I just did some more research and I found out the SIZE is not the issue, but the background color behind it! When I select over the area, it turns orange, and it scans 100% of the time. What do you think the best solution for this would be, since the theme is dark?

PR #721, opened 2015-02-08
Increased QR code size (Relates to #496)

RyanCopley:


Afterster:
Not really user friendly but I agree, outside this size the match rate is too low.

PR #745, opened 2015-02-25
Update any mysql password placeholders

oppai:
Update any mysql password placeholders.
I did not notice that mysql password is required.


Afterster:
Yep merged, thanks.

PR #759, opened 2015-03-04
Fix "Various" when multiples artists on album

ronane:
Use f_album_artist_name when it's know instead of "Various" when there are multiples artists (a featuring for example)

### DIFF
--- a/lib/class/album.class.php
+++ b/lib/class/album.class.php
@@ -620,7 +620,7 @@ class Album extends database_object implements library_item
             $this->f_artist = $artist;
         } else {
             $this->f_artist_link = "<span title=\"$this->artist_count " . T_('Artists') . "\">" . T_('Various') . "</span>";
-            $this->f_artist = T_('Various');
+            $this->f_artist = ($this->f_album_artist_name) ? $this->f_album_artist_name : T_('Various');
             $this->f_artist_name =  $this->f_artist;
         }
 


Afterster:
I'm not sure we should, there is already a distinction between artists and album artist, this would override the information. Maybe album artist should be displayed by default on some template view instead?

ronane:
well you're right, i've to admit i haven't well studied the model and
searched just a quick fix for home. It would probably better to modify the
view.
Clearly i've been a bit discouraged when i saw $album_artist,
$artist_name,  $f_artist_name, $f_artist, $album_artist_name,
$f_album_artist_name or $name, $f_name, $full_name, $f_title in the model.
Is there a description for each attribute anywhere ? because some seems
really close by denomination to me ^^


2015-03-04 22:22 GMT+01:00 Afterster <notifications@github.com>:

> I'm not sure we should, there is already a distinction between artists and
> album artist, this would override the information. Maybe album artist
> should be displayed by default on some template view instead?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ampache/ampache/pull/759#issuecomment-77251816>.
>



-- 
Ronane

Afterster:
By convention in Ampache code, `$f_*` are formated values in attempt to be displayed.

In album:
`artist_name` is the artist name without prefix (a, the, das, die, le, la, ...).
`f_artist_name` is the display artist name with prefix (or Various)
`f_artist` = `f_artist_name`
`name` is the album name without prefix
`f_name` = `full_name` = `f_title` is the album name with prefix

There is some old variables from previous maintainers that would require some cleanup...

Afterster:
This PR doesn't seems valid anymore according to the discussion. Can I close it?

ronane:
I'm just not motivated enougth to make this clean for now.
You can close it. I'll try to make a new pull request when i do something
clean ^^

Afterster:
Haha ok. Don't hesitate.

Issue #768, opened 2015-03-13
Right panel overlay menu displaced when window is scrolled


jpbaril:
With 3.8-beta1, on Firefox 37 beta3 on Mac OS X 10.9.5, when the browser window is scrolled, in the right panel, the overlay menu is displaced far below. This makes it impossible to select options in that menu.
![capture d ecran 2015-03-12 a 20 40 44](https://cloud.githubusercontent.com/assets/5779633/6631274/9533e65c-c8f9-11e4-8a2e-8266e160e81a.png)

SUTJael:
The class "fixedrightbarsubmenu" which is applied to the rightbar menu, when fixed options are disabled, should already handle that. Could you try to clear your cache, refresh and then see if the submenu has the ".fixedrightbarsubmenu" class? Working great on my side.

SUTJael:
@thegeekrv, @jpbaril  BTW, I reverted the commit to the initial value, you should update before trying again :)

jpbaril:
You are right, it seems cache was not properly emptied because after a few more click on "clear cache" and refreshes, the bug eventually disappeared.

I can only speak for what's in 3.8 beta-1, but that issue still sometimes happens when after scrolling the window down, the window is scrolled back to the top. The rightbar menu is then sometimes not positioned correctly and the submenu is again too far below.

![capture d ecran 2015-03-23 a 13 03 17](https://cloud.githubusercontent.com/assets/5779633/6785858/cdda5f64-d15e-11e4-91d4-c07f5f324375.png)

The quick fix is to simply slightly scroll down again and the menu will get positioned right again. Yet, that happens sometimes.

Hopefully this time it will not be a waste of your time.

SUTJael:
Ok closing this as that was a cache issue. Thanks.

PR #769, opened 2015-03-14
Solve #768

thegeekrv:
Solve bug #768

### DIFF
--- a/themes/reborn/templates/default.css
+++ b/themes/reborn/templates/default.css
@@ -739,7 +739,7 @@ input[type=button]:focus:active, input[type=submit]:focus:active {
 #rightbar .submenu {
     display: none;
     position: fixed;
-    top: 120px;
+    top: 0px;
     right: 15px;
     border: 2px solid;
     width: 210px;
--- a/themes/reborn/templates/default.css
+++ b/themes/reborn/templates/default.css
@@ -739,7 +739,6 @@ input[type=button]:focus:active, input[type=submit]:focus:active {
 #rightbar .submenu {
     display: none;
     position: fixed;
-    top: 0px;
     right: 15px;
     border: 2px solid;
     width: 210px;


jpbaril:
I perhaps did not test the fix correctly (I manually changed the css value with firebug), but that did not fix it. In my view, the bug relies in the fact that 'position: fixed' is used and that the top position is not modified in the same way that the 'rightbar' div top value is modified when scrolling the browser window.

thegeekrv:
I see where is the problem, I ll fix it maybe this afternoon, or later :)

thegeekrv:
Ok, now it's ok :)

jpbaril:
Seems OK.
Often, when trying to move cursor over the submenu, the submenu disappear like if there was a gap between the menu and submenu, but I think it's another bug.

thegeekrv:
I think it's because of the border...

Afterster:
Looks fine to me too, thanks.
@SUTJael, as you did this css part, could you give some hints on your choices to @jpbaril? Thanks!

thegeekrv:
You are welcome :)

jpbaril:
@thegeekrv I tried removing the border and still got the problem.

I took the time to recheck the fix and as I mentioned, I think that even if visually it seems like the bug is fixed, I think that unfortunately it is not completely solved.

I used Firebug to change the background color of the submenu (for sake of visual clarity while testing) and indeed, there is a gap between the menu items and the submenu, which is obviously the source of the problem. Now, why it does not always happen is beyond me.
capture d ecran 2015-03-21 a 18 38 28

In the end, this bug will need more work. Unfortunately my CSS is very far so I cannot say at this moment what should be changed.

jpbaril:
I think I found it.

I think the gap is the "line-height" of the <li> in the menu (rb_action <ul>).

PR #779, opened 2015-03-21
Transcoding not finishing process

nan4k7:


Afterster:
Ok thanks, now it's better as a PR and simpler to review your changes :)
I have few questions now:
 - Why do you added a prefix to run in background? What was wrong with the previous solution
 - The manual process kill stuff you added shouldn't be done only on Windows? On linux we could use the regular `proc_terminate` function isn't it?

nan4k7:
About the things you asked:
- The prefix is because of what it says here (http://php.net/manual/es/function.proc-get-status.php comment by Mark Seecof). When you don't use it, proceses are opened by sh and that leaves a child.

- The kill function is because of something similar (http://php.net/manual/es/function.proc-terminate.php) if you use proc_terminate there are chances of leaving a child (comments by  jerhee at ucsd dot edu and  v dot denegin at yahoo dot com)

Afterster:
Interesting, thanks for the links (I will add them to source code comments when merging) and thanks for your work to resolve your issue.
I'm not a big fan to always send kill -9, maybe it could be an option that you enable depending your environment/transcode utility, what do you think?
Could some other people watching Ampache repo also try this fix and give feedbacks before the merge? Thanks!

Afterster:
After several feedbacks (not here) and testing your fix really looks the best way to process.
Transcoding is an important feature, really helpful. I merged it, thanks again for your time and this fix.

PR #785, opened 2015-03-25
Upnp localplay implementation and other

SeregaPru:
1) Upnp localplay implementation
2) Fix: special chars in songs names and tags correct support


Afterster:
I was waiting for this. Thanks a lot @SeregaPru, you did a great work here :)
Merged, thanks again !

PR #816, opened 2015-04-22
Converting Boolean compare to identical operator

e3betht:
Comparisons against Booleans should use the identical operator to compare to avoid any inconsistencies or false positives. I found this issue using Code Climate to evaluate the code, so I added the Code Climate badge to the README.


Afterster:
We are already using Scrutinizer, Codacy and SensioLabsInsight (I will add missing badges too), why not this one too... But after that we should really consider adding a new one.
I see you are working at Code Climate so this pull request looks a bit promotional for me as you only fixed a really little issue compared to the thousands of changes in the report...
Anyway, let's give a try ; we can still report if we are not satisfied. And thanks for this short fix ;).

PR #826, opened 2015-04-26
Fixing Boolean comparison operators and removing closing PHP tags.

e3betht:
I found some more Boolean comparisons that weren't using the identical operator, and I also saw that some PHP-only files had a closing PHP tag as their last non-whitespace line, so I removed those as well.


Afterster:
Thank you for this new PR @e3betht.
It seems you also updated external dependencies in /modules (using Composer is a todo that we really should take care of one day). Could you change your PR to not update files under /modules folder?
Thanks!

e3betht:
So, as you can see, I'm still learning git, but I think I got the changes all reverted except for the two non-module files. Let me know if you need anything else changed on this.

Afterster:
Much better thanks. Could I be fussy? This is a lot of commit for only two line changes, it's good practice to clean the commit history when you're doing a pull request, could you merge your 6 commits into one? This is good opportunity for learning as you said ;). Thanks!

RyanCopley:
@e3betht This can be accomplished via `git rebase -i`, make sure you read up on it beforehand. A good google search term is `git squash commits`

SUTJael:
@e3betht, any chance to merge your commits to merge your PR? :)

Afterster:
Because of the history noise for just two lines changes, I did the modification in commit https://github.com/ampache/ampache/commit/2424c76ab71caadd34e0072105337b938dc4c2d7 outside this PR. Thanks anyway!

PR #830, opened 2015-04-28
Translation update + improvement

Psy-Virus:
+ German and French translation updated due to latest changes.
+ Other languages updated to latest state via `gather-messages.sh`
+ Small improvement arount line 286 in `templates/header.inc.php`. I think it's better to have one coherent sentence because to translate one sentence in pieces is slightly confusing.

:-)


Afterster:
Ok thanks for this ;)
About the cookie sentence, I prefer to not have any html code in the translation file. We should found another way to do this, maybe by changing the English sentence to have more independent translation? But we really should avoid html here.

Psy-Virus:
Yep, I completely agree but the only way I see to benefit from both, a coherent sentence and avoiding html within it would be to formatting it without breaks. I also tried this and it wasn't looking that bad.

Psy-Virus:
Yep, I completely agree but the only way I see to benefit from both, a coherent sentence and avoiding html within it would be to formatting it without breaks. I also tried this and it wasn't looking that bad.

But to split one sentence into 2 or more parts is also ugly. :-(

Psy-Virus:
maybe
<code>            noty({text: '<?php printf(json_encode("<pre>" . T_('We have placed cookies on your computer to help make this website better. You can change your %s at any time.
Otherwise, we will assume you are OK to continue.

Click on this message to not display it again.') . "</pre>"), "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>"); ?>',</code>

With style options for that <pre> tag?

Afterster:
Mhh in extension to your idea, why not just calling `nl2br(T_('We have....` ?

Psy-Virus:
> why not just calling nl2br(T_('We have.... ?

hmmm because I didn't know about it :D

I think you mean it that way (?)
```php
noty({text: '<?php printf(nl2br(T_('We have placed cookies on your computer to help make this website better. You can change your %s at any time.\nOtherwise, we will assume you are OK to continue.\n\nClick on this message to not display it again.')), "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>"); ?>',
```
But the funniest thing is that the `\n` are not considered. I also had to remove the `jason_encode` for it to work...

Psy-Virus:
Okay, it seems that gettext interferes with \n

but this works:

```php
        noty({text: '<?php printf(str_replace( '\n', '<br />', T_('We have placed cookies on your computer to help make this website better. You can change your %s at any time.\n Otherwise, we will assume you are OK to continue. \n\n Click on this message to not display it again.')), "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>");
```

The only thing is, that gettext changes the `\n` in `\\n` so that the original text is processed as:

```text
We have placed cookies on your computer to help make this website better. You can change your %s at any time.\\n
 Otherwise, we will assume you are OK to continue. \\n
\\n
 Click on this message to not display it again.
```
And you also have to place these `\\n` in the translation otherwise it seems to break the the noty (noty is not shown then)

For example in German it has to be:

```text
Wir haben Cookies auf Deinem Computer platziert. Dies hilft die Funktionen der Webseite zu verbessern. Du kannst Deine %s jederzeit ändern.\\n
Andererseits gehen wir davon aus, dass Du mit den Verhältnissen einverstanden bist.\\n
\\n
Klicke auf diese Nachricht um sie nicht wieder anzuzeigen.
```

Do you think, that's something we can deal with?

Psy-Virus:
Okay, if there's no translation the original English(US) message is displayed.
So the only thing should really be, that the translator has to pay attention on these double backslashes.

Psy-Virus:
Alright, now it's done that way:

```php
        noty({text: '<?php printf(str_replace( '\n', '<br />', 
                    /* HINT: Translator, pay attention on double backslashes "\\" for the new line tags! Example: \n in source code is processed as \\n by gettext.
                    You must use it in your translation!
                    FYI "%s" is replaced by "cookie settings" */
                    T_('We have placed cookies on your computer to help make this website better. You can change your %s at any time.\n Otherwise, we will assume you are OK to continue. \n\n Click on this message to not display it again.')),
                    "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>"); ?>',
                type: 'warning',
                layout: 'bottom',
                timeout: false,
                callback: {
                    afterClose: function() {
                        $.cookie('cookie_disclaimer', '1', { expires: 365 });
                    }
                },
            });
```
The comment with that "HINT:" is automatically added to the po files as comment for the translator.

Maybe we should comment all stings which use variables in it, so the translator has no need to dig into the code for a lookup what's happening there, shouldn't we?^^

Okay, enough spamming for now. :D

Afterster:
Ok for the translated string, but not about the function.
Why using print_f? It should be removed and json_encode must remains, it makes sure the text string is well formated for javascript use (it has issue in French language otherwise for instance).
nl2br should be used instead of str_replace and double backslashes avoided. Just use double quote instead of single quote and it should work `T_("...`.
I know it's lot of discussions for few changes, but if we do it rights now then I can put it officially in the wiki. Thanks ;)

Psy-Virus:
Okay,
I've used printf because I was/I am running into issues if I use echo. I also thought it is cleaner to insert the html link for the cookie settings with a simple %s where "cookie settings" is translated as single string. And last but not least, to get a sentence what isn't split up into pieces and have to be translated part to part (except "cookie settigs"). :-)

The other thing, the single quotes... I also ran into issues when I was using double quotes. I think it's because of that html string.

I tested it now twice. I've swapped single and double quotes and tried various combinations. But it simply won't work. :D I bet it's the lack of know-how... in other words: I don't get it.^^

Psy-Virus:
So as far as I see, the only way I get the noty displayed on Ampache is:

```php
        noty({text: '<?php printf(nl2br(T_('We have placed cookies on your computer to help make this website better. You can change your %s at any time.\nOtherwise, we will assume you are OK to continue.\n\nClick on this message to not display it again.')),
                    "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>"); ?>',
                type: 'warning',
                layout: 'bottom',
                timeout: false,
                callback: {
                    afterClose: function() {
                        $.cookie('cookie_disclaimer', '1', { expires: 365 });
                    }
                },
            });
```
But without line breaks. 
It's one continuous line. But the \n are not displayed they're simply not processed. I've read that other escape sequences can interfere with nl2br. And if I see it right, in the html string are two. Both \".
Maybe that's the point?

Psy-Virus:
Yey, I got it!

```php
        <script type="text/javascript" language="javascript">
        noty({text: '<?php printf(json_encode(nl2br(T_("We have placed cookies on your computer to help make this website better. You can change your %s at any time.\nOtherwise, we will assume you are OK to continue.\n\nClick on this message to not display it again."))),
                    "<a href=\"" . AmpConfig::get('web_path') . "/cookie_disclaimer.php\">" . T_('cookie settings') . "</a>"); ?>',
                type: 'warning',
                layout: 'bottom',
                timeout: false,
                callback: {
                    afterClose: function() {
                        $.cookie('cookie_disclaimer', '1', { expires: 365 });
                    }
                },
            });
        </script>
```
That works like I want it. I don't know what I've done false before... Lack of concentration.

But this should be more in your opinion, shouldn't it?
Also if it's printf ^^

And sorry for that big discussion ~.~

Afterster:
Yeah looks good now! Thanks for the discussion and the translation ;)

Psy-Virus:
puh! *whiping sweat off my forehead* :D
I've to thank you for your patience! But you're also welcome though ;-D

PR #831, opened 2015-04-29
Added return true; for successfuly deleted catalogs.

wagnered:
I wonder how this was missed? ; ' >


Afterster:
Indeed... thanks!

Issue #840, opened 2015-05-04
Clear Cache When Searching Album Art


minnesotags:
This is a wishlist item. It would be good if the feature to search for album art cleared the cache first. The current behavior is that the displayed art options will be the same as your last previous search. You currently need to manually clear the cache before doing a search if you are doing multiple albums.

minnesotags:
Thanks!!!!! This is really great! :-)

PR #854, opened 2015-05-11
Stop image caching in art selection

hyperactivated:
Added "cache buster" parameter to image url in show_arts.inc.php as browsers seem to ignore cache control directives when images are inserted into DOM using javascript.

I believe this resolves issue #840.


Afterster:
That would do the trick indeed. Thanks ;)

Issue #810, opened 2015-04-21
Can't seem to stream channel using ogg


dhsc19:
I was able to successfully set up and stream a channel using mp3 stream type, but it doesn't seem to work if I try with ogg.  Ampache seems be initiating transcoding, my media player seems to be connecting, but no music coming though.

2015-04-20 17:11:35 [ampache] (channel) -> Now listening to /Music/Rush/Caress of Steel/01 Bastille Day.mp3.
2015-04-20 17:11:35 [ampache] (stream.class.php) -> Starting transcode for {/media/serverhdd/backup_source/netstore/Music/Rush/Caress of Steel/01 Bastille Day.mp3}. Type {ogg}. Options: Array
2015-04-20 17:11:35 [ampache] (stream.class.php) -> (
2015-04-20 17:11:35 [ampache] (stream.class.php) ->     [bitrate] => 64
2015-04-20 17:11:35 [ampache] (stream.class.php) -> )
2015-04-20 17:11:35 [ampache] (stream.class.php) -> }...
2015-04-20 17:11:35 [ampache] (media) -> Explicit format request {ogg}
2015-04-20 17:11:35 [ampache] (media) -> Transcode settings: from mp3 to ogg
2015-04-20 17:11:35 [ampache] (media) -> Command: ffmpeg Arguments: -i %FILE% -vn -b:a %SAMPLE%K -c:a libvorbis -f ogg pipe:1
2015-04-20 17:11:35 [ampache] (stream) -> Final transcode bitrate is 64
2015-04-20 17:11:35 [ampache] (stream) -> %MAXBITRATE% not in transcode command
2015-04-20 17:11:35 [ampache] (stream) -> %RESOLUTION% not in transcode command
2015-04-20 17:11:35 [ampache] (stream) -> %QUALITY% not in transcode command
2015-04-20 17:11:35 [ampache] (stream) -> Transcode command: ffmpeg -i '/Music/Rush/Caress of Steel/01 Bastille Day.mp3' -vn -b:a 64K -c:a libvorbis -f ogg pipe:1

mrpurplenz:
In my freshly installed system (3.7.0) libvorbis tries (and fails) to transcode to ogg.  For me the solution was to go to the config file and edit  the encode_args_ogg line.
I replaced "%SAMPLE%K" with  "max\(%SAMPLE%K\,45K\)"
[...]

PR #872, opened 2015-05-28
Enables Ogg (also works with Opus) channel streaming

Deathcow:
- bin/channel_run.inc:
        * extensive rewrite and restructuring
        * implements a variable-size server-side buffer
        * small bugfixes here and there
        * allow for more precise bitrate handling (via $nb_chunks_remainder)
        * small improvements to http serving

- lib/class/channel.class.php:
        * add $header_chunk variable and remember header of current filer for later use in channel streaming
        * added $chunk_size variable

### DIFF
--- a/bin/channel_run.inc
+++ b/bin/channel_run.inc
@@ -23,6 +23,8 @@
 define('NO_SESSION','1');
 define('CLI', 1);
 
+$chunk_buffer = '';
+$nb_chunks_remainder = 0;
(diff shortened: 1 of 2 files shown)


Afterster:
Merged, this fix #810.
I guess it took you severals hours to have good implementation so thanks a lot for this pull request!

PS: for the next one, I would just suggest to combine your commits related to bugfix on your not yet merged pull request with the original commit.

Issue #875, opened 2015-05-29
Writing theme: modify CSS and HTML


Niols:
Hello there,

I'm trying to write a theme for Ampache.

I created my own CSS file, and started writing it. Sometimes I had to write hacky things to be able to have the correct rendering, but most of the time that was OK, even if that would've been easier with the possibility to modify HTML.

Now I would like to have a better control on the iframes elements (in particular, I would like to customize the web player). I've not found any theme doing it and I wonder if this is possible.

Maybe by adding an other CSS file? The best way to do it would be to be able to have acces to the HTML. I see there are fils `show_*` in `templates/` so maybe this is the way to do it?

Whatever is the answer, I think that would be good to update the wiki page with it:
https://github.com/ampache/ampache/wiki/Themes

Thanks for answering and helping.
Niols

PS: link to 3rd party themes in wiki is dead

Afterster:
Because Ampache themes was developed ten years ago by people who doesn't maintain it anymore we removed all themes in current development version, except the Reborn theme. We found others ugly so we didn't wanted to maintain it. Our opinion ; if someone would like to take care about theme migration from https://github.com/ampache/ampache/tree/048032f7c3c6aeb36d2b3e5cd1a9173ca92982f4/themes and become the officiel maintainer for theme, feel free to suggest a PR.

Back to your original question, theming is not currently supported on web player as /templates/jplayer.midnight.black.css or /templates/jplayer.midnight.black-iframed.css style files are always included indifferently from theme selection. We should be able to define custom theming for web player too => marked as enhancement.

Niols:
I would say that we could replace all path in the template files (includes, css import, images) by a function that tests if the the file exists in theme's directory. If it's the case, this one is imported, and if not Ampache falls back on the default file. That would give a great freedom to themes writes that could completely rewrite the graphic chart.
I think that wouldn't be too hard to do. If you find this enhancement interesting (I do), I could write this and suggest a PR.

Afterster:
Yes that's how I would have fixed it too. Great if you can take care on this point, I will wait for the PR then.

Niols:
Even if I think that would be great to be able to modify the php files in templates, wouldn't that be too dangerous for the user? Because that would give a very important power to the theme writer…

SUTJael:
[...]

PR #878, opened 2015-06-04
Allow overwriting any template file in theme

Niols:
This is the pull request we talked about in issue #875.

I added the `UI::find_template` method, that tries to find the asked file in `$theme_path/templates/`, and, if it's not there, falls back on the file in `templates/`.

To prevent themes from having PHP access to the server, I added a `allow_php_themes` option set to false by default (infact null, because I didn't find out how to define the default value of an option).

Then I modified almost all the `/templates/…` occurrences by `UI::find_template(…)`. I say almost, because I did not modify all scripts related to installation, since I was not sure they had access to `UI`, and almost sure they had not access to `AmpConfig`.

This seems to be working for me, for the moment. I'll continue using it and testing it. I'll also start writing a theme.

I'm just opening this pull request so that we can talk about it and change things if they're not OK for you.


Afterster:
Ok for the first changes. I only have a short remarks on AmpConfig::get('allow_php_themes') check that should be first on the list in find_template function, for performance reason.

Take your time to finish your changes as best (and please clean your git history for the merge once completed), I will only merge it on develop branch after 3.8.0 release (currently waiting few more days for issues to throw, just in case).
Thanks.

Niols:
OK, I followed what you said: I recrafted my commit history so that it looks much cleaner, and I changed the order of the tests in find_template function. The function is firstly testing if the extension is php, and if it's the case testing if php is allowed.

I can still do some modifications.

Afterster:
Ok thanks for the change.
I just think you did some mess on your history clean because now one of my commit is on your change list :)

Niols:
Yes, I saw that. I don't know how I did that, and I didn't know how to cancel it. At least you're still the commit author ^^

SUTJael:
Thanks for you work @Niols. Any chance to update the PR with latest sources and based on new files versions? Thanks

PR #886, opened 2015-06-18
Improved support for the opus audio format

mrpi:
This patch enables playback of *.opus files in the web player (tested with current versions of firefox and chrome).


Afterster:
Looks fine for me, merged.
Thank you!

PR #889, opened 2015-06-26
Missing semicolon halts install/update, MariaDB 5.5.2

jbrain:
Missing semicolon halts install/update, MariaDB 5.5.2


Afterster:
Oh, you're right. This is important issue, I believe a quick minor update will be released asap.
Thanks!

PR #893, opened 2015-07-04
Fix calculate_art_size binary so that it works with new configurations

nan4k7:
calculate_art_size does not have into account the new possibility ampache has to store metadata on disk. With this changes this binary works again.


Afterster:
This was missing indeed. Again, thanks.

PR #895, opened 2015-07-04
Bug detecting OS

nan4k7:
I realized that this (which I pulled before) but was somehow changed is exactly the other way round in the IF.

It's ok now with the change I made.


Afterster:
Indeed... thanks!

PR #896, opened 2015-07-06
Add implode in array

nan4k7:
As I suggested in a bug report, implode should be added here to and not only in the "else"


Afterster:
I did it in a way to avoid code duplication with commit https://github.com/ampache/ampache/commit/9f51e5e38895883f337a807385f50e6eb0955933.
Thanks.

PR #911, opened 2015-07-29
POT-File update for Transifex

Psy-Virus:
The rest should be done by Transifex automatically :-)

I've also updated the Translation-words.txt

Write access to the locale directory would be nice :-P

### DIFF
--- a/locale/base/messages.pot
+++ b/locale/base/messages.pot
@@ -1,13 +1,14 @@
-# LANGUAGE Translation of Ampache.
-# Copyright (C) 2007 Ampache.org
-# This file is distributed under the same license as the Ampache package.
+# SOME DESCRIPTIVE TITLE.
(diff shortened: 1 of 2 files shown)


Psy-Virus:
Oh and @Afterster, did you notice my [comment](https://github.com/ampache/ampache/pull/852#issuecomment-100885279) about the php-cs-fixer.phar?

the link Travis uses, links to an outdated file. the newest one is always available here: http://get.sensiolabs.org/php-cs-fixer.phar

The version that Travis uses (http://cs.sensiolabs.org/get/php-cs-fixer.phar) is an alpha or beta 0.3 development version. And it sometimes does not fix the codestyle.

Afterster:
Yes thanks!

About php-cs-fixer.phar, I will update the script to use latest version, thanks for the information.

PR #917, opened 2015-08-05
Added functionality for streaming playlists to support track numbers

Fondor1:
* New column 'track_num' added to table stream_playlist
* Added new item 'track_num' to properties array within Stream_URL
* Added track number for xspf playlist format

xspf playlist generation now includes track numbers (if track numbers are included in the song) for streaming. This functionality should be feasible to add to other playlist formats provided the playlist supports track numbering.

### DIFF
--- a/lib/class/stream_playlist.class.php
+++ b/lib/class/stream_playlist.class.php
@@ -142,6 +142,7 @@ class Stream_Playlist
                         $url['info_url'] = $object->f_link;
                         $url['image_url'] = Art::url($object->album, 'album', $api_session, (AmpConfig::get('ajax_load') ? 3 : 4));
                         $url['album'] = $object->f_album_full;
+                        $url['track_num'] = $object->f_track;
                     break;
                     case 'video':
                         $url['title'] = 'Video - ' . $object->title;
@@ -393,6 +394,9 @@ class Stream_Playlist
             }
             if ($url->album) {
                 $xml['track']['album'] = $url->album;
+            }
+            if ($url->track_num) {
+                $xml['track']['trackNum'] = $url->track_num;
             }
 
             $result .= XML_Data::keyed_array($xml, true);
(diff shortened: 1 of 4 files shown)


Fondor1:
In order for this functionality to work on my existing database, I had to manually add the 'track_num' column to the table stream_playlist. For new installations, my modification of ampache.sql should be sufficient. I do not know how to handle the case for users with existing installations however. Is there an sql file that can be run to add just that column for existing users?

SUTJael:
You shouldn't edit directly the ampache.sql file. Please use the update scripts in: ``lib/class/update.class.php``.

Fondor1:
I've edited update.class.php as suggested, but I am a bit confused about how the update/new install process works. 

I expected that for new installs, ampache uses ampache.sql to generate the new database and doesn't touch update.class.php. For upgrading existing installations, only update.class.php is called. Is this understanding incorrect? If so, it would seem that both the sql file and update.class.php need to be changed.

I've made modifications in the latest pull request with the understanding outlined above. If this is not correct I would be happy to make any corrections and edits.

SUTJael:
Well, you're right and wrong.. :)
``ampache.sql`` is often aligned when we create a new release or when we believe that there's too much updates to apply.
``update.class.php`` is always checked to determine if there's some DB updates to apply since the installation.

In your commit, you changed manually the ``ampache.sql`` version. That's wrong, the database version should always remains to the last "released" schema. Then, after executing updates, your database version will automatically increase itself.

In conclusion, if someone install a new version, the ``ampache.sql ``will be executed AND all updates (number which depends on the actual db version). Hoping it's now understandable :)

Fondor1:
I think I understand! I will roll back all changes to `ampache.sql`, leaving the updates to `update.class.php` as this file is what will handle migrations from existing installations AND new installations.

PR #924, opened 2015-08-11
check mbstring.func_overload value during installation

anonymous2ch:
So no one will end up with a thousand corrupted mp3 files like me :)


Psy-Virus:
Hey @anonymous2ch,
would you please replace line 173 in `show_test_table.inc.php`;
```php
<?php echo T_('This tests whether PHP <a href="http://php.net/manual/en/mbstring.overload.php">mbstring.func_overload</a> is set as it may break the id3 tag support.'); ?>
```
into
```php
<?php printf(T_('This tests whether PHP %s is set as it may break the id3 tag support.'), "<a href="http://php.net/manual/en/mbstring.overload.php">mbstring.func_overload</a>"); ?>
```
?

So we don't have html within translation strings :-)

Psy-Virus:
oh and it's ID3 not id3...

Sorry for grammarnaziing ^^

Psy-Virus:
Another small improvement for lines 177-179

you could replace it with

```php
echo T_('Enabling Ampache ID3 tag write support along with mbstring.func_overload may result in irreversible corruption of your music files.\n\n
YOU HAVE BEEN WARNED!');
```

anonymous2ch:
Sorry for my bad English & even worse git skills - should I make a new commit or is there a way to just edit this one?

Psy-Virus:
I mostly reset my local repository to the latest commit (`git reset <hash> --hard`) before mine, reedit the files or readd them to the repo and do a `git push origin develop --force` So everything in my online repo gets overwritten and the PR has only one commit^^
It's actually less work that it sounds :D

And I bet there's a simpler way but I'm also not that familiar with git^^

PR #925, opened 2015-08-12
check mbstring.func_overload during setup & some video player CSS fixes

anonymous2ch:


templates/show_test_table.inc.php
```diff
@@ -164,6 +164,21 @@
...
 </tr>
+<tr>
+    <td valign="top"><?php echo T_('PHP mbstring.func_overload'); ?></td>
+    <td valign="top">
+    <?php echo debug_result(check_mbstring_func_overload()); ?>
+    </td>
+    <td>
+    <?php printf(T_('This tests whether PHP %s is set as it may break the ID3 tag support.'), '<a href="http://php.net/manual/en/mbstring.overload.php">mbstring.func_overload</a>'); ?>
+    <?php
+    if (!check_mbstring_func_overload()) {
```
Afterster: Don't do an additional check here, better to have only one unique sentence describing the test (why would we add specific process on this check and not on the other?).

Afterster:
Indeed this could result to corrupt files and having a new test as suggested would be good.
This PR is a dual PR with highly different subjects. Better to create two branch in your fork to create two PR so it can be discussed (and merged) separately. Could you details the videos changes? 
Thanks!

Psy-Virus:
Maybe we should add an additional checkbox for those warnings like "I've read the warnings and accept the Risk". And maybe, for security, if some known bad functions are enabled, the `test.php` should show up also after installation. And to not bother the user, maybe a config entry like "don't show this again" would be nice.

Or those things could be directly disabled through the `.htaccess`.

Oder would that be overkill?

Just my two cents^^

Afterster:
Ok I did the change and merge.
Thanks for this PR and the explanation on IRC.

@Psy-Virus I'm not a big fan to add additional installation steps. Maybe overkill in my opinion yes, but thanks for coming up with ideas ;)

Issue #912, opened 2015-07-30
Ampache stops playing on certain songs


hypfvieh:
Hi,

I have problems playing some songs with ampache.

Ampache stops playing those songs before they are finished.
Example: Ampache shows song length of 3:33 minutes, but song stops right after 3:20.

Other players, like foobar2000 or Audacious are playing these songs correctly.

All songs which behave like this, are regular MP3 VBR (Joint Stereo) files with id3Tag 1.1 and 2.3.

I tried to find out what is happening using the ampache debug log, but there is no usable information:
```
2015-07-28 10:49:50 [ampuser] (play) -> Playing file (/music/Artist/Album/01. Artist - Song.mp3}... 
2015-07-28 10:49:50 [ampuser] (play) -> Media type {mp3} 
2015-07-28 10:49:50 [ampuser] (play) -> Custom play action {} 
2015-07-28 10:49:50 [ampuser] (play) -> Transcode to {} 
2015-07-28 10:49:50 [ampuser] (play) -> Decided not to transcode 
2015-07-28 10:49:50 [ampuser] (play) -> Starting stream of /music/Artist/Album/01. Artist - Song.mp3 with size 6590694 
2015-07-28 10:49:50 [ampuser] (play) -> Registering stream stats for {Artist - Song}... 
2015-07-28 10:49:50 [ampuser] (user.class.php) -> Updating stats for {song/4711} {Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:39.0) Gecko/20100101 Firefox/39.0}... 
```
Other MP3s with the same encoding options (VBR, joint stereo) are working as expected.

I don't think it is a PHP memory problem, because the song is short, other tracks play well and my PHP memory limit is already set to 512MB.

Is there any way to debug this issue?

Afterster:
Are you using html5 as the audio back end ? If so, this could be related to the browser. Could you try to create a simple html5 test page with the audio tag playing the same mp3 file? Thank you.

hypfvieh:
Yes I'm using the HTML 5 player.

I've tested what you suggested, and created a small page using HTML 5 audio-tag to play the "broken" song (using same firefox version). 

Works like a charm, song is playing like it should (including the last 10 seconds).

Tried it again with ampache, still no luck. It stops playing before the end. I have this problem on the whole album, but with different "gaps". Some songs play. Some songs misses 5 seconds to the end, and one song misses 10 seconds.
I know that I have some more albums with that issue, but I can't remember which ...

[...]

PR #929, opened 2015-08-15
Patch for wrong file size in database

hypfvieh:
Patch which fixes the wrong size saved in database, as discussed in Ticket #912.

In short, Ampache stores the file size as found in some id3v2 tags. I have some songs which contains a "size" frame in the comment area of the id3v2 tag. Sadly the saved size is not the real file size, which could cause the song to stop playing too early when listening through Ampache.

I'm not sure if this patch is the ultimate solution. It works for me, but it may not work if the catalog is remote or something like that.
As I only have local catalogs, I cannot test it with other options.


Afterster:
Good catch.
I think comment->size is probably referred as TSIZ tag right? If so it was deprecated in ID3 2.4 so I think your way to fix it is good.

Thanks for the PR!

Issue #922, opened 2015-08-11
Last.fm API are deprecated, upgrade planned?


nioc:
Hello, Ampache uses the old API for handhsake and submission of now-playing to Last.fm user profiles in [/lib/class/scrobbler.class.php] (https://github.com/ampache/ampache/blob/47a6585c88f3fb93b1e9516467efbcd5a6cc921a/lib/class/scrobbler.class.php).
As we can see in the [Last.fm API page] (http://www.lastfm.fr/api/submissions) :
>This protocol is now deprecated! Please use the Scrobbling API.
Note: There is now a new way for sending scrobbles called the Scrobbling API which is integrated into the rest of the Last.fm Web Services as opposed to being a separate service like 1.2.1. New clients should all use the Scrobbling API and not this protocol as we will not be issuing new client ids for this protocol and are likely to disable the "tst" client id at some point.

Are there any plans to migrate to theses new services? I just take a look, and the main problem  will be how (and where) to add the button to link user's Lasf.fm account and Ampache (it opens a lasf.fm GUI where the user have to login with his Last.fm account for granting the Ampache access to his profile) and return a token by callback.
I will try to add some functions to call authentication and scrobbling services in a new branch.

Any suggestions will be welcome :smiley:

SUTJael:
Any kind of help is appreciate ;)
If you can take care about upgrading the API, that would be awesome ;)

nioc:
Ok, I'll start tonight :smile:
Do I have to create a new branch or working on "develop"?
I will need to add an hyperlink which opens last.fm website for the granting process, on which ampache GUI can I add it?

SUTJael:
You could maybe add the complete enrollment process into the Last.fm settings? (Settings->Plugins)
You can create a new branch on your local repository and then you will be able to create a pull request when everything works.

nioc:
Hello, here are a few news :
-due to the Last.fm website upgrade, the API documentation was not accessible... (thanks Google cache),
-Libre.fm and Last.fm exposed the sames API,
-here is the sequence :
![diag-libre fm](https://cloud.githubusercontent.com/assets/5225411/9410925/5ac793b6-4823-11e5-95ba-d0fdaffab34b.png)
-I use the preference screen (plugin tab) for the GUI with a new link for each plugin :white_check_mark:,
-it opens the authentication url for the plugin and catch the returned token (by passing callback parameter) :white_check_mark:,
-receiving the token triggers the call of the getSession API [TODO :construction:],
[...]

Issue #349, opened 2014-07-26
last.fm multiple users with same scrobbling login failure


stebe:
I noticed that if I'm playing songs from two different ampache accounts and they both have the same last.fm login I get what may be session errors. 
 
2014-07-26 09:39:28 [admin] (Last.FM) -> Submission Successful
2014-07-26 09:41:01 [stebe] (Last.FM) -> Handshake Failed: Invalid POST URL retu
rned, unable to continue. Sent: 
2014-07-26 09:41:01 [stebe] (Last.FM) -> GET /?hs=true&p=1.2&c=apa&v=0.1&u=stebl
ublu&t=1406381860&a=52d27fbd894dc3ade2fc9cd91c0fd07e HTTP/1.1^M 
2014-07-26 09:41:01 [stebe] (Last.FM) ->  
2014-07-26 09:41:01 [stebe] (Last.FM) -> ---- 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Received: 
2014-07-26 09:41:01 [stebe] (Last.FM) -> HTTP/1.1 200 OK^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Date: Sat, 26 Jul 2014 13:37:40 GMT^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Pragma: no-cache^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Cache-Control: no-cache^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Content-Type: text/plain;charset=ISO-88
59-1^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Content-Length: 8^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Server: Jetty(8.1.11.v20130520)^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> ^M 
2014-07-26 09:41:01 [stebe] (Last.FM) -> BADAUTH 
2014-07-26 09:41:01 [stebe] (Last.FM) ->  
2014-07-26 09:41:01 [stebe] (Last.FM) -> --------- 
2014-07-26 09:41:01 [stebe] (Last.FM) -> Expected:Array 
2014-07-26 09:41:01 [stebe] (Last.FM) -> ( 
2014-07-26 09:41:01 [stebe] (Last.FM) ->     [0] => BADAUTH 
2014-07-26 09:41:01 [stebe] (Last.FM) ->     [1] =>  
[...]

PR #935, opened 2015-08-23
Upgrade scrobbling API (last.fm and libre.fm)

nioc:
Upgrade the scrobbler class, Lastfm and Librefm plugins to use the new API.
Authentication is changed:
-admin needs to create a last.fm API account, set their API key and secret in the ampache.cfg file,
-users need to grant Ampache application to last.fm / Libre.fm services by using preference tab > plugins.

Resolves ampache/ampache#922

Add the "spread the love" function to share your favorites on Last.fm and Libre.fm.

Change-Id: Ia81334f4f4506258a27d40b6c257a1e38bd69097
Signed-off-by: nioc <nioc@users.noreply.github.com>

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -477,6 +477,11 @@ art_order = "db,tags,folder,musicbrainz,lastfm,google"
 ; recommendations and metadata.
 lastfm_api_key = "d5df942424c71b754e54ce1832505ae2"
 
+; Last.FM API secret
+; Set this to your Last.FM api secret to actually use Last.FM for
+; scrobbling.
+lastfm_api_secret = ""
+
 ; Wanted
 ; Set this to true to enable display missing albums and the
 ; possibility for users to mark it as wanted.
--- a/lib/class/scrobbler.class.php
+++ b/lib/class/scrobbler.class.php
@@ -23,34 +23,89 @@
 class scrobbler
 {
     public $error_msg;
-    public $username;
-    public $password;
     public $challenge;
-    public $submit_host;
-    public $submit_port;
-    public $submit_url;
+    public $host;
+    public $scheme;
+    public $api_key;
     public $queued_tracks;
-    public $reset_handshake = false;
-    public $scrobble_host = 'post.audioscrobbler.com';
+    private $secret;
 
     /**
      * Constructor
      * This is the constructer it takes a username and password
      */
-    public function __construct($username, $password,$host='',$port='',$url='',$challenge='',$scrobble_host='')
+    public function __construct($api_key, $scheme='https',$host='',$challenge='', $secret='')
     {
         $this->error_msg = '';
-        $this->username = trim($username);
-        $this->password = trim($password);
         $this->challenge = $challenge;
-        $this->submit_host = $host;
-        $this->submit_port = $port;
-        $this->submit_url = $url;
+        $this->host = $host;
+        $this->scheme = $scheme;
+        $this->api_key = $api_key;
+        $this->secret=$secret;
         $this->queued_tracks = array();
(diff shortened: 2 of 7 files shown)


Afterster:
I reviewed the code and your PR looks good, thank you for taking care about this part.
I think it could even fix few issues referenced in #349.

Last.FM is currently migrating to their new website, I will test it properly when it become stable again.

nioc:
Good news :bowtie:
I noticed, the last.fm website badly handle the https auth call. Sometimes it redirect to http page, sometimes not. I don't know if it is a website migration issue (and https will work at each call), or if we have to call on http only.
I don't touch the async scrobbler.class.php, so may be the issue you referred shouldn't be fixed. I'm waiting for your tests :smile:

Afterster:
@nioc I was reviewing again your PR before merge and I got something wrong. You're always using the global lastfm api key/secret. Global lastfm api key (in ampache.cfg.php) should be used for catalog related operations only (similar artists, ...). Each Ampache user should be able to scrobble to his own lastfm account and then lastfm/librefm api key/secret should be plugin parameters.
Sorry to report this only now after few weeks, and thank you!

nioc:
Hi!
I have to take a look, but if I remember correctly, we have the application key (with customized icon and specific callback url set in last.fm) which is the same for each users of a same installation, and the user session key (which is specific to each users and obtained by the auth sequence).
Why do you think this is wrong?

Afterster:
Oh you're probably right in fact that could be a misunderstanding on my side. Give me some time to review/test it properly again once back at home.

Afterster:
Ok after testing better this time, works good. I'm gonna merge it, thanks for the good PR!

nioc:
Great :smiley: 
May be, we should update the wiki for the new authentication sequence and the need to set a custom API key?

Afterster:
Yes wiki should be updated indeed.
I just opened the wiki for public contribution btw, no reason to limit it as long we are frequently keeping a eye on changes ;)

ciroiriarte:
It it possible to disable all interaction with lastfm?, I'm having issues with timeouts and errors when no connected to Internet

PlusMinus0:
Just a question, is an api key now mandatory for scrobbling? I skimmed through the code and it seems to have replaced the username as means of authentication. Am I right?

nioc:
Hello !
@ciroiriarte yes you can disable the lastFM function in the plugin tab. If it is not enabled, it should not make any call to lastfm server (neither scrobbling or love songs, however, the related song may be called I don't know about it).
@PlusMinus0 yes API key is mandatory with the new lastFM API, you can have one by submitting [this form] (http://www.last.fm/api/account/create), then you have to set both `lastfm_api_key` and `lastfm_api_secret` variables it in your `config/ampache.cfg.php`

PlusMinus0:
@nioc thank you for your reply. I did that and I still couldn't get it to work. I entered my the keys into the ampache.cfg.php and my credentials into the plugin's page. But when I do, my Ampache just doesn't play songs anymore. I haven't yet gotten to skimming through the logs because I'm running Ampache in a docker container and `docker logs` doesn't print anything useful.

nioc:
You should have two informations in the config file.
And you don't have to enter credentials anymore, but you should have a button to go on last.fm website and sign-in. Can you check the last.fm plug-in installed version?

PlusMinus0:
Version 000004.

I have the key and the secret in the config file.

nioc:
Yes, you need to upgrade it by the plugin page, and then process the authentication by the button on the user's plug-in setting.
Let me know if it solves your issue.

PlusMinus0:
Yes, updating the plugin helped. Thank you!

PR #938, opened 2015-08-27
Fixed getAlbumList2 to always return a album list

hypfvieh:
getAlbumList2 now correctly returns an empty album list,
if the requested offset is larger than the album count in catalog.

This is required for some players (like clementine) to allow fetching
of all albums.


Afterster:
Done in another way with https://github.com/ampache/ampache/commit/26e3327f3497c9c9e17cf187b59f76574fbd6722
Thanks for the issue identification!

PR #948, opened 2015-09-08
Speed up (a lot) subsonic getArtists method

nicklan:
- Make Catalog::get_artists return an album count as "albums".
- Pass through to subsonic api calls to avoid running query for every
artist to count albums (as was happening previously)


Afterster:
Merged, thanks!

PR #949, opened 2015-09-09
Add advanced_search api call

nicklan:
This allows using the xml api with the full advanced search possibilities.  It does require understanding how the parameters relate to the search performed however, which requires looking at the code.  Something more self documenting would be more code, but easier to use.  Another option would be adding some documentation to the wiki detailing the possible parameters to this call.

### DIFF
--- a/lib/class/api.class.php
+++ b/lib/class/api.class.php
@@ -706,6 +706,23 @@ class Api
     } // search_songs
 
     /**
+     * advanced_search
+     * Perform an advanced search given passed rules
+     * @param array $input
+     */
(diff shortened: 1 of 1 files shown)


Afterster:
Thanks, indeed it is currently missing!

But the search doesn't necessary return songs if `type` argument is passed on the search and target another object type (album, artist, playlist, label, user). You should force the type to song or handle xml result according to the type.

nicklan:
Okay, will have a look at fixing that when I have a minute, thanks.

PR #959, opened 2015-09-17
Translation base catalog updated + others

Psy-Virus:
+ Translation base catalog updated
+ `translation-words.txt` renamed to
`translatable-database-strings.txt`. It makes more sense and everyone knows where the strings within come from.
+ `gather-messages.sh` updated variables and switches + new header text for `translatable-database-strings.txt`

:-)


Afterster:
I agree, it's more explicit.
Thanks!

PR #978, opened 2015-09-29
Changed filetype of blankalbum.jpg -> png + others

Psy-Virus:
Reason;
PNG files are better for theming because of the supported alpha channel. And everything else is also png. :-)

Changes:
/image.php
+ changed `$mime` and `$defaultimage` to fit png file
type

other
+ deleted old blankalbum.jpg/gif and added the png
replacement in
`/images/` and `/themes/reborn/images/`


Afterster:
I agree, thanks!

PR #983, opened 2015-10-03
Improve performance around SQL query

thinca:
In my environment, the time it takes to open the top page:
Before: About 2.6s
After: About 0.08s


Afterster:
Oh god you're right. Thanks for this, merged.

PR #988, opened 2015-10-13
Update show_random_albums.inc.php

Bidules079:
If you choose to resize images with the "resize_images" thumb 3 provides a 80x80 image but it is displayed in 100x100. Thumb 1 must be asked, wich gives an image of 100x100.


Afterster:
Correct according to the code. Thanks!

PR #994, opened 2015-10-19
Added logic to remove leading 0 from season and episode number.

wagnered:
Tvdb can't deal with leading 0 in URL. This fix strips out the leading 0's.


Afterster:
Thank you for this PR @wagnered.
Nothing to say about the code but just something to say about the commit history.
It would be great if you try to have a better git history with your PR as I believe you will continue to do few of them in the future again :). When you want to synchronize with origin, could you do a rebase instead of a merge? This would avoid merge junk history on PR.
I'm waiting for a cleaner commit history, let me know. Thanks!

wagnered:
"Rebase" a new word.  Actually, I heard of it, but did not understand it's use.  I'll close this PR and do a reset HEAD or something then try to understand the rebase command.

PR #998, opened 2015-10-29
various formatting fixes on different *_row.inc.php

Psy-Virus:
+ Fixed consecutive php close and open tags ( ?> <?php )
+ Made some oneliner a little bit prettier


Psy-Virus:
The braces? This makes no sense. All opening brackets have its closing brackets.
Or does this include the positioning? If so, we should get rid of the script or find fixers that do not change good looking code into eye cancer. ^^

PR #1014, opened 2015-11-19
Fix metadata errors

Razrael:
Hi again

This one fixes the errors introduced with the Metadata feature.

Main changes:
- Switch class loading to composer (psr-4). The old classloader was still working somehow (at least the import still worked for me), but the composer one threw errors.
- Fixed some typehints and similar stuff. Probably introduced this when I refactored some of my code.
- Improved filtering of metadata. The old one was not working as intended because I did something stupid...so just all data got imported. If you already imported songs with the new metadata feature, you could just truncate the `metadata` and `metadata_field` tables and update your library.
- updated needed PHP version to PHP 5.4 because this feature is using traits

Please make sure to run a `composer dumpautoload` after checking out the feature.

Hope I have found all introduced errors :-)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -318,5 +318,11 @@
             "replace": false,
             "merge-extra": false
         }
-    }
+    },
+
+	"autoload": {
+		"psr-4": {
+			"Lib\\": "lib/class"
+		}
+	}
 }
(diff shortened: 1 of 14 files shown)


Afterster:
Looks like to fix many issues indeed. Thanks for taking care of this until the end, merged ;).

PR #1019, opened 2015-11-21
Lib needs to be lib on GNU/Linux

hristoast:
This was changed in commit `69a4a5f1f47a8e0caeb77c6835e0361549253937`, after which ampache stopped working with errors like this in my logs:

    2015/11/21 08:11:29 [error] 4311#0: *750664 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Trait 'Lib\Metadata\Metadata' not found in /home/ampache/html/lib/class/song.class.php on line 25" while reading response header from upstream, client: 1.2.3.4, server: domain.tld, request: "POST /server/ajax.server.php?page=index&action=reloadnp HTTP/1.1", upstream: "fastcgi://unix:/run/ampache-php-fpm.sock:", host: "domain.tld", referrer: "http://domain.tld/index.php"

This PR reverts that, so ampache works again. It seems the cause is [this](https://github.com/ampache/ampache/blob/develop/composer.json#L325) not mapping as expected.

### DIFF
--- a/lib/class/Interfaces/Model.php
+++ b/lib/class/Interfaces/Model.php
@@ -21,7 +21,7 @@
  *
  */
 
-namespace Lib\Interfaces;
+namespace lib\Interfaces;
 
 /**
  * Description of Model
--- a/lib/class/Metadata/Metadata.php
+++ b/lib/class/Metadata/Metadata.php
@@ -21,7 +21,7 @@
  *
  */
 
-namespace Lib\Metadata;
+namespace lib\Metadata;
 
 /**
  * Description of metadata
@@ -59,8 +59,8 @@ trait Metadata
      */
     protected function initializeMetadata()
     {
-        $this->metadataRepository      = new \Lib\Metadata\Repository\Metadata();
-        $this->metadataFieldRepository = new \Lib\Metadata\Repository\MetadataField();
+        $this->metadataRepository      = new \lib\Metadata\Repository\Metadata();
+        $this->metadataFieldRepository = new \lib\Metadata\Repository\MetadataField();
     }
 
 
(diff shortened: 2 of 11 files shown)


gabeos:
I have same issue. Additionally, Beets module in modules/Beets is not found. Solved by adding mapping in composer.json.

Afterster:
Thanks for this PR @hristoast but I'm not sure why this is happening. If correctly mapped, uppercase shouldn't be an issue. I will also include @Razrael in the discussion as he did this commit :).
@gabeos, about your Beets error, could you describe your fix? Thanks.

Razrael:
I quickly fixed the beets error. I basically missed some uses when refactoring.

About the  upper / lowercase: It is not dependent on a specific system like GNU/Linux because its just plain PHP.
Did you update with a `git pull` or with the update button?
The update button also includes `composer install`, which could get missed after a `git pull`.

I _think_ it might be because of the autoloading. When you turn it back to lowercase lib, it gets included via the old autoloader. Switching to uppercase Lib it will be included with PSR-4 from composer. But it may need to get informed about those classes.
Please try a `composer dump-autoload`. If this is not helping, please tell me on what actions the error occurs.

hristoast:
Hello, thanks for the responses!

The problem happened when I clicked "Upgrade", via the UI. I've done the same thing just now and it's all good. I'll close this out. Thank you so much for looking at this and for your work on this project!

gabeos:
ditto on the thanks for the quick fix. haven't tested new update yet but I assume it's all good. will post if not. 

just FYI though, this problem persisted after both updating via UI as well as a manual `composer autodump` and fresh  install.

PR #1022, opened 2015-11-24
Changed tvshow.summary to tvshow.overview to remove

wagnered:
conflict with tvshow_episode.summary.
Modified edit dialog and update functions to edit tvshow.overview.
Modified tvdb and tmdb plugins to reflect change to tvshow.overview.
Fixed Vainfo::_parse_filename after regression tests exposed bug.

### DIFF
--- a/lib/class/tvshow.class.php
+++ b/lib/class/tvshow.class.php
@@ -274,7 +274,7 @@ class TVShow extends database_object implements library_item
      *
      * Checks for an existing tv show; if none exists, insert one.
      */
-    public static function check($name, $year, $overview, $readonly = false)
+    public static function check($name, $year, $tvshow_summary, $readonly = false)
     {
         // null because we don't have any unique id like mbid for now
         if (isset(self::$_mapcache[$name]['null'])) {
@@ -313,10 +313,8 @@ class TVShow extends database_object implements library_item
             return null;
         }
 
-        $sql = 'INSERT INTO `tvshow` (`name`, `prefix`, `year`, `overview`)' .
-            'VALUES(?, ?, ?, ?)';
-
-        $db_results = Dba::write($sql, array($name, $prefix, $year, $overview));
+        $sql = 'INSERT INTO `tvshow` (`name`, `prefix`, `year`, `summary`) VALUES(?, ?, ?, ?)';
+        $db_results = Dba::write($sql, array($name, $prefix, $year, $tvshow_summary));
         if (!$db_results) {
             return null;
         }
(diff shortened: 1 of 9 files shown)


Afterster:
Thanks. Looks ok but can't we just rename `summary` for tvshow to `tvshow_summary` inside plugins results and still keep `summary` as the db field name? I don't see a reason why this should be named `overview` here, the conflict is only when getting metadata from plugins, isn't it?
If I'm wrong, then you should also update https://github.com/ampache/ampache/blob/develop/lib/class/tvshow.class.php#L29 to `overview` (and refs to this var), but if I'm right then no need to change the db or classes but only plugins and their handling.

wagnered:
First, "overview" is how it is labeled at the point that the tmdb plugin
fetches it.

Second, It is a more accurate description of the data as it describes
with more details, the entire set of episodes/seasons.

Third, I really didn't catch it until a few days ago and just recently
made the change to the variable tvshow.summary to tvshow.overview. :'{ I
forgot to submit the change to the PR.

I guess my rational is for  accuracy and consistency in labeling.

BUT, if there exists some rational for not changing the column name from
summary, to overview, I will gladly make the necessary changes from
"overview" to "tvshow_summary" in all the affected classes.


On 11/26/2015 11:39 AM, Afterster wrote:
> Thanks. Looks ok but can't we just rename |summary| for tvshow to
> |tvshow_summary| inside plugins results and still keep |summary| as the
> db field name? I don't see a reason why this should be named |overview|
> here, the conflict is only when getting metadata from plugins, isn't it?
> If I'm wrong, then you should also update
> https://github.com/ampache/ampache/blob/develop/lib/class/tvshow.class.php#L29
> to |overview| (and refs to this var), but if I'm right then no need to
> change the db or classes but only plugins and their handling.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ampache/ampache/pull/1022#issuecomment-159984062>.
> 

-- 
73 From
Ernie D
Federal Way, WA | CN87ug

wagnered:
OK, I reverted the change to tvshow.overview back to tvshow.summary.

Afterster:
Thanks @wagnered :smile:. Merged!

wagnered:
Noticed problem with vainfo.class.php after fetching upstream to update my local and fork.  It seems that somehow I added an unclean merged file. I cleaned up the file but then ampache failed. I'm getting a blank screen.

 I'm debugging it now it now and will keep at it until I find out what is wrong. The Eclipse IDE doesn't indicate serious errors (only cvc type 3.1.3 errors which were not a problem before.)
So I'll step through it until I find the exception.

wagnered:
Found the problem:  Another vendor was added and I didn't realize it. Ampache is working now.

PR #1028, opened 2015-11-29
Cleaned merge info from vainfo.class.php.

wagnered:
Sorry about that.  I thought I cleaned the file before adding it to staging.


Afterster:
Oh I missed that...

PR #1058, opened 2015-12-18
Update upnpplayer.class.php

jahak7:


Afterster:
Thanks @jahak7 but any reason for such changes?

jahak7:
@Afterster https://secure.php.net/manual/ru/function.is-null.php#84161

jahak7:
@Afterster Micro optimization

jahak7:
@Afterster https://gist.github.com/jahak7/52856ec4a9cb54792305

Afterster:
This is really micro optimization, I don't think Ampache is at this level of performance to see a difference. I generally prefer to lost few nanoseconds and have a better code btw.
But here `is_null` or `=== null` doesn't really matter for code readability, I'm gonna merge your PRs thanks ;).

jahak7:
@Afterster Thank you for ampache :)

PR #1063, opened 2015-12-20
Improvements to let the readme look more personal

Psy-Virus:
First it only was the fact that Ampache's dev-version was described as "It definitely will be unstable" at least in my thought. So I changed it to be a little bit ~~softer and then it got me.

Hate it? Love it? Let me know! ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ choosing a suitable method for doing so.
 Recommended Version
 -------------------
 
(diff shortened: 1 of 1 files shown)


Psy-Virus:
Well, this phrase (Hate it?[...])should be in the footer with a Link to the Google Forum

Afterster:
Yep I'm a formal guy as you may have noticed :). Thank you.

PR #1065, opened 2015-12-21
Further small improvements and fixed typos

Psy-Virus:


Afterster:
I don't completely agree with your changes. Ok to add a link to webchat but most people are not using this specific web-based client, we should continue to provide address connection.

Psy-Virus:
Oh sure, than it simply was a mistake, I thought that it should point to the web client, didn't thought that it could be a special address for local clients...

Updated the commit and described the web client as alternative.

:-)

PR #1069, opened 2015-12-22
Fix too big Album Art on index while Grid View is disabled

Psy-Virus:
Without this, you'll get:

tobigalbumartonindex


Afterster:
This change was excepted and only happening if you disabled Grid View on albums list.

But in your case width/height are too big for your screen I agree, default size was changed to 200px now.
Would be better if the art size could be automatically adapted to screen size.

Psy-Virus:
Ah, okay.. But why not for the videos on the index?

PR #1082, opened 2016-01-01
Return up to 99 999 starred artists/albums/tracks to subsonic

zerodogg:
Subsonic clients (at least dsub) appears to expect all starred artists, albums and tracks to be returned when requesting a list of starred items. Ampache was returning only 10, which broke starred functionality in dsub. This changes the subsonic API class to return up to 99 999 items of each type instead, enabling starred functionality to work properly.

I have tested the change using dsub as a client to ampache.


Afterster:
I was excepting to see paging here but you're right Subsonic API doesn't support paging on this function.
Then your changes looks valid I'm gonna merge it, thank you for the issue identification and fix :smile:.

PR #1085, opened 2016-01-05
Update search.js

Bidules079:
remove(); is a jQuery method. Are you using jQuery? If so, you could do: $(this).parent().remove();.


Afterster:
Indeed, thanks

PR #1093, opened 2016-01-14
General Translation Update

Psy-Virus:
General translation update with Transifex Translation files.
Some of this Translations are NOT reviewed. So no guarantee for correctness!

I also corrected the help screen of gather-messages.sh, added missed functions.

Thanx to all Transifex Translators!

:-)


Afterster:
This would still be better than current translation anyway!
 Thanks.

PR #1107, opened 2016-01-22
Fix windowed web player playlist width

nearwood:
While keeping ajax playlist intact. For issue #1107


Afterster:
This is already done if you take a look on the code, but thank you :).

PR #1131, opened 2016-02-11
Fix playlist_songs API

thinca:
`<playlisttrack>` tag does not appear in each items.
And errors occurred:
```
[Runtime Error] Invalid argument supplied for foreach()
```


Afterster:
Yep, thanks!

PR #1147, opened 2016-02-21
Rework LDAP Auth

Niols:
What have been done:

- Ignore `composer.lock` files

- Add a default value to `AmpConfig::get`.
  The default behaviour is still the same: `AmpConfig::get($key)` returns the value associated to the given key, and `null` if the key doesn't exist.
  But you can now provide a default return value: `AmpConfig::get($key, $default)` returns the value associated to the given key, and `$default` if the key doesn't exist.

- Rework LDAP section in config file
  This section should be a little cleaner, but might need more documentation.

- Rework LDAP auth
  - Move it from `lib/class/auth.class.php` (`Auth::ldap_auth`) to `lib/class/ldap.class.php` (`LDAP::auth`). The parameters are the same, but the function has been fully rewriten:
  - Code cleanup (in particular, no nested ifs anymore...)
  - Parameter for LDAP protocol version (default to 3)
  - Parameter for LDAP group's member attribute name (default to "member")
  - Parameter for StartTLS (default to false)
  - Support for custom name fields. For instance, on some LDAP servers, cn=uid, so that's not giving you the name. But you can take the first name from `givenName` and the familiy name from `sn`. Now you can provide the parameter `ldap_name_field = "givenName sn"` that will cover this case

In particular, the default behaviour of Ampache (with no extra config) should be exactly the same (backward compatibility...), but this allows better future reworks (because LDAP is now in a specific file, and because the code is much cleaner imo), a wider compatibility with LDAP servers, … and that's already something ^^

This code is working great against my personnal LDAP server. But there are so many ways of configuring a LDAP server that I could not cover all cases in tests… I might need some help from people using the LDAP auth on this one.

If this PR is accepted, I'll also rework the LDAP wiki page.

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -680,41 +680,58 @@ site_charset = UTF-8
 ; LDAP login info (optional)                             #
 ;#########################################################
 
-; LDAP filter string to use (required)
-; For OpenLDAP use "uid"
-; For Microsoft Active Directory (MAD) use "sAMAccountName"
+; LDAP server URL (required)
 ; DEFAULT: none
-;ldap_filter = "(sAMAccountName=%v)"
+;ldap_url = "ldap://localhost/"
+;ldap_url = "ldaps://localhost/"
 
-; LDAP objectclass (required)
-; OpanLDAP objectclass = "*"
-; MAD objectclass = "organizationalPerson"
-; DEFAULT null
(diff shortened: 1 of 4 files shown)


Afterster:
Looks good, thanks for your work.
But why ignoring composer.lock (and not removing it from git btw)? This file should be committed.

Niols:
I removed this commit, thanks for noticing!

PR #1158, opened 2016-03-01
Only allow ID3 scanning for local catalog if "getID3" is declared in …

Deelight-fr:
…the metadata_order config parameter.


Deelight-fr:
I don't understand why the Travis test fails. Fixing the errors alters ~100 files I didn't edit... (eof_ending).

Afterster:
Yep indeed, thanks for the fix. And thanks for fixing Travis btw, my bad about EOF ending but I believe that's also a bug on php-cs-fixer.

Issue #1162, opened 2016-03-02
Server sent events while managing catalog doesn't work in Chrome based browsers


thejk:
# Environment
 
 * Ampache version: 3.8.x
 * Web server + version: lighttpd 1.4.33
 * Server operating system: GNU/Linux
 * Client operating system: Any
 * Client type:
   - [x] Web
 
# Description
Select a Chrome based browser, go to ampache, Show Catalogs and select to Add, Update or any of the other actions that displays a progress status via server sent events.
Instead of getting status events a "Server-Sent Events connection error. Re-connection..." message is shown and in the console of chrome you get:
EventSource's response has a charset ("iso-8859-1") that is not UTF-8. Aborting the connection.

Afterster:
Merged!

PR #1163, opened 2016-03-02
Specify utf-8 charset for event-stream content to make Chrome happy

thejk:
Otherwise chrome complains with:
EventSource's response has a charset ("iso-8859-1") that is not UTF-8. Aborting the connection.

Solves issue [#1162](https://github.com/ampache/ampache/issues/1162)


Afterster:
Nice catch, I wasn't aware of this requirement. Thanks!

PR #1168, opened 2016-03-07
General Translation update and fix for dup messages in messages.pot

Psy-Virus:
Hey, I know it's a while since the last update, sorry for that.

But here's another complete update, including the latest translations
from Transifex and a complete updated and fixed source catalog (messages.pot).

*I was poked by Transifex because of an error in the source strings.

:-)


Afterster:
Yeah I did some mess :). Thanks!

PR #1197, opened 2016-04-05
FIX for #1195 & New position for update notification

### LINKED ISSUE #1195: topmenu-music.png is missing from /images directory
# Environment
 
 * Ampache version: Ampache 3.8.3-develop
 * Web server + version: Apache 2.4.10
 * Server operating system: Debian 8
 * Client operating system: *
 * Client type:
   - [x] Web
   - [ ] Ampache
   - [ ] SubSonic
   - [ ] Plex
   - [ ] DAAP
   - [ ] UPnP
   - [ ] WebDAV
 * Client application + version: 
 
I get a 404 on page load for the 'Artists' icon on the top menu.  I found that topmenu-music.png is missing from the /images directory, and a workaround is to do the following:

`ln -s topmenu-artist.png topmenu-music.png`

Psy-Virus:
FIX for #1195: file path correction.

New position for update notification:
Since ever, I hated it that the Update notification is within the headerbox. Now it's on the top of the content div like the "updated config" notification.

I also kicked out one permission check so the "autoupdate" notification and/or the "new config" notification is only shown if the permission is 100... Like before but in another way. ^^

Before, there was a `<span>...</span>`, no matter what user was logged in. Now there is a div with `id="update_notify"` what is also only shown, if the admin is logged in.

And it looks much cleaner on the UI. :-)

I also cleaned the depending CSS rule in `/reborn/templates/default.css`

### DIFF
--- a/lib/class/autoupdate.class.php
+++ b/lib/class/autoupdate.class.php
@@ -226,7 +226,7 @@ class AutoUpdate
      */
     public static function show_new_version()
     {
-        echo '<div>';
+        echo '<div id="autoupdate">';
         echo '<font color="#ff0000">' . T_('Update available') . '</font>';
         echo ' (' . self::get_latest_version() . ').<br />';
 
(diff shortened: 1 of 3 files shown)


Afterster:
Yep better indeed.
Thanks ;)

PR #1202, opened 2016-04-11
Translation catalog update...

Psy-Virus:
... due to latest changes. I'll do a general translation update after this is merged and Transifex has synced it.

:-)


Afterster:
Thanks!

PR #1205, opened 2016-04-14
Promised general translation update

Psy-Virus:
As Promised in #1202, here it is.... (drum-roll) ... yagtu (yet another general translation update). *yey*

:-]


Afterster:
Yeah! Back from being AFK, always a pleasure to see your translation PR :). Thanks.

PR #1222, opened 2016-05-17
Base translation catalog and translation update

Psy-Virus:
As always, fresh generated base translation catalog and pulled latest
translations from Transifex.

:-)


Psy-Virus:
The Travis fail isn't my fault. It's the `lib/class/catalog.class.php`. It's not within my PR.

:-)

Afterster:
Thanks ;)

Issue #123, opened 2014-04-15
Column's filtering review


Afterster:
_Issue by **[SUTJael](https://github.com/SUTJael)** from Friday Mar 21, 2014 at 13:21 GMT_
_Originally opened as https://github.com/ampache-doped/ampache-doped/issues/169_

---

First implementation is fine but there's still few improvements to do before next release:
- No transparency for the menu
- Save parameters to local cookies
- Columns are reset when filtering (example: go to the albums, then filter by name)
- Columns overlay if screen is too small (improve the display/hide algo to limit the table size to the middle panel content)
- Add a new option to reset the column filtering to the default state
- Maybe some others, need to test it more.

nearwood:
I'm having the same problem as anettles. It doesn't remember my column filter settings. Lemme investigate a bit more and clear out my cache and cookies since I just updated to the new ui branch.

PR #1239, opened 2016-06-06
Fix proposition for issue #123.

Phyks:
This commit introduces a `nT_` function to handle pluralization in
localizations, and avoid the need to call ngettext, which was resulting
in errors if PHP Gettext module was not available (and this was not
enforced by Ampache).

Any occurrence of `ngettext` has been replaced by this `nT_` call.

Also removed the useless keywords in the `gather-messages.sh` script,
and added the newly introduced `nT_` keyword.

Closes issue #123.

### DIFF
--- a/admin/catalog.php
+++ b/admin/catalog.php
@@ -122,12 +122,12 @@ switch ($_REQUEST['action']) {
             foreach ($songs as $song_id) {
                 Song::update_enabled(true, $song_id);
             }
-            $body = count($songs) . ngettext(' Song Enabled', ' Songs Enabled', count($songs));
+            $body = count($songs) . nT_(' Song Enabled', ' Songs Enabled', count($songs));
         } else {
             $body = T_('No Disabled Songs selected');
         }
         $url      = AmpConfig::get('web_path') . '/admin/catalog.php';
-        $title    = count($songs) . ngettext(' Disabled Song Processed', ' Disabled Songs Processed', count($songs));
+        $title    = count($songs) . nT_(' Disabled Song Processed', ' Disabled Songs Processed', count($songs));
         show_confirmation($title,$body,$url);
     break;
     case 'clean_all_catalogs':
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -596,9 +596,9 @@ abstract class Catalog extends database_object
         $hours = $hours % 24;
 
         $time_text = "$days ";
-        $time_text .= ngettext('day', 'days', $days);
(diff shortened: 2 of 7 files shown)


Phyks:
Hmm, I've just updated the PR as I missed some coding standards (trailing whitespaces). Still, the Travis-CI build is not passing, due to the `php-cs` downloaded version being `v2.0.0-alpha` (https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases) which seems to rename `--fixers` to `--rules`. I tried to update the codestyle check script to fit with this new version, without success as replacing `fixers` by `rules` lead to an unknown rule `linefeed`. It seems the doc is not up to date at all on PHP-CS side.

I did run `php-cs` with an older version on my laptop and here is the result:
```
Checking mandatory formatting/coding standards
..............F..................................................................................................................................................................................................FFFF.FFF................................................................................................................................................................................................................................................................................................................
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) lib/ui.lib.php (braces)
   2) modules/horde/Browser.php (align_equals, braces)
   3) modules/captcha/captcha.php (trailing_spaces, visibility, lowercase_constants, concat_with_spaces, align_equals, braces)
   4) modules/oauth/OAuth.php (indentation, trailing_spaces, visibility, elseif, lowercase_constants, concat_with_spaces, align_equals, braces)
   5) modules/Beets/Handler.php (align_equals, braces)
   6) modules/Beets/CliHandler.php (align_equals, braces)
   7) modules/Beets/JsonHandler.php (align_equals, braces)
   8) modules/infotools/AmazonSearchEngine.class.php (visibility, align_equals, braces)
Checked all files in 40.037 seconds, 18.000 MB memory used
Please check code Formatting
php-cs-fixer fix -v --fixers=indentation,linefeed,trailing_spaces,short_tag,braces,controls_spaces,visibility,align_equals,concat_with_spaces,elseif,line_after_namespace,lowercase_constants .
```

There are still some errors which were here before this PR, so this should be ok concerning code style.

Afterster:
The logic is good and you even thought to update shell script, great!
But are you not losing some information with your current `function nT_($msgid)` signature (plurial and counter)?

Phyks:
Indeed, I messed up with my branches, and this is not good at all… sorry.

Should be fixed by now, default behavior being returning the plural string.

Afterster:
Perfect now, thanks :smile:.

Psy-Virus:
As I created the last general translation update, I saw that plurals aren't recognized anymore if I gather the strings from source.

Issue #1213, opened 2016-04-27
Subsonic API Backend: Path Field


daneren2005:
# Environment
 * Ampache version: Ampache 3.8.3-develop (https://github.com/ampache/ampache/commit/66ea7bc403fa31cc353ca01539a15c2bcd5aef74)
 * Web server + version: Nginx
 * Server operating system: Ubuntu 14.04
 * Client operating system: Android
 * Client type:
   - [ ] Web
   - [ ] Ampache
   - [x] SubSonic
   - [ ] Plex
   - [ ] DAAP
   - [ ] UPnP
   - [ ] WebDAV
 * Client application + version: DSub

# Description
The path field isn't actually the path field, it is just the Artist/Album/Song.  The reason this can create issues is if the artist or album tags have slashes in them, you can end up with weird "paths": `Elton/ John/Road/ To El Dorado/01 - El Dorado.flac`. This was obviously a made up example since I didn't have any artists/albums with slashes in them already.  Is there a reason why Ampache isn't putting the actual file path in that param?  If there is, I can work around this by just building the path manually based off the Artist/Album/Title just for Ampache.  I was just curious about why it was setup like this.

PS I came across this via https://github.com/daneren2005/Subsonic/issues/681, but I am pretty sure there problem has nothing to do with this since I was able to play a song from an artist with slashes in it just fine.

Afterster:
thanks for the report @daneren2005.
Indeed, this was excepted behavior, well at least it was coded this way on Ampache side. Before that Ampache was returning the full file path, which is also wrong. I believe the correct value here should be relative path according to catalog root path (which is what is done by legacy Subsonic server I think so).
Gonna need to be fixed then.

daneren2005:
Yes the path is relative to the root of the music folder in Subsonic.

Phyks:
I did an attempt at implementing it in https://github.com/ampache/ampache/pull/1240. Let me know if it does match with the expected behavior.

Phyks:
I think this can be closed now, no?

Afterster:
Yes it can, thank you both.

PR #1240, opened 2016-06-06
Proposition of fix for #1213

Phyks:
Returns path relative to the catalog root path in Subsonic API.

<del>Note that for this I needed to implement the `Catalog::_get_info`
function, which was already called in the code, but was not declared
anywhere. This implementation may not be the best one…</del>

Closes issue #1213.

### DIFF
--- a/lib/class/subsonic_xml_data.class.php
+++ b/lib/class/subsonic_xml_data.class.php
@@ -465,8 +465,8 @@ class Subsonic_XML_Data
         }
         $xsong->addAttribute('suffix', $song->type);
         $xsong->addAttribute('contentType', $song->mime);
-        // Create a clean fake path instead of song real file path to have better offline mode storage on Subsonic clients
-        $path = $artist->name . '/' . $album->name . '/' . basename($song->file);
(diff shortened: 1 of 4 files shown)


Afterster:
Your fix is good thanks for taking care of this issue.

But is `Catalog::_get_info` method really mandatory? Don't you already have `path` member variable already defined at this point without calling this method?

Phyks:
Indeed, I should have it defined already at this point. The call for the `Catalog::_get_info` method was there, but not implemented, so I just implemented it in a symmetrical way with respect to other classes.

I will push a commit to remove this call as it should not be needed at all. Should I still keep the implementation of the `Catalog::_get_info` method?

Phyks:
Ok, so I updated my PR. I removed the call to `_get_info`. Also, I noticed that there was both a `get_info` and a `_get_info` method in `Catalog` class. So I removed the `_get_info` method and updated the other catalogs which were referencing it accordingly.

Afterster:
Looks good for me now, thanks!

Issue #1249, opened 2016-06-12
Implement current XML API in JSON


AshotN:
I think it be great to have the API have a JSON option in addition to XML... or to replace it if people see it fit.

JSON is supported natively in PHP, and in my opinion is a lot easier to work with.

```
$XML = "\t<error code=\"$code\"><![CDATA[$string]]></error>";
$JSON = json_encode(array("error" => array("code" => $code, "message" => $string)));
```

This is an example of the error output in JSON. 

I'm going to work on creating the API in JSON, but in truth I'm not very familiar with the Ampache code base, and would appreciate the help from anyone else. 

https://github.com/AshotN/ampache

Some work I've done towards a JSON API.

Phyks:
I am not a maintainer and may not have a deep enough knowledge about Ampache code base. So, my comment should be taken with care, and it may not be the best way to implement it.

As far as I know, XML API is handled by [this class](https://github.com/ampache/ampache/blob/develop/lib/class/api.class.php) and XML representation is handled by [this one](https://github.com/ampache/ampache/blob/develop/lib/class/xml_data.class.php).

I think you want to rewrite the functions in `xml_data.class.php` using JSON formatting rather than XML one. As it is rather isolated from the rest of the code, I think you can just replace this file by a similar implementation using JSON formatting in a first time, to test it out quickly.

Then, I think it could make sense to have a generic `API_data` class and two derived classes for XML and JSON output (and maybe more in the future if there is a need for it). To implement this, I think it can be worth to have a look at the way catalogs are implemented (see [this base class](https://github.com/ampache/ampache/blob/develop/lib/class/catalog.class.php) and [these implementations](https://github.com/ampache/ampache/tree/develop/modules/catalog)).

Finally, for the implementation of the choice of format on the API side, I think a common way to do it is to use a `format` GET parameter, which would default to `xml` to keep the current behaviour and avoid breaking anything.

Still, there are some design choices to be made, and these are just some ideas. I think it is important to wait for a feedback from @Afterster on this, especially on the implementation design.

Afterster:
This is basically the way to be done yes (as described by @Phyks).
But please don't break API compatibility. JSON should be an option based on application request and XML should remains the default.
Thank you for working on this.

AshotN:
That branch isn't ready ATM to be merged or anything. Just getting the hang of converting the XML stuff to JSON. In time hopefully we can find a way to merge both outputs in harmony

AshotN:
[...]

PR #1252, opened 2016-06-13
Use UTF-8 without BOM for all the files in Ampache

Phyks:
Remove leading character for BOM for some files that were using it. With this PR, all the files in Ampache should be encoded as UTF-8 without BOM.

This is to solve an issue discussed on IRC, which arised in https://github.com/ampache/ampache/issues/1249 resulting in invalid JSON output.


Afterster:
Stupid issue... well done, thanks a lot ;).

PR #1254, opened 2016-06-14
Use cache in PHPCs

Phyks:
Sorry, I forgot to push this in latest PR =(


Afterster:
Yep good little improvement, thanks again.

Issue #1247, opened 2016-06-11
Web API Version Requirement and XML problem


AshotN:
# Environment
 
 * Ampache version: fbbc3bd51736b99f4d56bfa5e6543f4d3b337197
 * Web server + version: Nginx 1.2.1
 * Server operating system: Debian 7
 * Client operating system: Mac OS X Yosemite
 * Client type:
   - [X] Web
   - [ ] Ampache
   - [ ] SubSonic
   - [ ] Plex
   - [ ] DAAP
   - [ ] UPnP
   - [ ] WebDAV
 * Client application + version: Firefox
 
Currently not supplying a version to the web API causes a 

> Login Failed: version too old

In my opinion, the API should just assume latest version, which is better than just completely failing. 

Regardless, another issue is that the XML document returned is blank for this error.

Phyks:
I think your empty XML error is somehow linked with https://github.com/ampache/ampache/issues/1242.

This should be fixed in latest commit, could you check it?

I cannot answer concerning the version, and you should wait for an advice from @Afterster I think.

AshotN:
Nope, I'm on the latest version. I still have the problem.

Afterster:
I agree with you @AshotN, we should assume to use latest version if not specified. Change to be made then ;).

Phyks:
@AshotN Ok, I figured out what the problem with the empty XML document was. It was due to the missing version parameter.

I pushed a PR to make the version parameter optional, and this should solve the empty XML document problem as well: https://github.com/ampache/ampache/pull/1255.

As far as I know, there is nothing to change in the wiki as it was not explicitly stated that the version parameter was required. Let me know if it is not the case.

AshotN:
I'm not home to test. But your code looks like it fixed it.

VanBelkumW:
Same problem, but even with Fix issue #1247 problem still persisted.
The problem is that Plex web API is already at  0.9.17.2.2159-2bd156c.
changing line 179 of plex_xml_data.class.php to a version that Plex now expects > 0.9.12.... get u past this error.
BUT web-dashboard has change so much that additional coding is required... 
Libraries does seem to work with this change…

Phyks:
[...]

PR #1255, opened 2016-06-14
Fix issue #1247

Phyks:
XML API version is now assumed to be the latest available if not
specified as a parameter. Also fix the issue that an empty XML document
was issued when the version was not specified.

Closes issue #1247.


Afterster:
Looks good, thanks ;).

PR #1263, opened 2016-06-15
Fix for #1250

### LINKED ISSUE #1250: Ampache web player won't play ogg files
# Environment
 
 * Ampache version: 3.8.2
 * Web server + version: 2.4.18-2ubuntu3
 * Server operating system: Ubuntu 16.04 LTS
 * Client operating system: Mac OS 10.11.5
 * Client type: Web
 * Client application + version: Firefox 47.0
 
# Settings
[ampache.cfg.php.txt](https://github.com/ampache/ampache/files/310703/ampache.cfg.php.txt)

# Description
When I try to play any ogg file in the web player, it acts like it's starting to play, and then immediately stop/skips the track, never producing any sound, as if the audio track had zero length. MP3 files play just fine.

# Logs
Full Ampache log: [ampache.20160612.log.txt](https://github.com/ampache/ampache/files/310705/ampache.20160612.log.txt)
Ampache log excerpts: 
[trying-to-play-ogg-log.txt](https://github.com/ampache/ampache/files/310706/trying-to-play-ogg-log.txt) 
[playing-mp3-log.txt](https://github.com/ampache/ampache/files/310707/playing-mp3-log.txt)

Corresponding Apache access.log excerpts:
[trying-to-play-ogg-apache-log.txt](https://github.com/ampache/ampache/files/310709/trying-to-play-ogg-apache-log.txt)
[playing-mp3-apache-log.txt](https://github.com/ampache/ampache/files/310710/playing-mp3-apache-log.txt)

Phyks:
Aurora.JS was included, but we were not including the codecs JS files.
Including them fix the impossibility to play OGG vorbis audio files when
HTML5 player is disabled and only Flash player and Aurora.JS are
enabled.

**Important note**: Progression bar was not working even if the file was playing nicely. There might still be an underlying issue, that I did not find yet. I think we should open a new issue to track it and remember to fix it, if you think this change can be merged.

Closes #1250.

### DIFF
--- a/templates/show_html5_player.inc.php
+++ b/templates/show_html5_player.inc.php
@@ -295,7 +295,7 @@ if (AmpConfig::get('song_page_title') && !$is_share) {
 if (AmpConfig::get('webplayer_aurora')) {
     $atypes = array();
     foreach ($supplied as $stype) {
-        if ($stype == 'ogg') {
+        if ($stype == 'ogg' || $stype == 'oga') {
             // Ogg could requires vorbis/opus codecs
             if (!in_array('ogg', $atypes)) {
                 $atypes[] = 'ogg';


Afterster:
Mhh there is already something that try to load only required js codecs, see https://github.com/ampache/ampache/blob/develop/templates/show_html5_player.inc.php#L327
Would be better to fix this delay loading if the issue is here than always including heavy and unnecessary codecs don't you think?

Phyks:
Indeed, I missed it. I will try to fix it and edit my PR.

Phyks:
Fixed. The issue was that ogg files had `$stype` oga instead of expected ogg.

Afterster:
Yep, thanks.

PR #1264, opened 2016-06-15
Fix for issue #1248

### LINKED ISSUE #1248: Prompt user to change secret_key during install.
# Environment
 
 * Ampache version: fbbc3bd51736b99f4d56bfa5e6543f4d3b337197
 * Web server + version: Apache
 * Server operating system: Linux
 * Client operating system: Linux
 * Client type:
   - [ ] Web
   - [x] Ampache
   - [ ] SubSonic
   - [ ] Plex
   - [ ] DAAP
   - [ ] UPnP
   - [ ] WebDAV
 * Client application + version: 
 
# Settings
Describe your settings or attach ampache.cfg.php after removing sensitive information (server host, database connection...)

# Description
Here the bug/feature description and how to reproduce/integrate it.

Ampache should prompt the user to change the secret_key when asking for database info during the web installer.

# Logs
[Ampache logs](https://github.com/ampache/ampache/wiki/Troubleshooting#enable-logging) and web server access/error logs in attached files. Please also add client application logs if appropriate.

Phyks:
Generate `sceret_key` on Ampache installation

Generate a `secret_key` on Ampache installation and put it in the
generated config. This removes the need to do it manually, and the
secret key is ensured to be crypto safe.

Also move the `web_path` configuration setting under a "Various"
fieldset, as it had nothing to do with database settings.

Also added a comment about a missing `break` statement, which was on
purpose, but one could think it was a mistake at first sight.

Closes #1248.

### DIFF
--- a/install.php
+++ b/install.php
@@ -164,6 +164,7 @@ switch ($_REQUEST['action']) {
                 $created_config = $created_config && install_create_config($download);
             }
         }
+        // No break on purpose
     case 'show_create_account':
         $results = parse_ini_file($configfile);
         if (!isset($created_config)) {
--- a/lib/class/core.class.php
+++ b/lib/class/core.class.php
@@ -201,6 +201,32 @@ class Core
     } // form_verify
 
     /**
+     * gen_secure_token
+     *
+     * This generates a cryptographically secure token.
+     * Returns a token of the required bytes length, as a string. Returns false
+     * if it could not generate a cryptographically secure token.
+     */
+    public static function gen_secure_token($length)
+    {
+        $buffer = '';
+        if (function_exists('random_bytes')) {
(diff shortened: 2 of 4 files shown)


Afterster:
Thanks the logic is good to me, but would be better to do it after a check like `if (function_exists('openssl_random_pseudo_bytes'))` as I'm not sure `openssl_random_pseudo_bytes` is available for all installation.

Phyks:
Here is a fixed version. I introduced a function in `Core` to generate cryptographically secure tokens, and made use of it in the installation process. If no cryptographically secure methods are available, we stick with the current behavior.

Afterster:
Good for me now, thanks.

PR #1266, opened 2016-06-16
Apply trim on dropbox parameters

Phyks:
Copying dropbox secret key from the Dropbox UI easily adds extra
whitespaces at the beginning of the string, which leads to invalid
secret key and stack trace. As they are invalid, apply a trim to them
before storing them in database.


Afterster:
Why not, thanks.

PR #1268, opened 2016-06-17
JSON API

AshotN:
XML has had its time, its time for JSON!

*Implementation not complete


lib/class/api.class.php
```diff
@@ -172,7 +175,12 @@ public static function handshake($input)
                     ($timestamp > (time() + 1800))) {
                     debug_event('API', 'Login Failed: timestamp out of range ' . $timestamp . '/' . time(), 1);
                     AmpError::add('api', T_('Login Failed: timestamp out of range'));
-                    echo XML_Data::error('401', T_('Error Invalid Handshake - ') . T_('Login Failed: timestamp out of range'));
+                    if ($outputFormat == 'json') {
+                      echo JSON_Data::error('401', T_('Error Invalid Handshake - ') . T_('Login Failed: timestamp out of range'));
+                    }
+                    elseif ($outputFormat == 'xml') {
```
Phyks: Why not simply using `else`?

lib/class/api.class.php
```diff
@@ -251,25 +264,49 @@ public static function handshake($input)
...
-                    'clean'=>date("c",$row['clean']),
-                    'songs'=>$song['song'],
-                    'albums'=>$album['album'],
-                    'artists'=>$artist['artist'],
-                    'playlists'=>$playlist['playlist'],
-                    'videos'=>$vcounts['video'],
-                    'catalogs'=>$catalog['catalog']));
+                if ($outputFormat == 'json') {
+                  echo json_encode(
+                      array(
```
Phyks: I think you should avoid repeating the array, which is really error prone when editing the array (and then not replicating it in every branch of the if).

Why not using an auxiliary variable to store the resulting array, and just call the right formatting method on it?

lib/class/api.class.php
```diff
@@ -455,7 +501,17 @@ public static function tag($input)
     {
         $uid = scrub_in($input['filter']);
         ob_end_clean();
-        echo XML_Data::tags(array($uid));
+
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+          echo JSON_Data::tags(array($uid));
```
Phyks: Same as above concerning the array.

lib/class/api.class.php
AshotN: Just in case someone wants to add another data type. No real reason.

lib/class/xml_data.class.php
```diff
@@ -221,18 +221,19 @@ private static function playlist_song_tracks_string($song, $playlist_data)
     public static function keyed_array($array,$callback='')
     {
         $string = '';
-
+        return $array;
```
Phyks: Changes in this method should be reverted in the end I guess.

server/xml.server.php
```diff
@@ -34,7 +34,7 @@
 
 /* Set the correct headers */
 header("Content-type: text/xml; charset=" . AmpConfig::get('site_charset'));
-header("Content-Disposition: attachment; filename=information.xml");
+// header("Content-Disposition: attachment; filename=information.xml");
```
Phyks: This should be removed in the final PR I think, no?

lib/class/api.class.php
AshotN: In truth, I got a little frustrated and made a cheap fix until I properly replace the XML function.

But your right I should make a variable and just store the array and use that, smarter than what I did.

server/xml.server.php
```diff
@@ -91,6 +91,7 @@
     // If the method is the same as the action being called
     // Then let's call this function!
     if ($_GET['action'] == $method) {
+        $_GET['format'] = 'xml';
```
Phyks: Do we want to force the format or should we set `$_GET['format']` only if not set by user?

lib/class/api.class.php
Phyks: I would personally rewrite it as:

```
if ($outputFormat == 'json') {
…
} else {  // Any other format, defaults to XML
…
}
```

So that if one adds a new output format, the XML default will be made explicit.

lib/class/api.class.php
Phyks: Ok :)

lib/class/xml_data.class.php
AshotN: I didn't notice that, but yes... we shouldn't merge my xml_data changes if I made any.

server/xml.server.php
AshotN: IMO the header all together should be removed, but the header is commented because it makes it much easier to debug.

lib/class/api.class.php
```diff
@@ -3,7 +3,7 @@
 /**
  *
  * LICENSE: GNU Affero General Public License, version 3 (AGPLv3)
- * Copyright 2001 - 2015 Ampache.org
+ * Copyright 2001 - 2016 Ampache.org
```
Phyks: I think this change is more general than these files. Just commenting to remember to do it.

server/xml.server.php
AshotN: Since the file names are different, its already decided by the user.

If you call xml.server.php you want XML
if you call json.server.php you want JSON

Phyks:
Did a quick review on this. Looks promising!

Did not review the `json_data_class.php` in details as it is still WIP, but looks good :)

lib/class/xml_data.class.php
Phyks: Yes, anyway we will still be able to remove them when the PR is ready and that you will squash your commits. Not really important at the moment. Plus, use whatever makes your debugging easier.

server/xml.server.php
Phyks: I agree this header is really painful to debug. But removing it might break compatibility I think. I do not know enough the XML API to have a clear idea on this. (cc @Afterster)

server/xml.server.php
Phyks: Ok, seems to be ok.

lib/class/api.class.php
AshotN: If this was merged right now, it wouldn't effect any clients. With or without the elseif

server/xml.server.php
AshotN: Ya, I'm not sure of it's purpose. It might have one. Better safe than sorry I guess?

lib/class/api.class.php
Phyks: I mean, using an `else` case for XML makes it the _de facto_ default case. And I find it clearer to understand that we should maintain XML as the default case, written in this way, for further modifications.

AshotN: Sure, but I thought it's easier to understand that XML is happening here. But sure, I'll change them to else

Phyks: I agree… `else` with a comment stating that XML is in use here sounds like a good compromise?

AshotN: Alright, will do tomorrow.

Phyks:
I added the waiting label to draw attention on this, and prevent an accidental merge.

@AshotN let me know whenever you consider it almost ready and want a full review of it :) Thanks a lot for this contribution!

Phyks:
@AshotN How is it going with this PR?

I am willing to make [some improvements](https://github.com/ampache/ampache/issues/1292), and we should see which one should be merged first.

AshotN:
Sorry, my aunt has visited from my home country. Been rather busy lately. Hope to work a bit more on it soon!

lib/class/json_data.class.php
```diff
@@ -694,118 +662,9 @@ public static function timeline($activities)
...
-            default:
-                $footer = "\n</root>\n";
-            break;
-        } // end switch on type
-
-
-        return $footer;
-    } // _footer
-
     public static function podcast(library_item $libitem)
```
AshotN: Need to implement still

AshotN:
We are ready to test, and merge.

Afterster:
Thanks for your work @AshotN :smile:

See my comment on https://github.com/ampache/ampache/issues/1292#issuecomment-231942293
(Git history cleanup will also be required at the end)

lib/class/api.class.php
```diff
@@ -118,6 +118,47 @@ public static function set_filter($filter,$value)
...
+        // Log the attempt
+        debug_event('API', "Destroy Auth Key Attempt, IP:$ip Key:$authKey", 5);
+
+        if(isset($authKey)){
+            if(Session::destroy($authKey)) {
+                if ($outputFormat == 'json') {
+                    echo JSON_Data::single_string('success');
+                }
+                else {  // Defaults to XML
+                    echo XML_Data::single_string('success');
```
Phyks: Why not returning the same kind of structure as in `handshake` method, with `NULL` `session_expires` and this kind of thing?

lib/class/api.class.php
```diff
@@ -313,14 +390,29 @@ public static function artists($input)
...
-        $artists = self::$browse->get_objects();
-        // echo out the resulting xml document
-        ob_end_clean();
-        echo XML_Data::artists($artists);
+        if ($outputFormat == 'json') {
+            // Set the offset
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
+
+            $artists = self::$browse->get_objects();
```
Phyks: This could be moved outside of the test I think.

lib/class/api.class.php
```diff
@@ -313,14 +390,29 @@ public static function artists($input)
...
+            // echo out the resulting xml document
+            ob_end_clean();
```
Phyks: Actually, this also could be moved outside of the test.

lib/class/api.class.php
```diff
@@ -346,11 +438,23 @@ public static function artist_albums($input)
...
-        ob_end_clean();
-        echo XML_Data::albums($albums);
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+        
+        if ($outputFormat == 'json') {
+            // Set the offset
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
+            ob_end_clean();
```
Phyks: Same, outside of the test.

lib/class/api.class.php
```diff
@@ -363,11 +467,23 @@ public static function artist_songs($input)
...
-        ob_end_clean();
-        echo XML_Data::songs($songs);
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+        
+        if ($outputFormat == 'json') {
+            // Set the offset
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
+            ob_end_clean();
```
Phyks: Same

lib/class/api.class.php
```diff
@@ -387,11 +503,23 @@ public static function albums($input)
...
-        ob_end_clean();
-        echo XML_Data::albums($albums);
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+        
+        if ($outputFormat == 'json') {
+            // Set the offset
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
+            ob_end_clean();
```
Phyks: Same. Will not repeat it afterwards.

lib/class/api.class.php
```diff
@@ -402,7 +530,16 @@ public static function albums($input)
     public static function album($input)
     {
         $uid = scrub_in($input['filter']);
-        echo XML_Data::albums(array($uid));
+
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+            echo JSON_Data::albums(array($uid));
```
Phyks: Array declaration should be moved outside of the test and `*_Data::albums` should take a variable argument.

lib/class/api.class.php
```diff
@@ -528,12 +738,23 @@ public static function songs($input)
 
         $songs = self::$browse->get_objects();
 
+        //TODO: Implement the JSON counterpart
```
Phyks: This TODO comment should be removed I guess?

lib/class/api.class.php
```diff
@@ -560,7 +790,16 @@ public static function url_to_song($input)
         // Don't scrub, the function needs her raw and juicy
         $data = Stream_URL::parse($input['url']);
         ob_end_clean();
-        echo XML_Data::songs(array($data['id']));
+
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+            echo JSON_Data::songs(array($data['id']));
```
Phyks: Same about the array. Will not repeat it in the following of the diff, but it still applies.

lib/class/api.class.php
```diff
@@ -707,12 +1039,25 @@ public static function search_songs($input)
...
-        XML_Data::set_offset($input['offset']);
-        XML_Data::set_limit($input['limit']);
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
+
+            $results = Search::run($array);
```
Phyks: Move outside of test case.

lib/class/api.class.php
```diff
@@ -684,12 +1001,27 @@ public static function playlist_remove_song($input)
...
-            echo XML_Data::error('401', T_('Access denied to this playlist.'));
-        } else {
-            $playlist->delete_track_number($track);
-            echo XML_Data::single_string('success');
+
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+          if (!$playlist->has_access()) {
```
Phyks: Test cases should be inverted (first test access, then test output format).

lib/class/api.class.php
```diff
@@ -724,26 +1069,55 @@ public static function advanced_search($input)
...
-        XML_Data::set_limit($input['limit']);
+        //Whatever format the user wants
+        $outputFormat = $input['format'];
+
+        if ($outputFormat == 'json') {
+            JSON_Data::set_offset($input['offset']);
+            JSON_Data::set_limit($input['limit']);
 
-        $results = Search::run($input);
+            $results = Search::run($input);
```
Phyks: Same, most of these lines should go out of the test case, they are not format-specific.

lib/class/api.class.php
```diff
@@ -792,18 +1185,33 @@ public static function localplay($input)
...
+
         switch ($input['command']) {
             case 'next':
             case 'prev':
             case 'play':
             case 'stop':
                 $result_status = $localplay->$input['command']();
-                $xml_array     = array('localplay'=>array('command'=>array($input['command']=>make_bool($result_status))));
-                echo XML_Data::keyed_array($xml_array);
+                $outputArray     = array('localplay'=>array('command'=>array($input['command']=>make_bool($result_status))));
```
Phyks: Readability could be enhanced by splitting on multiple lines.

lib/class/api.class.php
```diff
@@ -792,18 +1185,33 @@ public static function localplay($input)
...
+                if ($outputFormat == 'json') {
+                    echo json_encode($outputArray, JSON_PRETTY_PRINT);
```
Phyks: Just noticed the mix of `json_encode` calls and `JSON_Data` methods calls. This is a bit confusing. I think the methods from `XML_Data` should be replicated in `JSON_Data` and we should call the same methods either from XML or JSON. Of course, most of the time, the `JSON_Data` method will simply be a wrapper around `json_encode`.

lib/class/api.class.php
```diff
@@ -835,36 +1246,74 @@ public static function democratic($input)
                 ));
 
                 // If everything was ok
-                $xml_array = array('action'=>$input['action'],'method'=>$input['method'],'result'=>true);
-                echo XML_Data::keyed_array($xml_array);
+                $outputArray = array('action'=>$input['action'],'method'=>$input['method'],'result'=>true);
```
Phyks: Same, split on multiple lines.

lib/class/api.class.php
```diff
@@ -835,36 +1246,74 @@ public static function democratic($input)
...
+                    }
                 }
 
                 $uid = $democratic->get_uid_from_object_id($media->id,$type);
                 $democratic->remove_vote($uid);
 
                 // Everything was ok
-                $xml_array = array('action'=>$input['action'],'method'=>$input['method'],'result'=>true);
-                echo XML_Data::keyed_array($xml_array);
+                $outputArray = array('action'=>$input['action'],'method'=>$input['method'],'result'=>true);
```
Phyks: Split on multiple lines.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+     * This takes two values, first the key second the string
+     *
+     * @param    string    $key    (description here...)
+     * @param    string    $string    JSON data
+     * @return    string    return JSON
+     */
+    public static function single_string($key, $string='')
+    {
+        if (!empty($string)) {
+            $JSON = json_encode(array($key => $string), JSON_PRETTY_PRINT);
```
Phyks: I think `json_encode` calls in this method should go in a `_encode` private method, which will avoid code duplication and ease change of options such as `JSON_PRETTY_PRINT` across all the API responses.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+     *
+     * This returns the formatted 'tags' string for an JSON document
+     *
+     */
+    private static function tags_string($tags)
+    {
+        $JSON = null;
+
+        if (is_array($tags)) {
+            $atags = array();
```
Phyks: Variable name not very explicit. Maybe this method could use some extra comments.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+            foreach ($tags as $tag_id => $data) {
+                if (array_key_exists($data['id'], $atags)) {
+                    $atags[$data['id']]['count']++;
```
Phyks: Would rather write it as `+= 1`. I would personnally keep `++` notation for increment of loop variables.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+                } else {
+                    $atags[$data['id']] = array('name' => $data['name'],
```
Phyks: Would put the `'name'…` on the next line. Better readable I think.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+    /**
+     * playlist_song_tracks_string
+     *
+     * This returns the formatted 'playlistTrack' string for an JSON document
+     *
+     */
+    private static function playlist_song_tracks_string($song, $playlist_data)
+    {
+        if (empty($playlist_data)) {
+            return "";
```
Phyks: This will be a duplicate with the `return` on line 165, making the flow hard to follow.

I would write
```php
$playlist_track = "";

foreach ($playlist_data as $playlist) {
    if ($playlist["object_id"] == $song->id) {
        $playlist_track = $playlist["track"];
        break;
    }
}
return $playlist_track;
```

with a single exit point from the function.

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+        if (count($tags) > self::$limit or self::$offset > 0) {
+            $tags = array_splice($tags,self::$offset,self::$limit);
+        }
+
+        $JSON = [];
+
+        foreach ($tags as $tag_id) {
+            $tag    = new Tag($tag_id);
+            $counts = $tag->count();
+            array_push($JSON, array("tag" => array(
```
Phyks: Would put `"tag"` on a new line. Matter of tastes :)

lib/class/json_data.class.php
```diff
@@ -0,0 +1,638 @@
...
+    public static function tags($tags)
+    {
+        if (count($tags) > self::$limit or self::$offset > 0) {
+            $tags = array_splice($tags,self::$offset,self::$limit);
+        }
+
+        $JSON = [];
+
+        foreach ($tags as $tag_id) {
+            $tag    = new Tag($tag_id);
```
Phyks: Not sure about the policy for aligning variables declaration in Ampache. Sometimes it is aligned, sometimes it is not. Would need @Afterster input on this.

lib/class/xml_data.class.php
```diff
@@ -77,23 +77,6 @@ public static function set_limit($limit)
...
-    public static function set_type($type)
-    {
-        if (!in_array($type,array('rss','xspf','itunes'))) {
-            return false;
-        }
-
-        self::$type = $type;
-    } // set_type
-
-    /**
```
Phyks: Why was it deleted? Isn't it used? If this is the case, this may be a bug elsewhere.

Phyks:
I did a quick review of this PR and commented a bunch of things :) Most of them are just basic stuff which is quite easy to fix. We'll handle the squashing of commits in the end, it will be easier I think.

Still, there is a lot of code duplication in this API thing (but there used to be already before your PR). I don't like it a lot. As discussed in https://github.com/ampache/ampache/issues/1292#issuecomment-231942293 we should reduce it.

The global outline of a possible way to fix this easily would be to:

1. Add a `serializer` private member to API class.
2. Check for the `json` vs `xml` `format` input parameter. And based on this one, initialize `serializer` with either `JSON_Data` or `XML_Data`. Ideally, there should be an abstract class `API_Data` defining the interface for exposing the API data, and `JSON_Data` and `XML_Data` should inherit from it and implement the methods.
3. Call the same methods on `JSON_Data` and `XML_Data` (which will be `$this->serializer`) to format the output data.

There is also the whole code duplication between `xml.server.php` and `json.server.php` that I totally dislike. Not sure how to fix it in a retro-compatible way. @Afterster do you have any idea? Could we make a major improvement to the API, together with #1292 and break things?

lib/class/api.class.php
AshotN: I found it easier to just use json_encode rather than have a bunch of redundant code

lib/class/json_data.class.php
AshotN: I don't think that was my design

lib/class/json_data.class.php
AshotN: You're right

lib/class/xml_data.class.php
AshotN: Wasn't used anywhere as far I know

lib/class/api.class.php
Phyks: Why would you have redundant code?

lib/class/json_data.class.php
Phyks: Sorry, I just did a whole review of the file, including code that was there before. Not sure if you want to fix them or prefer I handle the fixing of such things?

AshotN:
Sorry for the inactivity, I've been moving homes and been in a state of disarray. My computer isn't working right now, no idea why. So hopefully I will get everything fixed, and catch up!

AshotN:
Can this be merged the way it is currently and then we can work from there on forward? I want the JSON implementation in so I can work on my music application using the newest version of Ampache. Also so I can share the music app.

elohmeier:
👍

lib/class/xml_data.class.php
Afterster: This shouldn't be removed, it is used somewhere :).

Afterster:
There is too much code duplication, merging it with result to unmaintained code and there is enough of this on Ampache already. Don't make me wrong, you did a great job and I definitely want to merge it at the end.

Eclektic:
Hi,
Any chance to see this PR accepted some day ? (after rework of course, I understand the code duplication problem, unfortunately I'm not able to help you on this...)
@AshotN are you still working on it ?

wagnered:
Almpache is being migrated to a laravel based platform.  Laravel has JSON capability built in. As I appear to be the only developer with time to work on it, there is no time frame.

As this PR has Travis-CI build errors and also conflicting files, it would have to be corrected by @AshotN himself or I can close this and a new PR  could be started by someone else who is interested in adding this enhancement.

AshotN:
I simply don't have the time right now due to work/school and other projects, if anyone is willing to pickup wherever I left that would be best, I'll try to be as accommodating as I can to anyone who is willing

lachlan-00:
This is a really out of date pull so i've merged it with current develop in #2050. 
In that branch i should be able to get this running in a few weeks to allow for new methods and formatting.

Thanks for this work @AshotN i'm really appreciative of your efforts back in the day.

PR #1270, opened 2016-06-18
Proposition of rework for issue template

Phyks:
Not sure if it is a better template or not. I tried to rework it a bit,
making use of HTML comments which are interpreted by Github to provide
extra information to the reporter, while not polluting resulting issue.

I also tried to remove the checkboxes templates which was resulting in
"1 out of 7 tasks done" in the issues list and could be misleading.

Let me know what you think about it.


Afterster:
Better resulting thread indeed. Thank you.

PR #1280, opened 2016-06-21
Add DuckDuckGo search engine

### LINKED ISSUE #1279: [FR] Add DDG search support in song and album details.
Add DuckDuckGo search in album and song details.

BoBeR182:
Fixes #1279


Phyks:
Could you squash the two commits please?

Apart from that, the PR looks good. Still, we might have a UI issue here, as if we add DDG, why not add many other search engines?

BoBeR182:
How do I squash commits? As for the UI, DuckDuckGo is privacy respecting unlike Google and has some nice instant results for music artists. https://duckduckgo.com/?q=tschaikowski

Phyks:
See for instance https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/. Note that you will have to force push to your fork.

I totally agree about Duck Duck Go. My latest statement was more to think in the UI, especially since we have some UI issues open, to see if we add this to UI requirements and so on.

Anyways, I cannot merge this PR, just give an advice on it :) You should wait for @Afterster to have a better feedback and a merge.

Afterster:
Thanks for this @BoBeR182 
I will merge it as it was requested several times but as @Phyks said, we should think about the UI next and no additional search engines will be added before.

PR #1287, opened 2016-06-23
General translation update

Psy-Virus:
Translation update as always^^

Modified `lib/general.lib.php` to fit current supported languages and
show more detailed what languages are not listed.

:-)


Afterster:
And thanks as always :)

PR #1288, opened 2016-06-23
Fix issue #955

### LINKED ISSUE #955: test.php isn't translated
Hey ho,
as the title says it, `test.php` isn't translated but the default language is set to German in my case.

And there's a small glitch on the top left.
Are these `bootstrap*.css` automatically generated? Because the formatting looks ugly or .. isn't .. existing. (O.o)
didn't work with it before..

Screeny:

ampache -- debug page 2015-09-17 04-04-02

Thanks in advance :-)

Phyks:
Introduce a "lang" cookie, and use it to translate `test.php`.

Closes #955.


Afterster:
The purpose of session_name is to avoid cookie name conflict.
Looks good to me then.

PR #1289, opened 2016-06-23
DSub not playing music over Subsonic API, attempt to give more infos

Phyks:
The return value from `curl_exec` call in `Subsonic_API::follow_stream`
was not checked, leading to silent error when an incorrect URL was
provided (typically when `local_web_path` was not correctly set).

Also, the use of curl in `follow_stream` function was quite
counter-intuitive and unclear, I added a comment about it to help others
digging into the code.


Afterster:
Better error handling indeed. Thanks.

PR #1310, opened 2016-06-30
add support for unicode in notifications

### LINKED ISSUE #1294: UTF-8 escaping in notifications
Due to a bug, notifications were not translated. This was fixed lately, but still UTF-8 characters are escaped in notifications and render as "u+NUMBER" instead of the character (this is particularly seeable with French translation which contains a lot of accentuated characters).

Niols:
fix #1294

### DIFF
--- a/lib/ui.lib.php
+++ b/lib/ui.lib.php
@@ -713,7 +713,7 @@ function toggle_visible($element)
 function display_notification($message, $timeout = 5000)
 {
     echo "<script type='text/javascript'>";
-    echo "displayNotification('" . addslashes(json_encode($message)) . "', " . $timeout . ");";
+    echo "displayNotification('" . addslashes(json_encode($message, JSON_UNESCAPED_UNICODE)) . "', " . $timeout . ");";
     echo "</script>\n";
 }
 


Phyks:
Nice spot!

lib/ui.lib.php
```diff
@@ -713,7 +713,7 @@ function toggle_visible($element)
 function display_notification($message, $timeout = 5000)
 {
     echo "<script type='text/javascript'>";
-    echo "displayNotification('" . addslashes(json_encode($message)) . "', " . $timeout . ");";
+    echo "displayNotification('" . addslashes(json_encode($message, JSON_UNESCAPED_UNICODE)) . "', " . $timeout . ");";
```
Niols: Note: this constant is PHP≥5.4

Fneufneu: If you don't want to break php < 5.4, you can add something like:
`if defined('JSON_UNESCAPED_UNICODE')`

Phyks: Maybe we should keep the escaped unicode and just replace https://github.com/ampache/ampache/blob/c053d438160817fd80204e4ea15afff07f0e8798/lib/javascript/base.js#L107

by

```
$("#notification-content").html(JSON.stringify(JSON.parse(message)));
```

?

Niols: Support for PHP <5.4 has already been dropped, as some parts of the code use the short array syntax (`[]`). Even Debian oldstable has PHP 5.4.

About where to escape/unescape, we indeed have two solutions. I don't know which one is the best.

Phyks: From the JSON RFC:

> 2.5.  Strings
> 
>    The representation of strings is similar to conventions used in the C
>    family of programming languages.  A string begins and ends with
>    quotation marks.  All Unicode characters may be placed within the
>    quotation marks except for the characters that must be escaped:
>    quotation mark, reverse solidus, and the control characters (U+0000
>    through U+001F).
> 
>    Any character may be escaped.  If the character is in the Basic
>    Multilingual Plane (U+0000 through U+FFFF), then it may be
>    represented as a six-character sequence: a reverse solidus, followed
>    by the lowercase letter u, followed by four hexadecimal digits that
>    encode the character's code point.  The hexadecimal letters A though
>    F can be upper or lowercase.  So, for example, a string containing
>    only a single reverse solidus character may be represented as
>    "\u005C".

So just go for whichever is easier. Not escaping it is already implemented, let's stick with it?

Afterster:
Nice spot indeed. PHP 5.4 requirement is fine to me as it is already the minimum required version for Ampache. Thanks!

Issue #1322, opened 2016-07-08
Database is getting hammered since upgrade


madeye-matt:
# Environment
- Ampache version: 3.8.2
- Web server + version: Apache 2.4.10
- Server operating system: Fedora 20
- Client operating system: Windows 10
- Client type:
  - Web: Chrome  51.0.2704.103  Win 10 (also Firefox Fedora)
# Settings

config_version = 12
web_path = ""
database_hostname = "localhost"
database_name = "ampache_test"
database_username = "ampache"
database_password = "XXXXXXXXXXXXXXXXXX"
session_length = "1800"
stream_length = "7200"
remember_length = "7200"
session_name = "ampache"
session_cookielife = "0"
session_cookiesecure = "0"
auth_methods = "mysql"
catalog_file_pattern = "mp3|mpc|m4p|m4a|mp4|aac|ogg|rm|wma|asf|flac|spx|ra|ape|shn|wv"
catalog_video_pattern = "avi|mpg|flv|m4v"
catalog_prefix_pattern = "The|An|A|Die|Das|Ein|Eine|Les|Le|La"
access_control = "true"
require_session = "true"
require_localnet_session = "true"
downsample_remote = "true"
track_user_ip = "true"
getid3_tag_order = "id3v2,id3v1,vorbiscomment,quicktime,matroska,ape,asf,avi,mpeg,riff"
metadata_order = "getID3,filename"
use_auth = "true"
default_auth_level = "admin"
ratings = "true"
sociable = "true"
art_order = "db,tags,folder,musicbrainz,lastfm,google"
amazon_base_urls = "http://webservices.amazon.com"
max_amazon_results_pages = "1"
debug = "false"
debug_level = "5"
site_charset = "UTF-8"
refresh_limit = "60"
max_bit_rate = "200"
min_bit_rate = "48"
transcode_m4a = "1"
use_rss = "1"
# Description

I have just updated my Ampache server from Ampache 3.6-alpha6+FUTURE to Ampache 3.8 and it is pretty much unusable. 

All operations are taking of the order of 10 times as long as with 3.6.  For example, to simply log in takes 15-20 seconds where it was under a second before.  A music search takes a minute or two where it would take around 1-2 seconds before.  All normal operations are similarly affected. 

While the operation is executing I notice that CPU on mysql goes up to 100% for the duration.   Admittedly it's a sizable collection (~20k songs) but I imagine there are much larger collections.

[...]

PR #1269, opened 2016-06-17
Fix issue #1260

Phyks:
Fix coding guidelines incoherences. Code should match PSR1/2 now, and
php-cs is set to check it on each commit.

Closes #1260.

PR #1323, opened 2016-07-09
Add a cache mechanism on user bandwidths

Niols:
This PR follows what has been said in issue #1322.
It adds a static variable in the User class, that is here to cache the result of bandwidth calculation. The first time a user needs it, it is calculated for all users. And after that, the value can be obtained without a SQL request.

On my instance, it save around 3 or 4 requests per page. According to what @madeye-matt said in issue #1322, it could save around 400 requests per page on large instances.
I believe that this pattern can be applied to more than bandwidth, reducing SQL traffic and speeding up Ampache (on my instance, Ampache is currently doing ~200 SQL request per page… I think that it's way too much).

I'm not really satisfied by this PR, this I don't like the use of static variables in dynamic classes.
However, this is to show that more consideration should be put into the caching of SQL requests.
By the way, I can see in Ampache footer that it counts the number of requests, cached requests and loading time. My cached requests number seems to always be 0. Does that mean that there's a way to add a SQL cache? Or that there used to be one that no longer exists? @Afterster, would you know anything about it?


Phyks:
@Niols it looks like you included a lot of code style changes in this PR making it really large (kind of related to https://github.com/ampache/ampache/pull/1269.

Concerning extending this cache mechanism further, I would personnally say it is also related to the Laravel rewrite as this is something that would be a lot easier and faster to do at this time.

Niols:
Yes, I ran PHP-CS-Fixer. I don't know why develop wasn't up-to-date with the codestyle in wiki.
I can remove this if needed.

About the Laravel rewrite, I totally agree. However, this rewrite isn't ready yet, and wont be for a long time, so I think that we still need to think about it

Phyks:
> Yes, I ran PHP-CS-Fixer. I don't know why develop wasn't up-to-date with the codestyle in wiki.
> I can remove this if needed.

Hmm, codestyle settings have been updated recently, but the whole `develop` branch has not been set up to date with it, as per pull #1269 and related pull.

> About the Laravel rewrite, I totally agree. However, this rewrite isn't ready yet, and wont be for a long time, so I think that we still need to think about it

Sure, my point was just that it could be difficult, time consuming and hacky to implement it at the moment, and that we should take care to not waste too much time on optimizations if we plan a major rewrite, no?

Afterster:
Well if this is small things and it is proposed in such simple PR, I'm not against it. Thanks @Niols.
But this should be done according to user caching setting (`memory_cache`) and using existing caching mechanisms, see https://github.com/ampache/ampache/blob/develop/lib/class/database_object.abstract.php#L111

PR #1326, opened 2016-07-11
activate jplayer "preload" option

Tetram76:
I hope there wasn't good reason to not enable it: woks well with Chrome

### DIFF
--- a/templates/show_html5_player.inc.php
+++ b/templates/show_html5_player.inc.php
@@ -40,7 +40,7 @@ var currentAudioElement = undefined;
                 shuffleTime: 'slow'
             },
             swfPath: "<?php echo AmpConfig::get('web_path'); ?>/lib/vendor/happyworm/jplayer/dist/jplayer",
-            preload: 'none',
+            preload: 'auto',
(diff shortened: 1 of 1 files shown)


Niols:
What is this option for?

Tetram76:
Hi,

Depending on browser, it "should" allow jplayer to preload next file to be played. There're different options that could be interesting to have in user settings, but all I needed was "auto" so :)
http://jplayer.org/latest/developer-guide/#jPlayer-option-preload

My server upload bandwidth is not enough to allow a smooth transition. This option solved the problem.

As I understand, the biggest "risk" is to be ignored by user browser and being useless :)

Niols:
Alright, nice.
Now we need @Afterster's review.
By the way, don't mind the Travis-CI build failing, that's coming from Ampache, and not necessarilly from your PR.

Afterster:
Looks fine to me. For what I can remember it wasn't activated because few issues arise with it, but I cannot remember / reproduce so I believe there is no reason to not enable it now. Thank you for your PR!

Issue #1325, opened 2016-07-11
API Timeline Function


AshotN:
The Web API Timeline function seems to have a problem in it

`<b>Fatal error</b>:  Call to a member function format() on a non-object in <b>/var/www/ampache/lib/class/xml_data.class.php</b> on line <b>676</b><br />`

If I comment out line 676
`$shout->format();`

It works

AshotN:
https://github.com/ampache/ampache/pull/1268/commits/04e1b57604f8d528a5e15465c47b2600120a94cb

I think this was the intent originally.

Phyks:
Why do you have `$activity->shout` in XML but `$activity->action` in JSON?

Niols:
Yes, there are two `$shout->text` and one becomes `$activity->action` and the other `$activity->shout`. Why is that so?

AshotN:
That my friends, is what we call. A mistake. Thank you

AshotN:
https://github.com/ampache/ampache/pull/1268/commits/61178322fc4daf3a2cbc89d602c3b2cbad3d5971

Afterster:
Indeed that's an issue. As JSON API pr will not be merged right now, do you thing you can send a separate PR just to fix this? Thank you.

AshotN:
I'm really bad at Git, and I'm not 100% sure how to :(

I'm going to be trying to figure it out though.

Niols:
Should be fixed now

PR #1327, opened 2016-07-12
Fix Timeline Function

AshotN:
Fixes the timeline function in the XML API.


Niols:
This should fix #1325.

Afterster:
Yep, thanks.

Issue #1286, opened 2016-06-22
Editing a song infos need two clicks


Phyks:
# Environment
- Ampache version: latest develop
- Web server + version:
- Server operating system:
- Client operating system:

- Client type:
  - Web
# Settings

# Description

As noticed by @Kolokial in #1271, we need two clicks to edit songs info in the current UI. The first one does not give any feedback to the user but seems to be loading some things in the background.

Further investigation is needed on this.
# Logs

lachlan-00:
It looks like the image needs to be clicked to load something in the entire table then a second single click works on any row.

I'll try this pull from @AshotN as it links up with tools.js where there links are generated.

AshotN:
I don’t remember exactly what was happening. I haven’t touched the code in years. But I’m here to be of help if within my scope

lachlan-00:
I've pulled your patch into my fork and I'll be looking at it this week. So far it fixes the issue of two single clicks so that's a great start. :)

lachlan-00:
fixed in current core branch

Issue #1236, opened 2016-06-03
Video edit name not working


nasariasl:
afterster: movies have bug !
afterster: i can edit summary,tags,video image but => i cant edit video name !!! <= when i change video name and the clicking on update button nothing happend

wagnered:
@nasariasl,

If, by "video name" you mean  "title",  what version are you using? I
just verified that title change in 3.8.3 works correctly. Please update
to latest development version 3.8.3.

If you are referring to "Original Name" Yes, it is a bug.  I think I
fixed it in the new pull request, but the PR was put on hold because of
the development of Ampache based on a new framework.

73 From  AA1AD
Ernie D
Federal Way, WA | CN87ug

Phyks:
@nasariasl Any news on this?

Also, I think it may be somehow linked to https://github.com/ampache/ampache/issues/1271.

AshotN:
I could not reproduce. The only problem I had was when I clicked on the edit cog, it would load something but nothing would happen. Not until I clicked a second time.

But I was able to rename the title.

AshotN:
I fixed the problem I was having, but I'm such a noob at Git that I've messed some stuff up. Once I fix my Git stuff, I'll push a fix.

mitchray:
Closing due to inactivity, reopen if this is still relevant in latest version

PR #1331, opened 2016-07-13
Cleaned up the showEditDialog Function

AshotN:
This function had some odd code in it, using recursion to fetch data. So I updated it, should fix the problem where you had to click twice on the edit cog for an edit dialog to display.

### DIFF
--- a/lib/javascript/tools.js
+++ b/lib/javascript/tools.js
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
     parent.editType = edit_type;
     parent.editId = edit_id;
 
-    // Convert choices string ("tag0,tag1,tag2,...") to choices array
-    parent.editTagChoices = new Array();
-    if (tag_choices == undefined && tag_choices != '') {
-        // Load tag map
-        $.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_tag_map', {
-            success: function(data) {
-                tag_choices = $(data).find('content').text();
-                if (tag_choices != '') {
-                    showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_row_prefix);
-                }
-            }, type: 'post', dataType: 'xml'
-        });
-        return;
-    }
-	parent.editLabelChoices = new Array();
-    if (label_choices == undefined && label_choices != '') {
-        // Load tag map
-        $.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_labels', {
-            success: function(data) {
-                label_choices = $(data).find('content').text();
-                if (label_choices != '') {
-                    showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_row_prefix);
-                }
-            }, type: 'post', dataType: 'xml'
-        });
-        return;
-    }
-    var splitted = tag_choices.split(',');
-    var i;
-    for (i = 0; i < splitted.length; ++i) {
-        parent.editTagChoices.push($.trim(splitted[i]));
-    }
-	splitted = label_choices.split(',');
-	for (i = 0; i < splitted.length; ++i) {
-        parent.editLabelChoices.push($.trim(splitted[i]));
-    }
+    $.when($.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_tag_map'), $.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_labels')).then(function( a1, a2 ) {
 
-    parent.dialog_buttons = {};
-    this.dialog_buttons[jsSaveTitle] = function () {
-        $.ajax({
-            url: parent.saveUrl,
-            type: 'POST',
-            data: $(parent.editFormId).serializeArray(),
-            success: function (resp) {
-                $('#editdialog').dialog('close');
+        if(a1[2].status != 200 || a2[2].status != 200){
+            //Some sort of error handling?
+            alert("Error");
+        }
+        parent.editTagChoices = [];
+        parent.editLabelChoices = [];
+        tag_choices = $(a1[0]).find('content').text();
+        label_choices = $(a2[0]).find('content').text();
 
-                if (parent.refreshRowPrefix != '') {
-                    var new_id = $.trim(resp.lastChild.textContent);
+        var splitted = tag_choices.split(',');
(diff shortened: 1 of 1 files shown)


AshotN:
Should fix #1286 and maybe #1236?

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
-
-    parent.dialog_buttons = {};
-    this.dialog_buttons[jsSaveTitle] = function () {
-        $.ajax({
-            url: parent.saveUrl,
-            type: 'POST',
-            data: $(parent.editFormId).serializeArray(),
-            success: function (resp) {
-                $('#editdialog').dialog('close');
+    $.when($.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_tag_map'), $.ajax(jsAjaxServer + '/ajax.server.php?page=tag&action=get_labels')).then(function( a1, a2 ) {
```
Phyks: Should be splitted across multiples lines.

Also, `a1` and `a2` are not really clear names. Why not `xhr1` and `xhr2` or `response1` and `response2`?

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
 
-                if (parent.refreshRowPrefix != '') {
-                    var new_id = $.trim(resp.lastChild.textContent);
+        if(a1[2].status != 200 || a2[2].status != 200){
+            //Some sort of error handling?
+            alert("Error");
```
Phyks: Should send a localized error through the standard notification system, if error handling is to be.

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
+        }
+        parent.editTagChoices = [];
+        parent.editLabelChoices = [];
+        tag_choices = $(a1[0]).find('content').text();
```
Phyks: Variables should be defined using the `var` keyword as below.

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
+        label_choices = $(a2[0]).find('content').text();
```
Phyks: Same

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
+
+        var splitted = tag_choices.split(',');
+        var i;
+        for (i = 0; i < splitted.length; ++i) {
```
Phyks: This could be rewritten using an `Array.map` (`$.each` in jQuery if I remember correctly) and defining `parent.editTagChoices` to be `tag_choices`. This is a detail, not sure it is better.

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
+                        }
+                        loadContentPage(reloadp);
                     }
+                },
+                error: function(resp) {
+                    $('#editdialog').dialog('close');
+                }
+            });
+        }
+        parent.dialog_buttons[jsCancelTitle] = function() {
```
Phyks: Why did `this` became `parent`?

lib/javascript/tools.js
```diff
@@ -227,120 +227,105 @@ function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_ro
...
-        close: function (e) {
-            $(this).empty();
-            $(this).dialog('destroy');
-        },
-        buttons: dialog_buttons
-    });
 
-    $('#editdialog').dialog('open');
+        $("#editdialog").dialog("open");
+        });
```
Phyks: Isn't there an indent problem here?

lib/javascript/tools.js
Phyks: Also, this means we are doing the two requests every time we open the dialog box, isn't it?

I mean the
```
if (tag_choices == undefined && tag_choices != '')
```

and
```
if (label_choices == undefined && label_choices != '')
```
have disappeared.

Cannot we replace `$.ajax` with our own `Promise` doing an ajax call if necessary and returning directly otherwise?

Phyks:
Did a quick review. Noticed issues are mostly details. I tested it on my instance and it works great!

AshotN:
When I get a second I'll fix the small problems. Can you show me how error handing should be done though

Phyks:
Typically https://github.com/ampache/ampache/blob/c053d438160817fd80204e4ea15afff07f0e8798/lib/javascript/dynamicpage.js#L46

Thanks

Afterster:
Thanks @AshotN and thanks for the review @Phyks. I agree with your comments, specifically the one about error handling and the way to preserve the ajax call to retrieve tag map. Do you think you can handle it @AshotN ?

lib/javascript/tools.js
AshotN: The ifs were useless

And I'm unsure about that Promise part, care to show me an example.

Phyks: As far as I understand, `tag_choices` is empty at the first run, then, the `if` ensures that we load them from the AJAX call. On the subsequent runs, `tag_choices` is no longer empty and the AJAX call is not done anymore.

I have the impression that we are making an AJAX call systematically with the refactor, whereas we were saving one call with the already existing version. Isn't it?

AshotN:
I haven't been very active in Github of lates. What was needed in this request to be complete?

Afterster:
Me neither @AshotN :smile:. Phyks explained the changes to be done on July 15. You should display "other commented on an outdated diff" by clicking on such links.

AshotN:
How's that looking @Afterster and @Phyks?

Afterster:
Unfortunately @Phyks comments are still valid, specifically about saving Ajax calls (tag_choices and label_choices checks).

AshotN:
I'm not sure I'm understanding where we are making excessive Ajax calls.

lachlan-00:
closing as this was added previously so there's no reason to keep it open.

Issue #1328, opened 2016-07-12
No error message when an album does not exist


Phyks:
# Environment
- Ampache version: latest develop as of today
- Web server + version:
- Server operating system:
- Client operating system:

- Client type:
  - Web
# Settings

# Description

There is no error message when the user browse to a page corresponding to an album which does not exist.

Typically, `https://SERVER/#albums.php?action=show&album=3873` where album `3873` does not exist. Instead, an empty album is shown, which is misleading.
# Logs

lachlan-00:
Closing! This is now in my branch and will merge into core soon.

I've added a debug message and a visual warning when this happens on an empty page.
![image](https://user-images.githubusercontent.com/1305249/59321848-fd4d2880-8d16-11e9-95db-87c8380925e8.png)

Loading a blank album page is confusing and weird. especially because Ampache falls back to Various as the name when blank.

PR #1332, opened 2016-07-13
Missing album handler

AshotN:
I believe this is an appropriate fix for #1328

### DIFF
--- a/albums.php
+++ b/albums.php
@@ -119,8 +119,11 @@ switch ($_REQUEST['action']) {
     default:
         $album = new Album($_REQUEST['album']);
         $album->format();
-
-        if (!count($album->album_suite)) {
+        
+        if($album->id == null){
+            require AmpConfig::get('prefix') . UI::find_template('show_missing_album.inc.php');
+        }
+        elseif (!count($album->album_suite)) {
             require AmpConfig::get('prefix') . UI::find_template('show_album.inc.php');
         } else {
             require AmpConfig::get('prefix') . UI::find_template('show_album_group_disks.inc.php');


Phyks:
I could not test it in my installation at the moment, but the code looks ok.

Phyks:
@AshotN Actually the fix is not doing very well. See attached screenshot:

Phyks:
Maybe the best thing to do would be to return a 404? Not sure if Ampache has customs 404 (does not seem to be the case)

AshotN:
I didn't get such an error. I also assumed that the file I used way already made the way you guys liked it

Phyks:
@AshotN Could you post a screenshot of what you get?

Are you sure the fact you don't have any error is not coming from your PHP install not being really verbose?

AshotN:
screenshot from 2016-08-26 13-06-29

Phyks:
Could you check in db the album 212567 does not actually exist? Could be an album that was added in your db, but is empty? This is really weird…

AshotN:
Its a fresh install. I only have two songs.

Phyks:
Ok, I am sorry, I am lost with this PR at the moment. If I use your branch, I have a PHP error as in the above screenshot.

Your latest screenshot is what I have with the actual up to date `develop` branch of Ampache, prior to applying your patch. This is the behavior described in #1328 which is considered not satisfactory.

Also, we should differentiate "missing" template as in "a known album, with all the infos, but missing in the library" (typically infos are coming from Last.fm) and "missing" album as in "a wrong ID, a total nonsense and no info to display).

AshotN:
So right now, what should we do you think

Phyks:
I am a bit lost with the correct way to handle it at the moment, and don't have much time right now to have a deeper look at it. Maybe @Afterster or @Niols have an insight on this?

I promise I will give a look before the end of the week though and come back with some feedback :)

Phyks:
Sorry for the delay. I gave it a try and I confirm I have the exact same behavior as I had on my screenshot above. An error message:
```
Fatal error: Uncaught Error: Call to a member function show_action_buttons() on null in /srv/http/ampache/templates/show_missing_album.inc.php:56 Stack trace: #0 /srv/http/ampache/albums.php(124): require() #1 {main} thrown in /srv/http/ampache/templates/show_missing_album.inc.php on line 56
```

Without checking out your branch, I have the same behavior as the screenshot you posted, which is the behavior right now, and is not what is expected.

AshotN:
Maybe we should compare our PHP INI files

Phyks:
Mine is pretty standard PHP.ini file, except for `display_errors` set to `on`.

lachlan-00:
Closing as not required now. This fix is in my branch and will merge into core soon.

I've added a debug message and a visual warning when this happens on an empty page.
image

I don't know why Ampache would forcibly load an album when it doesn't exist but it will now error out instead. The reason Various appears is because the template will assume various when null.

Issue #1301, opened 2016-06-29
Localplay improvements


Phyks:
# Environment

 * Ampache version: develop
 * Web server + version:
 * Server operating system:
 * Client operating system:

 * Client type:
   - Web

# Settings

# Description

- [ ] We should add doc in the wiki about localplay setup. Some steps are not clear for a beginner (should we add a catalog etc). I tried to setup a localplay instance at home, I will try to update doc whenever I have something satisfying.

- [ ] Translation is weird. There is "Activate" but "Disable" for a song. Should not it be "Enable" / "Disable"?

# Logs

# Notes

@Psy-Virus who manages translations in Ampache for the above bug. I do not know much of how the translation system works in Ampache =(

Psy-Virus:
I'm sorry again, it seems, my Mail-Filter wasn't filtering posts addressed to me anymore... But this is fixed.

To your question: Where do you see it? on my side, it is enable and disable (the last symbol on song actions like download). Or has it been fixed within the last 21 days?

:-)

Phyks:
I'm sorry, I messed up with the location of "Activate" vs "Disable". It was in the localplay controllers list. You can "activate" a localplay controller. Or "disable" it. And when you disable it, the controller gets "deactivated".

Psy-Virus:
Okay, I updated #1335 to include your suggestion.

And don't be confused by the other commits above here. They're nothing more than a leftover, but GitHub is still showing them...

:-)

Phyks:
Thanks! :)

PR #1309, opened 2016-06-30
Fix typo: s/folowing/following/

gturri:

PR #1335, opened 2016-07-17
General translation update...

Psy-Virus:
...due to changes on Transifex and the re-pluralization.

Some small changes to harmonize the texts and make them more
clear/userfriendly.

partial fix for #1301

:-)


Psy-Virus:
POT file now includes changes of #1309

PR #1338, opened 2016-07-20
Important changes

Psy-Virus:
General Text overhaul, partially because of my changes on the gather script. Also some hits for commands and a reference to me for translation related questions.

:-)


Afterster:
Thanks for the explanations.

PR #1343, opened 2016-07-22
[WIP] Improve API, as discussed in #1292

### LINKED ISSUE #1292: Improved API design
Hi,

I am working with Ampache API lately and here is a list of the issues I encountered so far. This is to discuss and maybe fix them. I try to offer solutions without any breaking change. But we can of course decide to break the API with a new major version.

- [x] Sometimes we use `songs`, sometimes we use `tracks` in the returned fields. Is there a particular semantics for them? I did not find any, and it is not in the wiki.

<del>Proposed solution: Use only `tracks` everywhere, and whenever there used to be a `songs` field, replicate it in a `tracks` field to keep retro-compatibility.</del>

Actually, I think I finally understood it. Song is a music file, track is a song when it is part of an album.

- [ ] HMAC computation is rather complicated and not standard. **Validated**

Proposed solution: Use standard HMAC as https://secure.php.net/manual/fr/function.hash-hmac.php. Keep both ways side by side to keep retro-compatibility.

- [x] No logout from the API **Will be implemented in @AshotN PR**

Proposed solution: Add a `logout` method to destroy a token and ensure it cannot be used anymore.

**Done** in https://github.com/ampache/ampache/pull/1268.

- [ ] Session expiration is not really known. It is sent at connection, but it is not sent again on further queries although it seems it is expanded.

Proposed solution: Always send the `session_expire` key with the latest up to date value, and/or the session lifetime.

- [x] Artists art are not given in `artists` method. **Validated**

Proposed solution: They seem to be available at `image.php?object_id=OD&object_type=TYPE&auth=TOKEN`, why not include it in an `art` field in the `artists` response?

**Done** in https://github.com/ampache/ampache/pull/1343.

- [x] Wiki does not talk about CORS. We should add something on it in the wiki.

**Done** [here](https://github.com/ampache/ampache/wiki/API).

- [x] I will try to udpate the doc with updated examples and stuff.

**Did a bit of lifting for https://github.com/ampache/ampache/wiki/XML-API.**

- [ ] Returned artworks are returned in a wide variety of sizes. Not sure if it is Ampache job to resize them or not, but should be better if art were in a consistent aspect ratio I think. That is a real problem for artists art. For album art, it is mostly ok as they only differ by a few pixels.

Phyks:
* Provide artists art in the `artists` method.
* Add `name` as an alias for `title` attribute for `songs` and stuff.

Closes #1292.

### DIFF
--- a/lib/class/xml_data.class.php
+++ b/lib/class/xml_data.class.php
@@ -306,11 +306,15 @@ class XML_Data
             $rating     = new Rating($artist_id,'artist');
             $tag_string = self::tags_string($artist->tags);
 
+            // Build the Art URL, include session
+            $art_url = AmpConfig::get('web_path') . '/image.php?object_id=' . $artist_id . '&object_type=artist&auth=' . scrub_out($_REQUEST['auth']);
+
             $string .= "<artist id=\"" . $artist->id . "\">\n" .
                     "\t<name><![CDATA[" . $artist->f_full_name . "]]></name>\n" .
                     $tag_string .
                     "\t<albums>" . ($artist->albums ?: 0) . "</albums>\n" .
                     "\t<songs>" . ($artist->songs ?: 0) . "</songs>\n" .
+                    "\t<art><![CDATA[$art_url]]></art>\n" .
                     "\t<preciserating>" . ($rating->get_user_rating() ?: 0) . "</preciserating>\n" .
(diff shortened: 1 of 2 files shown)


Phyks:
* Added a `.htaccess` to enable CORS header.

Afterster:
I agree with proposed changes, thanks!

PR #1346, opened 2016-07-22
Use correct fields type in modules and localplay

### LINKED ISSUE #1339: Password shown in cleartext in localplay instances
# Environment

 * Ampache version: latest dev
 * Web server + version:
 * Server operating system:
 * Client operating system:

 * Client type:
   - Web

# Settings

# Description

Under localplay instances settings tab, the password are shown in cleartext. They should be "***" and an input of type password in the form.

# Logs

Phyks:
I fixed the `types` associated to modules in localplay and catalog and used them to display the correct input, with the correct type, preventing passwords from appearing in cleartext.

I also took profit of the new HTML5 types (int and url) to have a better experience on browsers which support them.

This fixes #1339.

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -283,11 +283,8 @@ abstract class Catalog extends database_object
                         case 'checkbox':
                             echo "<input type='checkbox' name='" . $key . "' value='1' " . (($field['value']) ? 'checked' : '') . "/>";
                             break;
-                        case 'password':
-                            echo "<input type='password' name='" . $key . "' value='" . $field['value'] . "' />";
-                            break;
                         default:
-                            echo "<input type='text' name='" . $key . "' value='" . $field['value'] . "' />";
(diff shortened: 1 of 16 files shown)


Psy-Virus:
@Phyks, does this include a password check field?
Because If you cant see your password input, you can't check, that you didn't misspelled anything...

Phyks:
Indeed, it does not. I could either add a password check or a toggle between password and text type.

Psy-Virus:
If I get it right, these passwords are pre set, right? I think a toggle would be the better idea.

Phyks:
Depends on the page. For edit forms, yes. For adding a new catalog / localplay instance, they are not.

Afterster:
Merging it anyway as it is still better than current code :smile:. Thanks.

Phyks:
Thanks :)

@Psy-Virus Would be better to open a new issue with the expected best behavior please :)

PR #1348, opened 2016-07-25
Basic include support in the XML API

Phyks:
Includes are specified by the `include[]` array `GET` parameter.
Supported includes as of now are:

* For `artists` and `artist` actions, you can include `albums` and/or `songs`.
* For `albums` and `album` actions, you can include `songs`.

For this purpose, I had to edit the `output_xml` method to be able to generate an XML document without any header nor footer.

I just did these four actions as they are the only one I really know at the moment, and needed includes for. I can handle includes for any other action if you think it might be interesting, or I will complete it whenever someone sees a need for it.

Note: @Afterster do you know why the methods listed above are marked as `DEPRECATED` in docstrings?

### DIFF
--- a/lib/class/api.class.php
+++ b/lib/class/api.class.php
@@ -326,7 +326,7 @@ class Api
         $artists = self::$browse->get_objects();
         // echo out the resulting xml document
         ob_end_clean();
-        echo XML_Data::artists($artists);
+        echo XML_Data::artists($artists, $input['include']);
     } // artists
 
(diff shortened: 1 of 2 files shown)


Afterster:
I don't know why it is marked as deprecated, it already was when I took over the project. This deprecation looks invalid to me as there is no alternative...

Thanks for your PR, this will be useful for several people developing around Ampache!

PR #1358, opened 2016-08-03
Minor correction to tab theming for reborn theme

plabadens:
There was a small issue with the light theme: in artist view (among others), the tabs were orange and not blue. This is a very small tweak to override the tab color in light.css and dark.css.


Afterster:
Thanks

Issue #405, opened 2014-08-20
export playlists


mrpurplenz:
Is there any chance playlists could be exported - or more precisely - backed up?
I dont know how many times I (as an average user) have lost playlists due to reinstalling (usually a newer version) of ampache from clean. There are many reasons someone would reinstall ampache to a clean server and to have an importable backup of playlists (and settings) would be just super. 

I do realise some of the difficulties but perhaps this could be a place to start a discussion about potential solutions?

Afterster:
Okay! Finally done as a tool inside /bin folder with commit https://github.com/ampache/ampache/commit/14dcce0cce3e2bb617d03b274de23ad43b318874.
To use it just call `php bin/write_playlists.inc <DIRNAME> <TYPE> [<EXT>]`.

For instance:
 * to export all playlists to xspf files: `php bin/write_playlists.inc /var/local/playlists playlists xspf`
 * to export all albums to m3u files: `php bin/write_playlists.inc /var/local/playlists albums m3u`
 * to export all artists to pls files: `php bin/write_playlists.inc /var/local/playlists artists pls`

roba91:
Hint (could help the next one finding this): On calling `php bin/write_playlists.inc <PATH> playlist m3u` I got

    PHP Warning:  require_once(../lib/init.php): failed to open stream: No such file or directory in /var/www/ampache/bin/write_playlists.inc on line 24
    PHP Fatal error:  require_once(): Failed opening required '../lib/init.php' (include_path='.:/usr/share/php:/usr/share/pear') in /var/www/ampache/bin/write_playlists.inc on line 24

Fixed it by calling `php write_playlists.inc <PATH> playlist m3u` inside of `bin` instead.

xeruf:
Now called via `cli export:playlist` in case anybody still stumbles upon this as I did ^^

PR #1375, opened 2016-09-05
Do not force write_playlists.php to be called from bin folder

Phyks:
As reported in https://github.com/ampache/ampache/issues/405#issuecomment-244550648.

### DIFF
--- a/bin/write_playlists.inc
+++ b/bin/write_playlists.inc
@@ -21,7 +21,10 @@
  */
 
 define('NO_SESSION','1');
-require_once '../lib/init.php';
+
+$path = dirname(__FILE__);
+$prefix = realpath($path . '/../');
+require_once($prefix . '/lib/init.php');
 
 $valid_exts = array ('m3u', 'xspf', 'pls');
 


Phyks:
@roba91 Could you check this matches the error reported [here](https://github.com/ampache/ampache/issues/405#issuecomment-244550648)?

roba91:
@Phyks: How should I check that?

Phyks:
You can check out [this branch](https://github.com/phyks/ampache/tree/Phyks-patch-1) and check that it actually works if not calling from `bin` folder. I am not with my usual computer with dev stuff for Ampache at the moment, and cannot check this modification actually match your issue :/

roba91:
I think it's a overhead of about an hour for me to make it run under similar conditions as my productive system now. I guess it would be much easier for you to do so as soon as you have your dev machine back...

Phyks:
Ok, I'll check.

Afterster:
Okay, finally merged :wink:. Thanks!

PR #1387, opened 2016-09-21
Composer info on Composer column in Itunes

manuelglez86:
i've noticed that the column composer isnt filled so i modify the daap_api.class.php and now is showing the info of composer on its respective column.


Afterster:
Indeed this was missing, good catch and thanks for your PR.

PR #1388, opened 2016-09-23
General translation Update

Psy-Virus:
Nothing special, only the last updates from Transifex and latest code
changes.

:-)


Afterster:
Thanks as usual :).

PR #1399, opened 2016-10-07
Make song comment editable

tdeuling:


Afterster:
Indeed, this was missing. Thanks!

PR #1407, opened 2016-10-17
Solves the Missing session ID showed on the ampache log

manuelglez86:
With this changes now the requests of itunes shows the revision number and the session id.

Solves the Missing session id on the ampache log.

I made changes on the server_info part, based on the server_info response of the forkeed-daapd project.


Afterster:
Mhh ok. It's good to know you're working and testing intensively the DAAP backend. Thanks.

PR #1408, opened 2016-10-17
Add sort by Disk number to all non-default sorts

SenorSmartyPants:
When using a non-default sort of an artists albums disk order is undefined. I was getting albums listed like this

Album A [Disk 2]
Album A [Disk 1]
Album B [Disk 1]
Album B [Disk 2]

Disk will now always sort ASC


Afterster:
I agree, thanks!

PR #1409, opened 2016-10-17
Only create entry for first artist in Artists MB data. Doesn't create…

### LINKED ISSUE #1402: Multiple entries for the same artist
# Environment
- Ampache version: 98094ad2bd9152f68bfbc88a4f6887959cba6ad2
- Web server + version: IIS 8.5 + PHP 7.0.11
- Server operating system:Server 2012 R2

# Description

There are multiple entries for the same artist in my newly imported catalog, for instance.
image

I have tagged all my albums with MusicBrainz Picard, here are a few examples from the Afrojack list.
image
image
When looking these up in Picard the Artist MBID and name are the same.

The Spark on Album Forget the World
image

Dynamite on Album Forget the World
image

Dynamite on Album Hitzone 70
image

Does it have something to do with multiple Artists on one song?

SenorSmartyPants:
… duplicate artist entries. Closes #1402

### DIFF
--- a/lib/class/artist.class.php
+++ b/lib/class/artist.class.php
@@ -671,9 +671,8 @@ class Artist extends database_object implements library_item
         $trimmed = Catalog::trim_featuring($name);
         $name    = $trimmed[0];
 
-        if ($mbid == '') {
-            $mbid = null;
-        }
+        // If Ampache support multiple artists per song one day, we should also handle other artists here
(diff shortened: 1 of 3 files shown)


Afterster:
Better to have a function yep, and this should fix #1402 as you said so it is important. Thank you.

Issue #1389, opened 2016-09-25
Ampache's subsonic api authentication protocol is depricated


mrpurplenz:
# Environment
- Ampache version: 3.8.2-Develop
- Web server + version: Apache/2.4.7 (Ubuntu)
- Server operating system: Ubuntu 14.04
- Client operating system: Ubuntu 16.04

- Client type:
  - SubSonic (mopidy-subsonic via py-sonic) py-sonic version 0.6.1
# Description

I have subsonic clients that work fine on my ampache build. However, When I try to connect with mopidy-subsonic, or more specifically with its subsonic connection library "py-sonic" the connection fails with the following error:
`libsonic.errors.VersionError: Incompatible Subsonic REST protocol version. Server must upgrade.`

I was hoping this issue would be solved by the addition of the `legacyAuth` option in py-sonic. However I guess the ampache version is just too old.

See here for a description from py-sonics end: [https://github.com/crustymonkey/py-sonic/issues/17](https://github.com/crustymonkey/py-sonic/issues/17)

Phyks:
Ampache's Subsonic API implementation is based upon [v1.11.0](https://github.com/ampache/ampache/blob/develop/lib/class/subsonic_xml_data.class.php#L32). That is a quite old version which only support legacy auth, passing hash of password. See [this API doc](http://www.subsonic.org/pages/api.jsp) for extra infos.

So two things here:

* Yes, Ampache should definitely upgrade Subsonic API version to the latest one, this is really starting to be blocking and will only get worse as we will keep diverging from up to date Subsonic API. However, we are very few active maintainers at the moment, and we actively seek community support on such features :/ We are mostly focusing on fixing bugs at the moment. Also https://github.com/ampache/ampache/issues/1352#issuecomment-236358951 is a related issue.

* On the py-sonics (or whatever client) side, they should state which Subsonic version they support. When a user is to use a Subsonic client, at the moment they should check the client supports legacy API version (< 1.13.0 typically). As far as I understand the py-sonics issue, they implemented a legacy auth in their client.

> Technically, the legacyAuth is actually 1.13.0 and earlier.

Then, AFAIK, it should work with Ampache. I know clients working fine with Subsonic in Ampache (I'm actually using Dsub quite a lot). So I would tend to think this is an issue in this lib, rather than Ampache.

mrpurplenz:
mopidy-subsonic uses pysonic. The legacyAuth option can now allow authentication.
[...]

Issue #1352, opened 2016-07-30
Ampache Subsonic backend with MusicBee plugin


PlaidPhantom:
# Environment
- Ampache version: 3.8.3-develop
- Web server + version: Apache 2.4.7
- Server operating system: Ubuntu 14.04.4
- Client operating system: Windows 10

- Client type:
  - SubSonic
# Description

I'm trying to use [MusicBee](http://getmusicbee.com/) and [this plugin](https://github.com/midwan/MB_SubSonic) to connect to Ampache using the Subsonic API.  I set up the API per the [wiki's](https://github.com/ampache/ampache/wiki/API#subsonic-api) instructions, and I'm able to connect to the API using some other apps I've tried, but the MusicBee plugin will not connect.  I worked with the developer in [this issue](https://github.com/midwan/MB_SubSonic/issues/7), and he suggested that Ampache is using an older authentication method and not the current Subsonic standard. Is this something that could be fixed, or is there something I can do to make this work?
# Logs

For what it's worth, I did a cursory scan of my logs and found a few instances of this line in the logs:

2016-07-28 10:54:44 [ampache](subsonic) -> Missing Subsonic base parameters

Phyks:
Indeed, the implementation of Subsonic API in Ampache is based upon version 1.11.0 (AFAIK). See "Authentication" section in `http://www.subsonic.org/pages/api.jsp`.

Passing password in cleartext in URL is mostly ok as long as you use HTTPS. Of course, implementing the updated authentication should be better.

I note this as an enhancement and will try to find sometime to update it. Actually, there is an ongoing discussion about the Ampache-API design https://github.com/ampache/ampache/issues/1292 that is kind of linked to this issue (concerning which future we want for all these APIs, eventually merge some of them).

If you want to give it a try, subsonic API is served from https://github.com/ampache/ampache/tree/develop/rest. 

Subsonic API is implemented in [this file](https://github.com/ampache/ampache/blob/develop/lib/class/subsonic_api.class.php#L298) (link to the ping view example).

For this very issue, the section to edit is [here](https://github.com/ampache/ampache/blob/develop/rest/index.php#L51-L72), putting a switch on the required version.

lachlan-00:
i've started updating the subsonic api in pull #1965 

Right now it's just supporting token auth with user + apikey but that's a start!

PR #1413, opened 2016-10-21
Subsonic API: Added token authentication handling. It just returns an unsupported e…

SenorSmartyPants:
…rror.

I do not see a way to implement this without altering password storage in the database.

This could address issues #1389 #1352

### DIFF
--- a/lib/class/subsonic_xml_data.class.php
+++ b/lib/class/subsonic_xml_data.class.php
@@ -36,6 +36,7 @@ class Subsonic_XML_Data
     const SSERROR_APIVERSION_CLIENT = 20;
     const SSERROR_APIVERSION_SERVER = 30;
     const SSERROR_BADAUTH           = 40;
+    const SSERROR_TOKENAUTHNOTSUPPORTED = 41;
     const SSERROR_UNAUTHORIZED      = 50;
     const SSERROR_TRIAL             = 60;
     const SSERROR_DATA_NOTFOUND     = 70;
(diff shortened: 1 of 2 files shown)


Phyks:
Did you check this is the only change to be 1.13.0 compatible?

I mean, if we merge this, we advertise Ampache to be compatible with Subsonic API v1.13.0, so I think we should implement the whole differences and not only the authentication part, to provide a correct user experience. Else, it is the client's fault (IMO) if he cannot check the version we of the API we offer, and adapt its UX to it.

SenorSmartyPants:
I have not checked for all the features of 1.13.

That's something I want to do but haven't yet.

Phyks:
OK. I think we should wait for this check prior to merging.

SenorSmartyPants:
Checking API support http://www.subsonic.org/pages/api.jsp

API 1.12
Present
getAlbumList2 - musicFolderId parameter present in code
getPlayQueue - but not supported, throws error.
savePlayQueue - but not supported, throws error.

Missing
getSongsByGenre - musicFolderId parameter missing
getStarred - musicFolderId parameter missing. Code comments this method is not supported, but returns XML
getStarred2 - see getStarred
search2 - musicFolderId parameter missing
search3 - just passes thru to search2
createUser - musicFolderId parameter missing. Several other parameters are ignored. Assuming ampache doesn't support those properties
updateUser - entire method is missing (was added in 1.10.1)

SenorSmartyPants:
Checking API support http://www.subsonic.org/pages/api.jsp

API 1.13
Present
getNewestPodcasts

Missing
getTopSongs

SenorSmartyPants:
I pushed an update which removes the API bump to 1.13.0

API published in files now stays at 1.11.0

Now this change can be merged and work can be done separately to add 1.12.0 and 1.13.0 features.

Afterster:
Thanks! Subsonic API evolved a lot since I implemented the backend... I didn't except that ^^.
This is a good temporary fix to have better error handling but #1389 #1352 should be keep in mind.

PR #1419, opened 2016-11-04
General Translation Update

Psy-Virus:
Only Updates from Transifex


Afterster:
Thanks as usual :wink;

Psy-Virus:
:-)

Issue #1421, opened 2016-11-05
Subsonic API does not use full artist names


SenorSmartyPants:
# Environment

 * Ampache version: 3.8.3-develop

 * Client type:
   - SubSonic

# Settings

# Description

# Logs

Subsonic API displays artists without prefix 

Beatles
instead of 
The Beatles

The Beatles should still be displayed in the Bs

lachlan-00:
This is fixed in develop.
i've coalesced the prefix and name in most places while also allowing you to search by
"beatles" or "the beatles"

PR #1422, opened 2016-11-05
Use Artist full name (with prefix) in subsonic xml responses.

SenorSmartyPants:
Add ignoredArticles to xml so they are still sorted properly

Fixes issue #1421


Afterster:
Good fix, thanks.

PR #1438, opened 2016-12-03
[Subsonic API] Added support for similar artist

cvium:
DSub (fork of DSub for Android) supports 'getSimilarSongs', but it only sends an artist id. I find it to be an incredibly useful feature when I'm not wanting a particular album/artist.

My changes are fairly simple and tested. Looking forward to your input. I wasn't too sure how to randomize and slice the list of songs. Right now it randomly sorts the list and returns $count of them. 
Another thing is that I don't include the 'original' artist. What are your thoughts on that?

As for supporting albums, the Last.FM API does not have a "similar album" method, so I'm not too sure how that should be implemented. imo, it should not be supported.

EDIT: I've had a lot of problems setting it up on my Windows machine, so I forgot to run `php-cs-fixer`. Sorry.


Afterster:
This looks fine to me for starting, thanks!
One remark, you may want to check for `lastfm_api_key` before trying to get similar songs otherwise it will fail if undefined.

cvium:
I wonder if that check should be done inside `get_artists_like` and `get_songs_like` rather than `get_similar_songs`.

cvium:
Minor update: I'm super swamped but I plan on finishing this in a month or so. Sorry for the delay.

cvium:
@Afterster Is this what you had in mind?

vicelversa:
Would love to see this feature, any updates?

cvium:
I gave up on ampache tbh. Slow development, slow response and too many little bugs. Maybe it's better now, but I will just continue to use Airsonic even though it's written in Java...

a7medo778:
Likewise, i am on madsonic right now, but the subsonic forks gui is just baaad, thats why i still actively watch ampache

wagnered:
Hello @cvium, @a7medo778, @vicelversa 

> DSub (fork of DSub for Android) supports 'getSimilarSongs',

I have DSub for Android and I don't see a menu item for "Similar songs" And I only saw "Similar Artists on the "Library" tab. Could someone point to where that menu item might be on DSub for Android?
I have tried the patch and don't see any difference in DSub. Of course, If I could find the "Similar Songs" menu, I might be able to. :smile: 

I have seen this issue, but thought it was an enhancement. @Afterster announced earlier that enhancements are sidelined in favor of working on the Laravel version.  In fact, he refused a PR of mine for that reason.  I have been doing my best to respond to possible bug issues in a timely manner.

Also, at any given point in time there seems to be only one developer and the Ampache repository gets a lot of PR's for enhancements, but almost none for bug fixes (read: No one volunteering to help with fixing bugs.)

I have DSub version 5.3.5 installed

vicelversa:
@wagnered I believe this enables the artist radio functionality of dsub to work when using Ampache. On dsub when viewing an artist, pressing the radio icon in the top right will build a playlist of similar songs related to that artist and begin playing the songs. It's a fantastic feature.

wagnered:
@vicelversa Thanks,

> On dsub when viewing an artist, pressing the radio icon in the top right will build a playlist....

I have the patch of @cvium's PR  applied, but I get "Playlist empty" when clicking on the radio icon. I've tried about 6 different artists.  I've tried this with my actual Subsonic server, the Ampache API and the demo subsonic server but all display "Playlist empty".

So, I watched the http/tcp/udp "conversation" between my Ampachi API and Dsub and what I saw when I clicked the "radio" button was a query to a chromecast device, whatever that is. I think it might have something to do with dlna service.

Me thinks that "Similar Songs" has been removed because I searched through the DSub code repository and found no references to similar songs.

cvium:
I use it daily with Airsonic...

https://github.com/daneren2005/Subsonic/blob/master/app/src/main/java/github/daneren2005/dsub/service/RESTMusicService.java#L663

vicelversa:
@wagnered I also applied these patches and it does indeed work for me. There have been random times when the playlist would be empty. I have not investigated those particular cases as it does work most of time.

wagnered:
@cvium Thanks, very much appreciated pointing out that is is in the function `getRandomSongs`.

@vicelversa Thanks for the info.  I'm a "child of the sixties" and that's the era that most of my music is from. maybe that's why I keep getting no results. :smile:.   Dsub isn't very intuitive for me and there is no help that I could find. :frowning:.

I'll also take  your last comment as a confirmation that the patch fixes this Ampache "**bug**". So I'll make sure it's clean and push the changes to the develop branch. @cvium Thanks for the PR.

wagnered:
Added patch to local respsitory, cleaned with php-cs-fixer and pushed changes to github develop branch,.

Thanks all for the feedback.

PR #1447, opened 2016-12-12
ipv6 addresses couldn't be  added to user's ip history.

wagnered:
Added test for ipv4/ipv6 addresses. Added insertion of brackets to the `parse_url` function for IPV6 addresses and removal of brackets for the `inet_pton` function.


Afterster:
A trick that will do it, thanks!

PR #1508, opened 2017-03-13
Prevent Duplicate Playlist Track Numbers

### LINKED ISSUE #1479: TrackNumber not unique when added over API
# Environment

 * Ampache version: Ampache 3.8.3-develop 
 * Web server + version: NGinx
 * Server operating system: Debian
 * Client operating system: Linux Mint

 * Client type:
   - Web
   - Ampache

# Description

It is not completely clear to me why, but whenever I use the playlist_add_song option in the API some songs end up having the same TrackNumber. This is really problematic for me since the only way to remove songs from a playlist via the API is by using the TrackNumber. Once I load the playlist with the regular web browser, it auto-sorts the tracks and the problem is gone. 

### Some things I noticed not sure if they are relevant

Sometimes I add a song that exists in one playlist to another playlist... it copies the TrackNumber

For example

I have two Playlists, Playlist A and Playlist B

If Track X is in Playlist A with a TrackNumber of 11
I add Track X to Playlist B and the TrackNumber is identical

I took a look, I didn't fully understand the mechanism. If someone is able to reproduce I will see if I can fix it myself

[The Mechanism](https://github.com/ampache/ampache/blob/624a2f0717395443dfdd5a5c4594f8512ae5bd65/lib/class/playlist.class.php#L388 )

AshotN:
Fixes #1479

### DIFF
--- a/lib/class/playlist.class.php
+++ b/lib/class/playlist.class.php
@@ -403,7 +403,11 @@ class Playlist extends playlist_object
 
             // Based on the ordered prop we use track + base or just $i++
             if (!$ordered && $data['object_type'] == 'song') {
-                $track    = $media->track + $base_track;
+                $mediaTrack = $media->track;
+                if($mediaTrack == 0) {
+                    $mediaTrack = 1;
+                }
+                $track    = $mediaTrack + $base_track;
             } else {
                 $i++;
                 $track = $base_track + $i;


manuelglez86:
Im fixed this too, but i didnt make the changes on playlist.class.php, in order to only affect the behavior on the api i just edit the api.class.php you can see PR #1564

AshotN:
I made this change a long time ago and I'm not really sure why I did it the way I did,but perhaps this fixes it on a lower level. I'm not sure, if someone who understands the code base better could weigh in, that would be best.

lachlan-00:
Closing, the bug fix was made in #1564

PR #1548, opened 2017-05-28
Seafile Catalog Module

PlaidPhantom:
Hi, I needed a way to connect Ampache to my [Seafile](https://www.seafile.com/) file server, so despite not really being a PHP developer I decided to write my own.  I based it initially off the existing Dropbox module and used [this Seafile API client](https://github.com/rene-s/Seafile-PHP-SDK). I've Added, Verified, and Cleaned multiple times and everything seems to work. Well, the "gather art" process tries to treat my fake file paths as  local paths and displays failures, however it still can gather from searches, etc.

I did make one change to the `Catalog` class itself: I wanted to be able to use the "Update from tags" feature as I was organizing my library and fixing bad tags. The Catalog class's `update_media_from_tags` function was assuming the files were local (and it being static there was no way to override), so I pulled much of that function into a member function so I could override it in my catalog module.

Suggestions, requests, etc. are of course welcome.

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -1536,21 +1536,11 @@ abstract class Catalog extends database_object
      */
     public static function update_media_from_tags($media, $gather_types = array('music'), $sort_pattern='', $rename_pattern='')
     {
-        // Check for patterns
-        if (!$sort_pattern or !$rename_pattern) {
-            $catalog        = Catalog::create_from_id($media->catalog);
-            $sort_pattern   = $catalog->sort_pattern;
-            $rename_pattern = $catalog->rename_pattern;
-        }
-
         debug_event('tag-read', 'Reading tags from ' . $media->file, 5);
 
-        $vainfo = new vainfo($media->file, $gather_types, '', '', '', $sort_pattern, $rename_pattern);
-        $vainfo->get_info();
+        $catalog        = Catalog::create_from_id($media->catalog);
 
-        $key = vainfo::get_tag_type($vainfo->tags);
-
-        $results = vainfo::clean_tag_info($vainfo->tags, $key, $media->file);
+        $results = $catalog->get_media_tags($media, $gather_types, $sort_pattern, $rename_pattern);
 
         // Figure out what type of object this is and call the right
         // function, giving it the stuff we've figured out above
@@ -1764,6 +1754,25 @@ abstract class Catalog extends database_object
             $field = $libraryItem->getField($tag);
             $libraryItem->addMetadata($field, $value);
         }
+    }
+
+
+    public function get_media_tags($media, $gather_types, $sort_pattern, $rename_pattern)
+    {
+        // Check for patterns
+        if (!$sort_pattern or !$rename_pattern) {
+            $sort_pattern   = $this->sort_pattern;
+            $rename_pattern = $this->rename_pattern;
+        }
+
+        $vainfo = new vainfo($media->file, $gather_types, '', '', '', $sort_pattern, $rename_pattern);
+        $vainfo->get_info();
+
+        $key = vainfo::get_tag_type($vainfo->tags);
+
+        $results = vainfo::clean_tag_info($vainfo->tags, $key, $media->file);
+
+        return $results;
     }
 
     /**
--- a/modules/catalog/seafile/SeafileAdapter.php
+++ b/modules/catalog/seafile/SeafileAdapter.php
@@ -1,0 +1,277 @@
+<?php
+/* vim:set softtabstop=4 shiftwidth=4 expandtab: */
+/**
+ *
+ * LICENSE: GNU Affero General Public License, version 3 (AGPLv3)
+ * Copyright 2001 - 2016 Ampache.org
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+use Seafile\Client\Http\Client;
+use Seafile\Client\Resource\Library;
+use Seafile\Client\Resource\Directory;
+use Seafile\Client\Resource\File;
+use GuzzleHttp\Exception\ClientException;
+
+class SeafileAdapter
+{
+    // request API key from Seafile Server based on username and password
+    public static function request_api_key($server_uri, $username, $password)
+    {
+        $options = array(
+            'http' => array(
+                'header' => "Content-type: application/x-www-form-urlencoded\r\n",
+                'method' => 'POST',
+                'content' => http_build_query(array('username' => $username, 'password' => $password))
+            )
+        );
+
+        $result  = file_get_contents($server_uri . '/api2/auth-token/', false, stream_context_create($options));
+
+        if (!$result) {
+            throw new Exception(T_("Could not authenticate with Seafile"));
+        }
+
+        $token = json_decode($result);
+
+        return $token->token;
+    }
+
+    /////////////////////////
+    // instance
+    /////////////////////////
+
+    private $server;
+    private $api_key;
+    private $library_name;
+    private $call_delay;
+
+    private $client;
+    private $library;
+
+    private $directory_cache;
+
+    public function __construct($server_uri, $library_name, $call_delay, $api_key)
+    {
+        $this->server          = $server_uri;
+        $this->library_name    = $library_name;
+        $this->api_key         = $api_key;
+        $this->call_delay      = $call_delay;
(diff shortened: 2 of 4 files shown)


PlaidPhantom:
I've cleaned up the commit history to make things a little easier to parse.

wagnered:
@PlaidPhantom ,

I took a cursory look at the changes, and looks ok for what you got. I have a few questions though.

How is the seafile catalog created?
How is the module activated?  It needs to have an entry on the  "modules/catalog modules" page to be enabled/disabled.  After it is activated it should display on the "Add a Catalog" catalog type drop down list.

"...the "gather art" process tries to treat my fake file paths as local paths and displays failures"

What do you mean by "fake file paths? The "gather art" function should not fail. How are the files accessed on the remote seafile server? by a full URI? A command via a remote API?

PlaidPhantom:
> How is the seafile catalog created?

I'm not sure exactly what your asking. Do you mean how a new catalog is initialized? From what I can tell, in the "Add Catalog" form, when you select "seafile" as the catalog type the form should refresh and add form fields as defined in that catalog class's [`catalog_fields()`](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/seafile/seafile.catalog.php#L125). (In my case it's the Seafile Server URL, Library Name, and an API call delay to get around Seafile's rate limiting.)  The API key gets created when `show_ready_process()` gets called; I don't offhand recall exactly when Ampache calls that; IIRC it might be during the first catalog operation, i.e. the first "Add". This was the way the Dropbox API worked, so I followed the same pattern.

All that said, the "Add a Catalog" form in my install currently seems to throw an exception on a missing javascript function called `catalogTypeChanged`.  I know that didn't happen when I initialized the catalog during development, so I don't know if it's something I did or if something else has changed in the interim; I'll try to look into it.

> How is the module activated?

Other than the actual activation of the module, this seems to all be handled by Ampache itself.  I don't know how exactly it scans for new modules, but when I added the folder, class, and `composer.module.json` file, the module simply appeared in the catalog module list and could be activated as you describe.  One thing I will note is that existing installations may have to run `composer update` to get the Seafile API library I'm using.

> What do you mean by "fake file paths?

When I generate the metadata for a song, I create a "virtual" path string [here](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/seafile/seafile.catalog.php#L523) in the format [LibraryName|/path/to/file|filename](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/seafile/seafile.catalog.php#L334). Files are retrieved from Seafile by (1) requesting an [apparently temporary](https://manual.seafile.com/develop/web_api_v2.1.html#download-file) download URL and (2) downloading the file via a GET request. You can see this in action in [`download_metadata()`](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/seafile/seafile.catalog.php#L490) and [`prepare_media()`](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/seafile/seafile.catalog.php#L735). I just followed the [Dropbox Catalog's example](https://github.com/PlaidPhantom/ampache/blob/develop/modules/catalog/dropbox/dropbox.catalog.php#L473); I'm open to other recommendations.

Looking through the [Art class `gather` function](https://github.com/PlaidPhantom/ampache/blob/develop/lib/class/art.class.php#L1008), assuming the `art_order` config value is left at the [default value](https://github.com/PlaidPhantom/ampache/blob/develop/config/ampache.cfg.php.dist#L496) ("db,tags,folder,musicbrainz,lastfm,google"), then `gather_tags()` will eventually call to [`here`](https://github.com/PlaidPhantom/ampache/blob/develop/lib/class/art.class.php#L1456), passing my "virtual" filename as if it's a local path. The same sort of thing happens in [`gather_folder()`](https://github.com/PlaidPhantom/ampache/blob/develop/lib/class/art.class.php#L1293).

My gut reaction would be a similar solution as the Catalog class's [`update_media_from_tags`], creating overrideable methods on the Catalog to allow catalog modules to "hook into" these operations.

wagnered:
@PlaidPhantom 

I meant "activated" (I guess that was redundant). it needs to be activated/disabled on the Modules=>Catalog Modules page before it can show up on the "catalog type" list when adding a new catalog. Ampache will only scan for "activated" catalog modules. Also, any uri, api key, username, etc should be able to be entered during creation of the catalog similar to what happens with the Beets catalog. 

The current version of Ampache is very unorganized as you may have noticed.  Its development began before the advent of frameworks.  If it may help, I use Eclipse-pdt to step through the code to get an understanding of the logic.

Also, the info in composer.module.json must be added to composer.json located in the root folder.  see the composer entry See the example for "Afterster/php-echonest-api" in composer.json.All development versions of packages must have an entry in the "repositories" section unless it has been published in the "packagist" or other public package repository.

Before you make too many commits, please do a git merge squash on what you have already. It gets very ugly and confusion when there are very many commts in the PR. and please refrain from doing more commits until the changes have been fully exercised. 

Also, if you haven't noticed, the one file you just committed faile the codacy tests.

 Keep in mind that you are developing this PR for the public, not just for yourself. It is necessary to think as an administrator while doing the development.

PlaidPhantom:
> I meant "activated" [...] Ampache will only scan for "activated" catalog modules

It just gets activated in Modules/Catalog Modules like you said. That list comes from [`Catalog::get_catalog_types()`](https://github.com/ampache/ampache/blob/develop/lib/class/catalog.class.php#L313) scanning the modules/catalog directory for directories and ".class.php" files.

> Also, any uri, api key, username, etc should be able to be entered during creation of the catalog similar to what happens with the Beets catalog.

Okay, I've moved the API key retrieval to catalog creation instead of using the ready process. That was ugly UI anyway. ;)

> Also, the info in composer.module.json must be added to composer.json located in the root folder.

If I'm understanding right, the ["extra" section](https://github.com/ampache/ampache/blob/develop/composer.json#L362) in composer.json should account for including all catalogs' composer requirements. Is that not correct? Is this pattern deprecated?

> Before you make too many commits, please do a git merge squash on what you have already.

Okay, I'll make sure the current state (and any future changes) stay squashed to a single commit.

> Also, if you haven't noticed, the one file you just committed faile the codacy tests.

The PR was passing all tests before, and as I'm writing this the codacy test has been removed...? Let me know what you'd like further changed. I kept to one class/file following the Dropbox example, but I can break it out more if desired.

PlaidPhantom:
okay, the codacy test is back, so I'll work on that.

wagnered:
@PlaidPhantom 

From my previous statement:  ...it needs to be activated/disabled on the Modules=>Catalog Modules before it can show up on the "catalog type" list.

Never mind.  I took the liberty of cloning your repository and installing it.  I didn't realize that Ampache scans the modules/catalogs/* folders, gets a list of existing modules and then checks install state before adding them to the "Catalog Types" list on the Modules=>Catalog Modules page. 

Ampache's code is so unorganized and complex, I am still learning to "untangle" it.

> If I'm understanding right, the "extra" section in composer.json should account for including all catalogs

I actually did not notice that section. I suppose it is because none of the existing modules contain  a composer.module.json file. That's probably because the catalog modules already exist in the modules folders, so the .json files are not necessary.

Looking at the "Add Catalog" page, I have a question or two:

> After creating the catalog, you must 'Make it ready' on the catalog table.

Since I can't follow through and create a seafile catalog, how is this done? What catalog table?  Will this be on the actions list on the "show catalog" page?

The PR for this should not be too complex and should not need "too many" commits.

wagnered:
@PlaidPhantom 

Let me know if it's ready. and I'll do the merge.

PlaidPhantom:
I'm currently testing some refactoring motivated by the Codacy test; I found a few opportunities for optimization & code reuse while I was at it. (Though now the test passes...I guess the test parameters changed again. ;) )

> Since I can't follow through and create a seafile catalog, how is this done? What catalog table? Will this be on the actions list on the "show catalog" page?

Per your suggestions earlier, I changed the initialization process so this "make ready" step will no longer be necessary; the username and password will be entered in the "Add Catalog" form. For some reason I originally didn't think to just put the username and password in the `catalog_fields` function.

For background: the "Catalog Table" is the list of catalogs on the Admin/Show Catalogs page. If a catalog reports as "not ready" (via the `isReady` function), a link shows _above_ the actions list to "make it ready" and the catalog has an opportunity to show an additional configuration form. The catalog handles the ready process via the `show_ready_process`, and `perform_ready` functions; you can see how I originally implemented it [here](https://github.com/PlaidPhantom/ampache/blob/e7b79f17e4bd035f5c55b38f3b561c9147dbd9b3/modules/catalog/seafile/seafile.catalog.php#L139).

PlaidPhantom:
@wagnered 

I've updated the PR. I broke out a lot of the Seafile API integration into a separate class to make the code a bit cleaner.

PlaidPhantom:
@wagnered ok it's passing all tests now  so whenever you're ready for it.

wagnered:
@PlaidPhantom 

By the way, you own it.   :smiley:   That is to say, You are the problem solver if anyone has a problem.

Thanks,

Ernie D

PlaidPhantom:
@wagnered Ha, sure. just message me if I don't see something. 😃

wagnered:
@PlaidPhantom 
Support suggested that the changes made to their build system may be the cause and wanted us to try a modification to the .travis.yml file.

Could you save the following link as [travis.patch](https://github.com/ampache/ampache/commit/5152609d1b721c0f79f149c376d080c106f95eb7.diff) to the root folder , run it and push the change?
command:  `patch .travis.yml travis.patch`

If you gave me push permissions to your github repository I could have saved you the trouble.

Thanks...

PR #1549, opened 2017-05-30
Read replaygain tags for all supported files and fix gain calculation

nightroan:
My original goal was to fix replaygain for flac files tagged with vorbis comments. It turns out getID3 provided replaygain information, so I removed existing special replaygain handling and replaced it with generic code.

After getting replaygain to be read properly, I noticed that songs with negative replaygain became louder when replaygain was enabled. I'm not 100% certain with my fix for this one, but it makes sense to me after reading through [this article](http://wiki.hydrogenaud.io/index.php?title=ReplayGain_1.0_specification)


mitchray:
Closely related to #2681, I'll double check this gain adjustment as it could affect the R128 calculation

PR #1554, opened 2017-06-12
JS autofocus on login and lost password pages replaced by au HTML5 attribute

Simounet:
Hi there,
I propose to replace the autofocus JS handling by a standard one. When the page take some time to load and we already filled the login and typing the password, the focus get back to the login. With this trick, it didn't.
Thanks for this amazing piece of software, have a nice day.

### DIFF
--- a/templates/show_login_form.inc.php
+++ b/templates/show_login_form.inc.php
@@ -43,12 +43,9 @@ define('TABLE_RENDERED', 1);
         <meta http-equiv="Content-Type" content="text/html; charset=<?php echo AmpConfig::get('site_charset'); ?>" />
         <?php require_once AmpConfig::get('prefix') . UI::find_template('stylesheets.inc.php'); ?>
         <title> <?php echo AmpConfig::get('site_title'); ?> </title>
-        <script type="text/javascript" language="javascript">
-            function focus(){ document.login.username.focus(); }
-        </script>
     </head>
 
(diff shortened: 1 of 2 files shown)


Simounet:
In the worst case scenario, there will be no autofocus so the user will have to click on the input field. The good side of these changes is that the browser understand what you want to and if the JS is not fully loaded and you started to type your username and password, you won't be "misfocused" to the username field where you were typing your password (as plain text) in front of other people.

This is a common feature included from Firefox 4, Chrome 5, Safari 5, Opera 10 and Internet Explorer 10. The last one was the last desktop browser to ship it on 2012 ([informations from caniuse.com](http://caniuse.com/#search=autofocus)).

I think this can be used safely.

wagnered:
@Simounet 
OK, but first, please download/clone the latest develop (3.8.4-develop) and run your changes through php-cs-fixer.  I updated the scripts and configure file to work again.  Then do a new pull request. Your code should then pass the Travis-cl build.

Simounet:
My PR is now up-to-date with the develop branch but the 5.6 PHP Travis
checker doesn't seems to be working.

wagnered:
Must have been a glitch on the Travis side.  I restarted the php 5.6 test and it passed. 

I think you have the rights to restart the test also.  On the right side is a very light smiley face that restarts the test.

Thanks...

PR #1559, opened 2017-06-23
General Translation Update

Psy-Virus:
All the latest updates from Transifex.
And added new language "Indonesian" to `lib/general.lib.php`.

Hey @wagnered,
it would be nice, if you push this to the master branch as well, so we have the latest translations in the stable too. If it's possible... ^^
If it's not possible I'll create a new PR for the master branch.

Thank you!

:-)


wagnered:
Thanks @Psy-Virus ,

Changes merged into both develop and master.

Issue #1516, opened 2017-03-31
malformed xml return data for songs action


phirestalker:
# Environment

 * Installed Ampache version: cd761ccc3b997ee0e219c81dd6cdf2668e70f734.
 * Apache/2.4.25 (Unix)
 * osx 10.12.3 client and server

 * Client type:
   - Web
   - Ampache

# Settings

# Description

this line isn't right. `<total_count>6640</total_count>\n<song id="5075">`

there should be an actual newline instead of the newline special character

lachlan-00:
Closing, just tested and fixed in 4.0.0-develop
![image](https://user-images.githubusercontent.com/1305249/59897541-72fc7700-9430-11e9-8db6-052c9cc75006.png)

PR #1571, opened 2017-07-10
Showing album artist on XML_API

manuelglez86:
Hi.

Ill try to explain this changes, if i have the album 18 Months of Calvin Harris and we have songs like Clavin harris ft Ellie Goulding, ampache gives an id for everyone so Calvin harris is an artist and Calvin Harris ft Ellie Goulding is another artist.

in that cases where the songs on an album have more than one artsits the api shows:

18 Months - Various, instead of 18 Months - Calvin Harris.

So with this changes now the api test if there is Album Artist on id3 and display it, if album artist is not present (usually when is just one artist on all the songs of the album) then it displays the artist tag on id3.

let me know what you think of this changes.


manuelglez86:
i have also fixed #1516, let me know you suggestions.

lachlan-00:
could you also fall back to various if album artist isn't found?
```
            if ($album->album_artist_name != "") {
                $string .= "\t<artist id=\"$album->artist_id\"><![CDATA[$album->album_artist_name]]></artist>\n";
            } elseif ($album->artist_count != 1) {
                $string .= "\t<artist id=\"0\"><![CDATA[Various]]></artist>\n";
            } else {
                $string .= "\t<artist id=\"$album->artist_id\"><![CDATA[$album->artist_name]]></artist>\n";
            }
```

PR #1593, opened 2017-08-23
Composer tag changes are not written on database

manuelglez86:
When i verify a catalog and a song has changed its composer tag, ampache detects it but didnt update the database, so i make this changes and now is working.

i have little knowledge about sql, so please review the changes.

Thankyou.


wagnered:
@manuelglez86 

Thanks.....

PR #1594, opened 2017-08-24
Multiple Disks, single Album

AshotN:
Fixes multiple Disks from being treated as seperate albums

### DIFF
--- a/lib/class/album.class.php
+++ b/lib/class/album.class.php
@@ -414,8 +414,8 @@ class Album extends database_object implements library_item
             return self::$_mapcache[$name][$disk][$mbid][$album_artist];
         }
 
-        $sql    = 'SELECT `album`.`id` FROM `album` WHERE `album`.`name` = ? AND `album`.`disk` = ? ';
-        $params = array($name, $disk);
+        $sql    = 'SELECT `album`.`id` FROM `album` WHERE `album`.`name` = ?';
+        $params = array($name);
 
         if ($mbid) {
             $sql .= 'AND `album`.`mbid` = ? ';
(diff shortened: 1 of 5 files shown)


wagnered:
Did you "use the command line as was advised in the comment "This branch has conflicts that must be resolved"?

if you do a `git status` you should see 

> 
...
	both modified:   lib/class/api.class.php
	both modified:   lib/class/xml_data.class.php
	both modified:   server/xml.server.php

This reveals the files that have the conflicts and have been modified to show the conflicts.

lachlan-00:
Just a question. Having an mbid in your tags for the album will join them for me.

Mbid is the same for all disks in a release id so this sort of exists already. Does this conflict with those albums?

It could be possible to link albums in the same way if there's no mbid using composer, disk number and title?

wagnered:
@AshotN 

It's been awhile, I know, but this is an "enhancement" request.

In lib/class/update.class.php it looks like you are modifying an existing update.  This is a no-no unless there is an actual problem with that specific database update.   You need to create a new function update_380013 at the end of the file.  Then enter the description at the end of the populate_version() function similar to how the other updates are entered. That should remove the merge conflict in this file.

As for the merge conflict of lib/class/song.class.php, I can remove that when the time comes. There is an unrelated bug I noticed that needs to be fixed anyway. That I think will fix the Codacy test failure.

lachlan-00:
This is not required now due to updates i've made in the core branch. I could do some extra stuff to display how many cd's are part of the album but that isn't really an issue as grouping works now everywhere.

I also thought about changing the structure of the database but it's not really important as we already have the ability to group all this information without changing the structure.

Now in core all you have to do is:
* Enable Group Multiple disks

After changing that setting Search/display/etc will only show 1 album id grouped by "name", "album_artist" and "mbid".
image

Thank you once again @AshotN I've really enjoyed looking over all your code changes and making Ampache better. :)

PR #1610, opened 2017-09-27
add global options to customize waveform size

Tetram76:
I found the waveform default size a bit small to really see the audio file wave form.

So I added two new options: "waveform_height" and "waveform_width" to permit the admin choose which size he wants.

The only problem I see is the memory consumption: on my conf, I first set waveform_height to 96 but it exceeded the php allocated memory. 
So I use 64 as a compromise.


Tetram76:
sorry, I just see the commits were done with my work account.
I'll do a new pull request.

PR #1611, opened 2017-09-27
add global options to customize waveform size

Tetram76:
I found the waveform default size a bit small to really see the audio file wave form.

So I added two new options: "waveform_height" and "waveform_width" to permit the admin choose which size he wants.

The only problem I see is the memory consumption: on my conf, I first set waveform_height to 96 but it exceeded the php allocated memory.
So I use 64 as a compromise.


Psy-Virus:
Okay, this would be safe to merge. IF you would clean up the commit history. I know it's a bit late but would you be so kind?
Only the first commit is relevant for this PR. Even if the others won't change anything, they mess up Ampache's History.

PR #1617, opened 2017-10-12
General Translation Update

Psy-Virus:
New and updated translations from Transifex.

:-)


wagnered:
@Psy-Virus 

Thanks...

PR #1632, opened 2017-11-09
Small changes to make translation easier

Psy-Virus:
Changed and combined some strings to make translation in Transifex
easier. It's always better to not split up translatable strings, so
translators can see how they're put together and don't have to search
for the parts.

Also, in the `dropbox.catalog.php`, I changed the strings to not behave
like html tags. Transifex doesn't like this / treats it as html tags.
;-)

If it's better, we also could change the `\'` to `"`. But I guess it doesn't really matter.

EDIT: Btw. I also put some new rules into the reborn themes CSS to highlight the links in the "add-catalog-help-section"

:-)

### DIFF
--- a/modules/catalog/dropbox/dropbox.catalog.php
+++ b/modules/catalog/dropbox/dropbox.catalog.php
@@ -185,7 +185,7 @@ class Catalog_dropbox extends Catalog
         try {
             $app = new DropboxApp($apikey, $secret, $authtoken);
         } catch (DropboxClientException $e) {
-            AmpError::add('general', T_('Invalid <api key, secret, or access token>: ' . $e->getMessage()));
+            AmpError::add('general', T_('Invalid "API key", "secret", or "access token": ' . $e->getMessage()));
 
             return false;
         }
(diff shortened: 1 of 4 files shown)


wagnered:
@Psy-Virus
Please run php-cs-fixer and push the changes: The Travis CL tests  failed.

Also, it looks like the codacy tests didn't run, but ignore that. I will check. Maybe they changed it to only report errors.

Thanks...

Psy-Virus:
Yeah, sure, sorry, done. :D

I like it, that you pay attention to this. I did it in the past until I realized that it didn't matter and gave up on it because the PR after mine ruined it again, every time.
Next time I won't forget it.

:-)

Psy-Virus:
Btw. wouldn't it be easier/smarter, to do it with the merging process? I mean our friend travis is checking it and if there's a code style error, would travis not be able to fix it too? Or can it not change the content of the project?

Psy-Virus:
Btw. wouldn't it be easier/smarter, to do it with the pre-merging process? I mean our friend travis is checking it and if there's a code style error, would travis not be able to fix it too? Or can it not change the content of the project?

Psy-Virus:
Did another small change, added a `target="_blank"` to the seafile link, so it will open in a new tab.

It should be ready now.

wagnered:
@afterster  set the script to run php-cs-fixer with the  "--dry_run" switch. Possibly becasue it can't fix all problems and wanted the developer to be responsible for the fix.

Thanks...

PR #1633, opened 2017-11-09
Addendum to #1632

Psy-Virus:
Totaly forgot to recreate the pot file for the source translation
strings.
But here it is.

:-)


wagnered:
Thanks...

PR #1673, opened 2018-02-15
Update CHANGELOG.md

guiweber:


wagnered:
@guiweber 

Thanks for bringing this file to my attention.  Didn't realize it existed.

guiweber:
There is actually a message displayed within Ampache pointing to this file when a new version is available. Just a thought, but the message could be changed to point to the github "Releases" page instead, so this file wouldn't need to be updated.

Issue #1706, opened 2018-05-02
Enable dropbox catalog module breaks "add catalog" action


ftoledo:
# Environment

 * Ampache version: develop (git up to date)
 * Web server + version: Nginx 
 * Server operating system: Debian Jessie
 * Client operating system: Debian stretch

 * Client type:
   - Web (firefox)
# Description

Enable dropbox catalog module breaks "add catalog" operation
i try to select dropbox for catalog type and show "undefined" message.
It's do not show input fields for api keys.

If you select local catalog type, also do not show the input field for path.

![dock sud bbs music puro rock inicio](https://user-images.githubusercontent.com/95441/39510034-adb2ace8-4dbe-11e8-9977-913d17cfe9af.png)

Disable the dropbox catalog module restore the normal behavior.

Chrome console show error:

"index.php:1 Uncaught ReferenceError: catalogTypeChanged is not defined
    at HTMLSelectElement.onchange (index.php:1)"

Firefox console show error: 
ReferenceError: catalogTypeChanged is not defined

Maybe a quotes problem in the javascript?
![pantallazo-2018-05-02 04-14-54](https://user-images.githubusercontent.com/95441/39510293-a19e5c6c-4dbf-11e8-9cfe-3e11e0dfe6a6.png)

wagnered:
@ftoledo 
I just cloned/installed the latest develop version from github and cannot replicate your issue:
![image](https://user-images.githubusercontent.com/6115438/39536800-7ffa40a8-4dec-11e8-9b09-ba500ddf9a36.png)

Please enable the ampache debug log.

Do you have JavaScript enabled in the browser?

Also, have you tried the release version [ampache-3.8.6_all.zip](https://github.com/ampache/ampache/releases/tag/3.8.6)?

ftoledo:
Hi i enable debug and level to 5, do not see anything related to dropbox, but:

` (autoload) -> 'Beets\Catalog' not found! `

when click on "add new catalog" menu item

I'm sure that javascript is enabled at the browser and beets catalog module are disable:
.
$ git pull
Already up-to-date.
i will try to clean dir and clone again =?

ftoledo:
i found the issue.
the spansh translation using double quotes on "Generar" word do break javascript string for the catalog fields:

```
	<td>
		<ul>
			<li>Ve a https://www.dropbox.com/developers/apps/create</li>
			<li>Selecciona 'Dropbox API app'</li>
			<li>Select 'Full Dropbox'</li>
[...]

PR #1712, opened 2018-05-04
General Translation Update

Psy-Virus:
The usual update + The fix for #1706

:-)


wagnered:
Thanks for responding so quickly.

PR #1715, opened 2018-05-08
Setting favorite album or artist wont work via the Subsonic-Api

robinxan:
The `type` in `$ids[]` is not properly assigned when `star.view` or `unstar.view` was called with `albumId` or `artistId` resulting in a `NULL` value in the `user_flag`.`object_type`-Column.


wagnered:
@robinxan 

Looks good.

Much appreciated, Thanks

PR #1718, opened 2018-05-09
Fix #1717 performance issues with playlist display

### LINKED ISSUE #1717: Slow performance because of playlist
# Environment

 * Ampache version: 4aa6f1332639009d66ba789d4068d611de450c5a
 * Web server + version: Apache 2.4.18
 * Server operating system: Ubuntu 16.04
 * Client operating system: Firefox

Hi there,
I noticed some performance issues since few months and just find out the reason. Any page takes at least 3 seconds to load because of the `Playlist::format()` from https://github.com/ampache/ampache/blob/develop/templates/rightbar.inc.php#L40 .
I kept diving into the code and found [this heavy method](https://github.com/ampache/ampache/blob/develop/lib/class/user.class.php#L1010) called for each playlist created.
I have only 4 playlists created and without this code, the page takes only 0.2 seconds to load against 3s.

I would love to push a PR but wanted to know before if anyone knows why we needed it?

Simounet:
This part call the playlist_object.abstract.php for each playlist from everywhere. It results on User::format calls with the heavy bandwitch usage sum MySQL query.
I'm a long time Ampache user so my bandwitch usage is pretty loaded. Before using playlists, I hadn't any performance issue. I never make the link between playlist and my performance issues until today.
I didn't spend a lot of time to test my fix but didn't encounter any trouble so far.

### DIFF
--- a/lib/class/search.class.php
+++ b/lib/class/search.class.php
@@ -421,7 +421,7 @@ class Search extends playlist_object
             $playlists = array();
             foreach (Playlist::get_playlists() as $playlistid) {
                 $playlist = new Playlist($playlistid);
-                $playlist->format();
+                $playlist->format(false);
                 $playlists[$playlistid] = $playlist->f_name;
             }
             $this->types[] = array(
--- a/templates/rightbar.inc.php
+++ b/templates/rightbar.inc.php
@@ -37,7 +37,7 @@
     Playlist::build_cache($playlists);
     foreach ($playlists as $playlist_id) {
         $playlist = new Playlist($playlist_id);
-        $playlist->format(); ?>
+        $playlist->format(false); ?>
                 <li>
                     <?php echo Ajax::text('?page=playlist&action=append_item&playlist_id=' . $playlist->id, $playlist->f_name, 'rb_append_playlist_' . $playlist->id); ?>
                 </li>


wagnered:
My apology,. I commented in the issue REF #1717 instead of here where it should have been.

> OK, I see what you are referring to. But:

>1. The value of the $details parameter should not be hard-coded like that. There are a number of procedures ("search" for one) that calls Playlist::format($details) with either false or true.
>2. Users should be able to decide if they want their full name displayed. This would require a config switch for the user to select.
>3. The work would be significant to properly incorporate your change and a policy is in place to not add any new features in order to have time to work on Ampache/Laravel (Ampache-4.0).

>I do appreciate your desire to contribute to the Ampache project and don't want to discourage you. I consider the fact that Ampache is so slow to be a bug and would be happy to include a PR that speeds up Ampache. see the coding standards and coding philosophy

>The reason that Ampache is slow is because to get the information needed from the database, each object is created, this takes at lease two additional scripts to create each song, artist, album, etc object.
Ampache would be much faster if all features accessed the information directly from the database.

>Ampache/Laravel will access the database directly and should be much faster than Ampache-3.*

a7medo778:
😍😍😍
Already hyped about the laravel version, cant wait to see it out

Simounet:
No worries mate! I'm used with free software development and I totally understand your position. The code changed here doesn't affect directly the user or maybe I missed something. I will try to update it.

ps: I didn't know the Laravel version was still a thing. Good to know!

wagnered:
@a7medo778 
Don't hold your breath :smile:.

Although it is going a little faster, there is still a lot of work to be done, and a lot of learning to do.

Simounet:
I updated my fix with a lower impact.

robled:
I've tested this patch and have observed a dramatic performance improvement with page loads and search queries.

wagnered:
@robled 
Thanks for bringing this to my attention.  As you gave this PR a "Seal of approval" I'll merge the PR. :smile: a
And I apologize to @Simounet for taking so long to follow up.

Simounet:
No worries. I'm using it locally for a while now. Glad to be useful to Ampache users. :)

PR #1725, opened 2018-05-14
Fixed an issue that artist name became empty in SubsonicAPI.

KanzakiKino:
I'm using latest Ampache (248d09e88c7e6cf0dbbe2f8872155ca212dc389b) with UltraSonic. But artist name of songs have been not displayed since the version. Because `$artistData['f_full_name']` doesn't exist.  
So I replaced `f_full_name` to `name` to fix the issue.

### DIFF
--- a/lib/class/subsonic_xml_data.class.php
+++ b/lib/class/subsonic_xml_data.class.php
@@ -483,6 +483,9 @@ class Subsonic_XML_Data
         
         $row = Dba::fetch_assoc($db_results);
 
+        $row['f_name'] = trim($row['prefix'] . ' ' . $row['name']);
+        $row['f_full_name'] = trim(trim($row['prefix']) . ' ' . trim($row['name']));
+
         return $row;
     }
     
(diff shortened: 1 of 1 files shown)


wagnered:
@KanzakiKino 

Thanks for the PR.
I tried to avoid creating an artist object (artist.class.php) to increase the access time of a large playlist. I forgot about having to use the Artist::format function to get `$artistData['f_full_name']`.

The statements in Artist::format that format "f_full_name":
>        $name              = trim($this->prefix . " " . $this->name);
>        /* Combine prefix and name, trim then add ... if needed */
>       $this->f_name      = $name;
>       $this->f_full_name = trim(trim($this->prefix) . ' ' . trim($this->name));

 should be done in Subsonic_XML_Data::getArtistData($artistId) using the data found in $row, not where you made the change in the PR.

If you would like to attempt the changes I suggest, you are welcome.  Just push any changes you make in your clone and they will automatically be added to this current PR.

Otherwise, I can correct my mistake(s) myself.  I make a lot of them.  :frowning_face: I appreciate that you brought it to my attention.

KanzakiKino:
I added codes that creates `$row['f_full_name']` after retrieving artist information.

wagnered:
@KanzakiKino 
Thanks...

PR #1730, opened 2018-06-04
General Translation Update

Psy-Virus:
Transifex Translation updates of May.

:)


wagnered:
@Psy-Virus  Thanks, Much appreciated.

Issue #1704, opened 2018-04-29
Can't play songs with clementine using the subsonic API


Gotos:
# Environment

 * Ampache version: current git master and develop branch
 * Web server + version: apache2.4.18
 * Server operating system: Ubuntu 16.04
 * Client operating system: Gentoo

 * Client type:
   - SubSonic (Clementine 1.3.1)

# Settings

Fresh install with community-settings and Subsonic enabled. I think I have reverted all other changes I tried to fix this issue. I already read [the FAQ about problems like this one](https://github.com/ampache/ampache/wiki/FAQ#when-using-an-api-ampache-subsonic-upnp--i-can-browse-my-library-but-playback-doesnt-work-whereas-it-is-working-using-the-web-interface-how-to-make-it-work), but that did not help. Reading through other issues, open and closed, and trying proposed fixes, didn't help either, but the logs mostly suggested different issues anyway.

# Description

I can not play songs using Clementine 1.3.1 via the subsonic API. I can load the catalog I created and see all the songs in it, but if I try to play one, the song is skipped immediately. In clementines logs I see a message about not being able to play the song because the "stream does not contain enough data" (translated from the error message that was originally in German - "Der Datenstrom enthält nicht genug Daten."), but no more details.

Interestingly, I can play music using other clients like Subsonic Music Streamer for android or the chrome app SubFire. Therefore I am not sure if this even is a problem with ampache, but others people successfully used clementine with ampache and because clementine has fewer options I could have wrong, I guess it's more likely that I missed something on the ampache side.

Looking at the logs, I can see a difference between clementine and subsonic music streamer: Only the later accesses /play. I don't know why. The access to /rest/stream.view is also totally different between the two clients, but I think that's because of different API versions?

# Logs

On my server, in the access.log of apache I see the following when trying to play a song with clementine:

```
[ipv6] - - [29/Apr/2018:19:03:05 +0200] "GET /rest/getCoverArt.view?id=300000005 HTTP/1.1" 200 3985 "-" "Clementine 1.3.1"
[ipv6] - - [29/Apr/2018:19:03:05 +0200] "GET /rest/stream.view?v=1.8.0&c=Clementine&u=admin&p=enc:[...]&id=17.225.163.5 HTTP/1.1" 200 3669 "-" "Clementine 1.3.1"
```

For subsonic music streamer, these lines are as follows. No idea why I get the first one twice.

```
[...]

PR #1734, opened 2018-06-10
add workaround for buggy clients

markuswernig:
Some buggy (supposedly QT5) clients send the id parameter formatted as an IP address (sic!) to the Subsonic API.
See https://github.com/clementine-player/Clementine/issues/6080 and https://github.com/ampache/ampache/issues/1704
This workaround tries to detect that and calculate the correct song id from the parameter before passing it on.


wagnered:
@markuswernig  Many thanks.... Much appreciated

Issue #1738, opened 2018-06-18
Missing translations


prschopf:
# Environment

 * Ampache version: 3.8.8

# Description

I couldnt find these strings in https://www.transifex.com/ampache/ampache/dashboard:

![trad1](https://user-images.githubusercontent.com/16743256/41543716-cb163f8a-72ed-11e8-89da-41e26b677cf9.jpg)
![trad2](https://user-images.githubusercontent.com/16743256/41543718-cb3fdf48-72ed-11e8-87a0-d44f56174623.jpg)
![trad3](https://user-images.githubusercontent.com/16743256/41543719-cb658c8e-72ed-11e8-8662-9cef8ba6c740.jpg)

How to translate them?

Psy-Virus:
Hi @prschopf,

"Open Basedir" is basically a PHP setting's/module's name like Zlib, Gettext, iconv and so on.
The others are, like the option descriptions, categories and subcategories, saved within the preference table of the database. But they are lowercase written there. So they definitely will be processed somehow.
I was digging a bit. But sometimes the Ampache code makes it kinda hard for me to find, what I'm looking for :D

But I guess to fix this, we need to modify the `gather-messages.sh` in `locale/base` dir.
A while ago I built in a workaround to also translate the preference descriptions from the database that were untranslated before. That's what the `translatable-database-strings.txt` in the same directory is for.

I'll see what I can do. Should be easy to fix once I get what I did there :-P

:-)

prschopf:
I have found an empty string to translate too and its impossible:

![transl](https://user-images.githubusercontent.com/16743256/45425000-80e5c600-b66e-11e8-816c-b65f4442bcb0.jpg)

PR #1760, opened 2018-08-13
Latest translation updates and maybe fix for #1738

Psy-Virus:
Finally I had some time and motivation to look into #1738 again. And I guess I fixed it. But the problem that the subcategory - database entries are processed somewhere to be written with first letter uppercase, (I still can't find it) could make translations ineffective.
I also see some conflicts to already existing translations for the word "upload".

The bad news is that we probably have to rewrite the translation process to be more modular and get rid of these descriptions in the database and put them into the modules, or where ever they belong to, directly.

But I guess this shouldn't be done for this version of Ampache anymore. But maybe we (by that I mean @wagnered, because I don't even have an idea where to start :P) fix that with the new Laravel version.
Do you agree @wagnered?

:)


Psy-Virus:
> I was considering a gettext composer package which makes manipulating/displaying languages very easy. It also provides a convenient way for the user to switch languages dynamically. The problem I see is that the *.po file doesn't allow for very much versatility. and requires the PO editor to maintain translations.

Gettext is probably one of the best ways for localization. At least does the internet say that. The question is: What is faster and more user friendly? If you think that way, that most software users aren't developers and fear the (At least at the beginning) overwhelming complexity of code, I guess the comfort of a GUI based PO Editor and/or Transifex is the best choice. And it's simple, fast and easy to use. We also have different roles there, like translators, that can translate but not review changes and new translations. Then, if someone wants to have more responsibility, he/she can ask in Transifex/GoogleGroup or here to become a reviewer and take care of the quality, the translators provide.

It is probably also possible to setup the Transifex CLI client via a GitHub service to automatically update the localization once a week, or more often if needed @Afterster tried something like that with a Amazon Cloud Instance, a while ago.
I update it more or less once in a month manually, depending on the activity on Transifex. An automated process wont change anything really because the most used languages are at about 100% and most translators translate just a few words or sentences and are done with it.

> Also some words like "open_basedir" should probably be replaced with more user friendly phrases like "File Access Restricted to".

`open_basedir` is something you only find in the PHP.ini. I would recommend to put an `(?)` behind it and make a hover text for it, that explains what it is for?

wagnered:
> @Afterster tried something like that with a Amazon Cloud Instance, a while ago.

That's probably why there is a .tx folder in the repository.  I looked into connecting github with transifex.

Quoting [Transifex](https://docs.transifex.com/integrations/github-txgh) it requires:
>A free Amazon EC2 instance. The basic free Amazon Linux AMI should be enough.

The problem is that at this time, Amazon EC2 is only free for a year. and the AMI only works with Amazon EC2. Also, from what I read, It looks like Transifex automatically commits to github  only when a language is 100% translated.

> I would recommend to put an (?) behind it and make a hover text for it, that explains what it is for?

I like that idea.  Thanks.

PR #1775, opened 2018-09-19
Saving rating as popularimeter into ID3 of MP3 songs.

tdeuling:
In case of setting write_id3 is activated, the user rating will be written in popularimeter tag of a MP3 song.
For saving this rating the user mail address will be used. If no user mail address is available, the user@example.net mail address will be used.

### DIFF
--- a/lib/class/song.class.php
+++ b/lib/class/song.class.php
@@ -1050,6 +1050,39 @@ class Song extends database_object implements media, library_item
                         $meta[$metadata->getField()->getName()] = $metadata->getData();
                     }
                 }
+                // Get user mail for rating
+                $ratingMail = trim($GLOBALS['user']->email);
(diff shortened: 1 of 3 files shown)


wagnered:
Hello @tdeuling 
PR's for new features have been put on hold to  develop the  Ampache (4.0) version based on Laravel.  Only one developer (me) is actively working on Ampache so I only have time for bug fixes and developing Ampache/Laravel.

I'll consider it an enhancement for the new version if you can help me understand why only write capability and why only mp3: Why not flac?

Also, the PR failed one of the tests.  It's necessary to run 'php php-cs-fixer -v fix`before pushing your changes to your repository.

wagnered:
I think this is a good idea, so I'll mark this as an enhancement to be added to the next generation Ampache.

PR #1783, opened 2018-09-28
Multiple changes

mrmajik45:
Here're some changes I made. (I'll be doing more later if I can.)

### DIFF
--- a/bin/catalog_update.inc
+++ b/bin/catalog_update.inc
@@ -125,7 +125,7 @@ if ($plimp == 1) {
 // --------
 if ($catname) $where = " name = '$catname' ";
 if ($cattype) { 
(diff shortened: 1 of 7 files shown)


wagnered:
Aurora.js is a package developed by [audiocogs](https://github.com/audiocogs/aurora.js). It is a well tested package of decoders. It works well in Ampache.  Many  changes codefactor made would actually cause the code to fail.
For example in aac.js:
```
if (this.commonWindow = !!stream.read(1)) {
```
is valid. `this.commonWindow` is tested on what is returned by  `stream.read(1)`.

I don't mean to discourage you from trying, but may I suggest that you acquire a deeper understanding of JavaScript, and the limitations of Codefactor before submitting PR's. :smile:

PR #1794, opened 2018-10-23
Addyear

noyse27:
add year as new column in several views(e.g. recently, now playing) / possible search from now playing for all albums with this year


noyse27:
unfortunately it was more commited than expected

PR #1804, opened 2018-11-10
Fix_Filenames: Error Opening Catalog

obel1x:
When i ran bin/fix_filenames.inc by executing:
sudo -u wwwrun php fix_filenames.inc
i get:
Quellzeichensatz (UTF-8):                                                                                                    
Verwende UTF-8 als Quellzeichensatz                                                                                          
Prüfe  ()
Fehler: Fehler beim öffnen von 
Überprüfung der Dateinamen, auf passende Zeichen, fertig gestellt

Thats it. I guessed that fetch $row['0'] will look for column named '0' instead of column Number 0 which is named 'id'.
I changed it and it works.
So here is my Version for you to check.

### DIFF
--- a/bin/fix_filenames.inc
+++ b/bin/fix_filenames.inc
@@ -68,7 +68,7 @@ $db_results = Dba::read($sql);
 
 while ($row = Dba::fetch_assoc($db_results)) {
 
-    $catalog = Catalog::create_from_id($row['0']);
+    $catalog = Catalog::create_from_id($row['id']);
     printf(T_('Checking %s (%s)'), $catalog->name, $catalog->path);
     echo "\n";
     charset_directory_correct($catalog->path);


wagnered:
Thanks @obel1x 

> I guessed that fetch $row['0'] will look for column named '0' ....

I noticed that with other files in the bin folder.  I'm wondering if there was some changes in php behaviour over the the years.  I think using $row[0], not $row['0'] might have worked too. Of course that's assuming that $row['id'] corresponds to $row[0].

Also, I noticed that when running the fix_filename.inc it went though my many files and folders very quickly.  I check the Ampache log and found:

> 2018-11-10 20:22:14 [ampache] (PHP) -> [Runtime Error] iconv() expects exactly 3 parameters, 2 given in file /home/ernie/workspace/ampache-develop/bin/fix_filenames.inc(51) 

Would you like to take a look at that error also?

Thanks....

wagnered:
I modified line 51 which removed the error, and there are no more errors in the log, but  something seems to be preventing recursion to the files.

So I'll step thorugh the logic to see what the problem is.

wagnered:
Never mind, Ignore my previous comment.  I stepped though the code and found that if no filenames need fixed the operation is very fast.

There is a problem with line 51: I take care of it after the merge.

Thanks again.

wagnered:
Line 128 in strcmp function has '0' instead of 0.   Will be necessary to check all the files in the bin folder.

PR #1809, opened 2018-11-17
Fixed use of \n on a single-quoted string

roobre:


wagnered:
@roobre 

Nice catch. Thanks again....

This is somewhat embarrassing.  I'm glad I'm not to blame.  :smile:

PR #1811, opened 2018-11-17
Change Song->Album + Albumartist with write id3 and CustomMetadata

obel1x:
1.  Moving Song to new Album would not work when writing of Metadata is enabled
As update-functions will only set db-ids to new value and will not change real names, the id3 tags will still be written with old names. When reading Tags afterwards, the changes are discarded. Added format() in between update of db.ids and writing prevents this.
2.  Changing AlbumArtist won't work when custom metadata is enabled
So i changed the order in which the Metadata is parsed in write_id3. This will set custom fields to values of song-instance as they are defined in get_metadata, overriding them as needed.

### DIFF
--- a/lib/class/song.class.php
+++ b/lib/class/song.class.php
@@ -1057,7 +1057,7 @@ class Song extends database_object implements media, library_item
                 //default fields back again, making it impossible to change e.g albumartist
                 //(albumartist will be overridden by custom metadatafield band which is the id3 field for albumartist, look at relations in get_metadata)
                 $meta = $this->get_metadata();
-                $id3 = new vainfo($this->file);
+                $id3  = new vainfo($this->file);
                 $id3->write_id3($meta);
                 Catalog::update_media_from_tags($this);
             }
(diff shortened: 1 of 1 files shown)


obel1x:
And last now made Travis CI happy with Style of code.

wagnered:
Hello, @obel1x 

I appreciate your enthusiasm and support of Ampache.

Above, you list what appears to be two distinct problems. Although the fixes are caused by the same file, each fix should be in a separate PR.  This keeps each PR smaller, better organized in the git history and is standard procedure.

This can be accomplished by:

- The easy way:  Redo the PR that was created from your develop branch. After that is merged, create the second PR with the second fix.

    OR
- For multiple complex fixes/features:

1.  Creating a PR for the first fix/feature in the develop  branch.
2.  Create and checkout a new branch for the 2nd fix/feature.
3.  Make the changes and push the commits to your remote.  Github will create the new branch on your fork.
4. On Github, switch to the new branch and a PR can be created.

Thanks....

PR #1815, opened 2018-11-25
Fix negative Year dba-error

obel1x:
Most mp3-tag- programs set Year = -1 for none given. This will lead to an error when reading tags and updating db after reading tags. This fix sets Year to 0 in this case.

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -1574,6 +1574,9 @@ abstract class Catalog extends database_object
         $new_song->file        = $results['file'];
         $new_song->title       = $results['title'];
         $new_song->year        = $results['year'];
+        if ($new_song->year < 0) { //no year = -1 in most programs, but this leads to error in db-update statement
+            $new_song->year = 0;
+        }
         $new_song->comment     = $results['comment'];
         $new_song->language    = $results['language'];
         $new_song->lyrics      = str_replace(


wagnered:
@obel1x 

Thanks,

What is the error given during updating?

Some ideas to think about (colloquially: "points to ponder" ):

I'm  wondering about your assumptions.  If you are correct, your assumption implies another problem: What about when a catalog is newly created,  wouldn't the same problem exist in vainfo.class.php with the grabbing of the year tag?

Theoretically,  year 0 is a valid year, although there probably wasn't much music tagged during that era :smile: .

What about untagged files with no year tag?   The TYER tag is a 9 digit string: would your changes be valid for null or empty strings? I tested EasyTAG for Linux.  I deleted the year and saved the tags.  There was no longer a TYER tag in the file.

How would the year 0 display on the songs page? Is it blank  or 0? I think blank would be preferable to 0.

--------

It's important to try to see the entire picture of the validity of your assumptions and of what your changes might cause.

I'm not trying to say this PR is wrong, it might just be incomplete, or require a different PR for creating new catalogs which probably have the same problems.

Thanks again,

obel1x:
Yes, that thoughts are to mention. The situation is/was:
I set up a new ampache and new catalog from scratch. The songs were all imported with no error. After Changing something at the song with EasyTag for faster working - which complained about wrong Date, but did change the Tags - changes would not apply when scanning album with "update from Tags", so i checked the Logs, which i have thrown away :( . They said smtg. like: Year to be Changed from OLD: 0 to -1, which was the Information still beeing in the Tags (this has not changed). This resulted in Error DBA: Wrong value for Column year (-1).
As i didn't find a way to fix the broken year i didn't check further and did some debugging, changing the value to what it seemed to be before in DB according to error-msg.

Maybe this is not the right way to set the value, but as i would like to manage the collection by ampache completely, such a wrong value should be handled somehow. Maybe it would be better to drop the Date completely in this case instead of doing 0. I have not checked how this would behave, because it looked like db did already insert 0 when scanning initially and changing it, could result in further problems, like Tag not beeing cleaned when rewriting it.

Unfortunatelly my time is very limited, i am currently not able to have a deeper look.

Maybe it would help if you could tell me if you are using a debugger for php and if so which one? i only found xdebug, but had no luck getting it to work. Maybe i could have a deeper look to the initial import-process than to find, what had gone on while doing catalog scan.

wagnered:
I use Eclipse-pdt to step through the code. It uses xdebug.  Once set up, it is easy to set  breakpoints. I can help you get it set up. 

The getID3 library in Ampache has a demo that can display the tags of a music file.

PR #1824, opened 2018-12-06
Vainfo::clean_tag_info cleans year and disk

obel1x:
It may happen that Tag "Year" is not well formatted (e.g. containing -1), which would lead to DB-Error in updating the field when "Reading from Tags" is done. The cleaning routines of Album::Check were moved into this function to get this working.

### DIFF
--- a/lib/class/album.class.php
+++ b/lib/class/album.class.php
@@ -400,10 +400,6 @@ class Album extends database_object implements library_item
         $mbid_group   = empty($mbid_group) ? null : $mbid_group;
         $release_type = empty($release_type) ? null : $release_type;
 
-        // Not even sure if these can be negative, but better safe than llama.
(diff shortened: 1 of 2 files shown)


wagnered:
Almost..

It failed the Travis CI build.  Please see the Ampache [Coding Standards](https://github.com/ampache/ampache/wiki/Coding-Standards#automatic-coding-standards-check) to see how to download and run php-cs-fixer to fix any problems with codestyle before pushing commits to your remote. 

It should have been included in the repository and I got to find out why it isn't.

wagnered:
@obel1x,  Ignore my previous message. There seems to be a problem with the Travis System. Also ignore the Codacy errer.

wagnered:
Thanks, I re-ran the Travis test and merged.

PR #1825, opened 2018-12-07
Illegal String Offset in edit.server.php

obel1x:
This fixes the error-message [Runtime Error] Illegal string offset 'id' in file /srv/www/htdocs/ampache/server/edit.server.php(143) showing up in log.

### DIFF
--- a/server/edit.server.php
+++ b/server/edit.server.php
@@ -140,7 +140,7 @@ switch ($_REQUEST['action']) {
         $libitem->format();
 
         xoutput_headers();
-        $results['id'] = $new_id;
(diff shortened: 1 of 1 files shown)


obel1x:
Yes, i did open the Issue #1826 for this.
The code is not really wrong, it is just not future-proof way of initialising arrays and could lead to problems in next php- versions. I added the links to phpdoc. So just a small slightly cleaner way to do it, but eliminates errors in logs.
Regards, Daniel

lachlan-00:
Merged into my core branch and will flow into official core branch soon.
Thank you!

PR #1830, opened 2018-12-12
Fix various Subsonic issues

### LINKED ISSUE #1829: Various incompatibilites with Subsonic API
# Environment

 * Ampache version: HEAD (develop branch)
 * Web server + version: Apache
 * Server operating system: Linux (Synology)
 * Client operating system: Raspbian
 * Client: py-sonic

# Description
Various parts of the Subsonic API appear to be implemented incorrectly. Firstly, there is a bug in lib/class/subsonic_api.class.php - Line 731 should probably be $song   = Subsonic_XML_Data::getAmpacheId($songid); because the way it is right now causes it to fail and getSong.view to not work (only "ok" status, but no song returned). Also, the xml2json function does not work correctly in this case because it converts the single-element XML list to a JSON Array which is in violation of the official implementation and breaks clients. This can be fixed by removing song from the alwaysArray config element, however, I presume this breaks other things. I think there are more issues with the implementation as I have not been able to get it to work at this point - I will update as I find them.

darkphoenix:
This PR tries to fix some incompatibilities with the Subsonic API - specifically, it fixes one bug and corrects some behavior in the XML-to-JSON converter. I am not entirely sure I got all the endpoints that need special treatment, and somebody else should definitely take a closer look at this, but for now, it seems to fix #1829.


lachlan-00:
I'm going to test this on my side install and see if i can get someone to push it through.

DSUB is probably the best android ampache player i've used so it would be good to keep it compatible.

lachlan-00:
I get a connection error in dsub using this file. does it work for you?

lachlan-00:
I've been using this for a week now and haven't seen any issues in dsub, is there a way you were able to break things before?

darkphoenix:
The issues were with https://github.com/Prior99/mopidy-subidy and py-sub, specifically, which used the JSON API. As far as I can tell, DSub uses the XML variant, so these fixes really shouldn't affect it. The only other thing I changed was getsong, which gets the properties of a song based on its ID and doesn't seem to be used all that often because those details are usually passed already in responses to other requests. mopidy-subidy uses it because it needs to resolve IDs from playlists.

lachlan-00:
I've merged this into my core branch and it will move into the official core branch soon.

Thanks for becoming an AMpache Contributor!

PR #1837, opened 2018-12-27
adding year to several views/ gather art when update from tags

noyse27:


noyse27:
adding column/row Year on:

- album
- song
- nowplaying
- recently played

making albums searchable with year from nowplaying year






gather art when updating album from tags

change last.fm scheme
from https to http as described in Last.fm scrobbler api
https://www.last.fm/api/scrobbling 

> As with all our write web services, requests must be sent as HTTP POST requests to http://ws.audioscrobbler.com/2.0/ with form urlencoded parameters in the body of the request. The text encoding must be UTF-8.

noyse27:
@wagnered 
i hope i did it well now :)

Issue #1600, opened 2017-09-16
Show ID3-Art-Image per song and not per album/artist


xanthos84:
Hi everyone,

is it maybe possible to let Ampache gather the art/image from the mp3 files (for every file) instead of gather it for the complete "Album"?

So it finally saves and shows the actual art from the mp3 file instead of the first image ampache found for one album tag...

Background: I dont have traditional "albums" in my collection, instead i use the albums tag for music "labels".. so I got various artists under one album tag. but they all have different images! So I want to see the image from the specific files...

Hope you understand my question.. as english is not my motherlanguage... Thanks!

xanthos84:
what I found so far...

Is it possible to add to the "image" table in sql some kind of "song" object_type? - at the moment only album&artist are inserted there.

And then it should be possible to call the 'song' database row to show on the player/search results?

I am nowhere near a real programmer.. just basic knowledge... so is maybe anyone able to code me that "song" insert sql for i guess the art.class.php? I would donate some money if its needed :)

Jucgshu:
I know this is not the answer you're looking for, but I use Ampache only as a music player (e.g. *not* as a tagging tool).

Instead, I use [beets](http://beets.io/) and I'm reaaaaally happy with it.

In your usecase, you should have a look on beets [singleton mode](http://beets.readthedocs.io/en/stable/guides/tagger.html?highlight=singleton).

xanthos84:
Thanks for your reply... but .. well I also dont use ampache as a tagging tool....

I only talk about SHOWING tags... showing the FILE/ID3/ART per file during playback and on the search results.
But to show images, I figured out that Ampache first needs them to read into the database.
And what I saw it only reads album+artist.
So I need to extend(?) the classes to read the image/art per song and write it to the database.
And then change the player+search results to show the song art/image instead of something else(usual it shows artist,album...)

So if anyone can help and want some donation for it... let me know..

xanthos84:
so far I got it now implemented.. maybe not perfect but at least it works, it gathers image id3 tags per file and writes it to the image table.. and the players and search results are showing the song artwork prefered over the album artwork.

maybe the devs could implement some kind of official implementation for this request,...

phynias:
why not put a pull request in with your changes ?

a7medo778:
yes please, it would be interesting to see the results, i am going to overhaul the api commands as soon as the laravel port is done, so this will be interesting

xanthos84:
because i am in no way a real developer :)
[...]

PR #1845, opened 2019-01-28
Solution for issue 1600

Greo83:
Original discussion: [https://github.com/ampache/ampache/issues/1600](url)

I changed the gather_art routines so that images of id tags are stored per song in image table. Furthermore I changed (hopefully) all places where the album image is shown into the song image.
It works for the webplayer and subsonic API. I was not able to check XML API because it does not work in my environment. All other APIs are disabled, so I did not touch them.


Greo83:
updated with rework 1

PR #1847, opened 2019-01-30
Solution for issue 1600 (rework 2)

Greo83:
Original discussion: [https://github.com/ampache/ampache/issues/1600](url)

I changed the gather_art routines so that images of id tags are stored per song in image table. Furthermore I changed (hopefully) all places where the album image is shown into the song image.
It works for the webplayer and subsonic API. I was not able to check XML API because it does not work in my environment. All other APIs are disabled, so I did not touch them.


Greo83:
I think the fail of the codacy check does not have to do anyhting with my changes. Please advice what I should change.

Psy-Virus:
Hey,
I'm just guessing here but it seems that codacy is maybe complaining about
`$art_url = Art::url($song->id, 'song', $_REQUEST['auth']);` because `$_REQUEST['auth']` seems not to be validated. (?)

Greo83:
But I only changed some parameters that are passed to the function behind. 
Code before:
`$art_url               = Art::url($song->album, 'album', $_REQUEST['auth']);`
Code after:
`$art_url               = Art::url($song->id, 'song', $_REQUEST['auth']);`

The only thing that I changed is to pass a different property of the song object and a different string into the function. Is it possible that codacy's check itself was changed, so that even the original ampache code is not valid any more and would fail the check?

Psy-Virus:
But I also have a question regarding your PR.
Will your changes mean that we will have a bunch of duplicate album art?
It would probably be better to only gather art per song if the ARTISTSORT tag field is "Various Artists" or "VA" for example.

Greo83:
I just doublechecked it in my environment. The steps to gather album art have not changed: For each song Ampache checks if its album already exists in the database. If not then it creates the album entry and uses the image stored in the ID tag of this first song as complete album art. The image is stored in the image table, related to the `object_id` of the album with `object_type` set to album.

The new thing now is that additionally the image of every song is also stored in the image table to `object_id` of the song with `object_type` set to song. From this point of view the picture of the first song exists twice in the image table: One related to the album and one related to the song.

Furthermore I changed the displayed art in the player and playlist to show the song's image. The album's image is only shown in the list/detail view of the album.
This makes sense for catalogues containing mainly single songs without albums. If the catalogue always contains full albums the image table possibly would contain the same image multiple times. 
Perhaps it is possible to create a setup so that the users can decide which way they prefer? Unfortunately I do not have enough PHP knowledge to implement it like that. :-(

Psy-Virus:
> Perhaps it is possible to create a setup so that the users can decide which way they prefer? Unfortunately I do not have enough PHP knowledge to implement it like that. :-(

That would probably a much better way and maybe you get some inspiration when you take a look at the following link :)

https://github.com/ampache/ampache/pull/1611/commits/608368a88992a899e99a983936992717fcbcf1cf

Greo83:
Thats everything? Okay, I think I can do that. I am on holiday now so please give me some time.
My idea is to create two setups: One that enables storing images per song, and one more that switches if album art or song art is shown in the player. That would give the users more flexibility. What do you think?

Psy-Virus:
> Thats everything? Okay, [...]

Of course it's a bit more complex what you want to achieve. But this is basically how you can make something in Ampache "config file-setting related"

> My idea is to create two setups: One that enables storing images per song, and one more that switches if album art or song art is shown in the player. That would give the users more flexibility. What do you think?

In theory it sounds great but it still should be optional. So the default behaviour should be how it is right now. And I can't promise if it gets implemented because Ampache is more or less undergoing a major rewrite into the Laravel PHP-Framework. What basically means that no big new things will get implemented into this version. But this or that way, it's worth to have here, either as implemented feature for this version of Ampache or as hint or inspiration for the laravel version that will come ... maybe... some ... time.. :D

@wagnered, if you're reading this, I still feel bad that I still didn't have the time to help you!
And I hope we hear something from you again. Let it be a ping or a "Hey guys, I'm still alive!"

Greo83:
Hi again,
after my holiday I implemented the changes with two setups. As you suggested the new setups are optional. That means if nobody changes them the system works exactly as before. The users can decide on their own if they prefer art per song.

I'm pretty sure that some code checks will fail again, but I'll keep an eye on this. :-)

Greo83:
Wow, only one commit later all checks have passed, cool! Very good thing those checking scripts.

Because I'm new to Github I'm unsure if I have to create a new pull request now. The headline says that I want to merge 5 commits into the develop branch. Sounds good to me, although only the last commit is the correct one. Did I do everything correctly?

Psy-Virus:
Can you squash your commits so there's only one? That would be nice, easier to read and so on.

:)

Greo83:
Well, I did not know that this is possible until yesterday. ;-)

I tried to squash the commits and first got some real mess because of the new changes in the original main repository since my fork was created.
But I think finally it looks good now... I hope! I'm an absolute newbie in Git, I'm sorry.

lib/class/art.class.php
```diff
@@ -784,8 +784,17 @@ public function generate_thumb($image, $size, $mime)
      * @param string $type
      * @return string|null
      */
-    public static function get_from_source($data, $type = 'album')
+    public static function get_from_source($data, $type)
     {
+        if (!isset($type)) {
+            $show_song_art = AmpConfig::get('show_song_art') ?: false;
+            if ($show_song_art) {
```
Psy-Virus: Uhhhhhhhh that's... uhhhhhh

This basically means:
If there's ANYTHING else than `0; "0"; ""; null` in the variable, we get `$type = 'song';`.
Even if we would have `'false'` in it, it would be true.

please take a look at this: https://www.php.net/manual/en/types.comparisons.php

We need a stricter comparison here. :)

arts.php
```diff
@@ -150,7 +150,12 @@
         // Prevent the script from timing out
         set_time_limit(0);
 
-        $image = Art::get_from_source($_SESSION['form']['images'][$image_id], 'album');
+        $show_song_art = AmpConfig::get('show_song_art') ?: false;
+        if ($show_song_art) {
```
Psy-Virus: Uhhhhhhhh that's... uhhhhhh

This basically means:
If there's ANYTHING else than `0; "0"; ""; null` in the variable, we get `$type = 'song';`.
Even if we would have `"false"` in it, it would be true.

please take a look at this: https://www.php.net/manual/en/types.comparisons.php

We need a stricter comparison here. :)

Same is further down. I guess your work is not done yet^^

Greo83: Oha! I didn't know that, sorry. Of course this is a very bad style. How can I change the code properly? `if ($show_song_art == "true") {` should work then, right? But then it is a string comparison. Is it possible somehow to convert `$show_song_art` into a correct boolean value?

By the way, I orientated myself in the existing code. It seems that there are a lot of more code snippets with this mistake. As an example have a look at `arts.class.php`, line no. 375: 
`if (AmpConfig::get('write_id3_art')) {`
`           if ($this->type == 'album') {`

Is there anything different to my change that I am missing? Or does this snippet simply have the same problem and should be reviewed within the redevelopment process?

Psy-Virus: > Oha! I didn't know that, sorry. Of course this is a very bad style. How can I change the code properly? if ($show_song_art == "true") { should work then, right? But then it is a string comparison. Is it possible somehow to convert $show_song_art into a correct boolean value?

We don't have to convert something. We can simply use a strict comparison like: `$show_song_art === "true"`

This way we don't have to struggle with how variables cast to the different types as PHP compares the types 'string' and the content.

All this casting stuff is also extremely confusing for me. That's why I always avoid it, if possible. If I see it right, you only get strings back from `AmpConfig::get` so a strict string comparison is the way to go. This way we son't have any problems with miscasts and so on.

You also could simply test it by setting the option to something else than a boolean-string-thingy in the config file. Or try what happens if you set the option to false / "false" ; true / "true" ; 0 / "0" ; 1 /  "1" or bogus.

> By the way, I orientated myself in the existing code. It seems that there are a lot of more code snippets with this mistake. As an example have a look at arts.class.php, line no. 375:
if (AmpConfig::get('write_id3_art')) {
if ($this->type == 'album') {

> Is there anything different to my change that I am missing? Or does this snippet simply have the same problem and should be reviewed within the redevelopment process?

Yepp, seems to have the same problem. But I'm not completely sure about that, as I didn't test it.

Also, as I'm not an expert, everything that I said can be total BS :D

But I think I got it right^^

Greo83: Sorry to bother you again, but I'm afraid you seem to be wrong here. I was not able to get it running with the changes we discussed. So I tested around a bit. In my opinion the `AmpConfig::get` in fact does a convertion into a boolean value. 

I tested the following expressions:

1. Having configured the setting to "true"
- `$show_song_art` === "true"` --> result FALSE
- `$show_song_art` == "true"` --> result FALSE
- `$show_song_art` === 'true'` --> result FALSE
- `$show_song_art` == 'true'` --> result FALSE
- `$show_song_art` == true` --> result TRUE

2. Having configured the setting to "false"
- `$show_song_art` === "true"` --> result FALSE
- `$show_song_art` == "true"` --> result FALSE
- `$show_song_art` === 'true'` --> result FALSE
- `$show_song_art` == 'true'` --> result FALSE
- `$show_song_art` == true` --> result FALSE

3. Having configured the setting to any rubbish
- `$show_song_art` === "true"` --> result FALSE
- `$show_song_art` == "true"` --> result FALSE
- `$show_song_art` === 'true'` --> result FALSE
- `$show_song_art` == 'true'` --> result FALSE
- `$show_song_art` == true` --> result TRUE

Conclusion: 
`$show_song_art` is casted into a boolean variable and cannot be compared to the related text values by a string comparison.
Would you agree or did I do something wrong in my tests? Any missing setting perhaps?

Robb110:
Hi, i'm reading the discussion, and i would really like to have that feature of getting the id3 image art for each song file. So, it's possible to merge those commits into develop?

I just hate that Ampache for each song get the image art from the album and not from the id3 tag embedded in the song file.

wagnered:
> I just hate that Ampache for each song get the image art from the album and not from the id3 tag embedded in the song file.

I didn't have access to the issues when issue 1600 was added.  It has been deleted, so I don't knowtrdetsis that tlhis PR was based  on.

I'm working in this 'area'  working long correctly writing images in the db to files.

wagnered:
> I just hate that Ampache for each song get the image art from the album and not from the id3 tag embedded in the song file.

I didn't have access to the issues when issue 1600 was added.  It has been deleted, so I don't know what this PR was based  on.

I'm working in this 'area' , working on  correctly writing images in the db to files.

The song  is used to gather images, but only images of the type 3 (front cover) and 8 (artist/performer) images. Ampache might not be correctly assigning 'artist' to an image when it gets the image from the file.    Also 'tags' must be in the `art_order` setting.

Are you asking if other image types like $00(other), be allowed? That will take major changes to the interface.

I will check into it
.

Greo83:
@wagnered 

The original issue was a feature request. Ampache's default behaviour is to scan the images contained in the ID3 tags and add them as album pictures. But some users often have songs with own pictures, not related to an album.

With the change I did at that time all images found in the files are treated as song-related images. I adjusted all (hopefully all) places to support this new image relation. This change can be switched on/off by a setting, default setting is off.

wagnered:
Issue #1600 was deleted, so I didn't get to review it for a better understanding of this PR.  I hope you don't mind if I ask you a bunch of questions to fill my gaps in understanding. Also, there might be alternative solutions.

> But some users often have songs with own pictures, not related to an album.

But related to the song or artist in  some way?  There are 21 acceptable  picture type IDs.

Technically, all images are related to a song.  So, if a track has a cover image, artist image and feasably a composer image, which one should be fetched, or should all of them in one api call?  Or am I not understanding the intent of your PR?  Could you give me an example?

Also, the emphasis of your PR seems to mostly be the Subsonic API.  I've been conceptualizing displaying other images on the song's web page.

I appreciate your feedback.

wagnered:
Thank to @kuzi-moto I found the issue and maybe understand the PR better.

If you are up to trying again, I think the Codacy errors can be ignored.  They seem to be about some  JavaScript packages.  Also, by now, they might be fixed by the package creators.

As there are file conflicts, you need to close this PR, create a new local features branch, then push the new branch to your fork.  Select the new branch on your fork and do another PR. This prevents the file conflict problem.

Greo83:
> But related to the song or artist in some way? There are 21 acceptable picture type IDs.

At that time my idea was setting the field "Type" of the pictures table to "Song" instead of "Album" and adjust the select statements in all player, API etc. classes. 

> the emphasis of your PR seems to mostly be the Subsonic API

So I have to disagree this - I just wanted to cover all possible players.

To be honest this PR is almost 2 years old. Meanwhile we are on Ampache version 4. It seems that there have been a lot of changes since then. I'm not sure if my idea can still work now, or as you mention if it is still necessary. When I read between your lines it seems that the structure of the database and especially the images have been changed?

wagnered:
No,  I've done nothing to the image table.  I've only limited tag writing to images of artist,  and covers.  

>  I'm not sure if my idea can still work now, or as you mention if it is still necessary. 

I meant to say that there might be alternative solutions. (I don't write between the lines).  For example, the image table has a `kind` column which defaults to 'default' and is hard coded into the logic. Maybe that could be used somehow.

This PR was brought to my attention about a month ago when someone else showed an interest. That's why I looked into it.

Anyway, as I have other priorities, I'll leave the decision to you.

Regards..

wagnered:
It looks like your local develop branch is not in sync with ampache/Ampache. I'll take  a look and see what can be done to fix this.

Also I wrote above that this PR needs to be closed, a new local (feature) branch  needs to be created,  changes made on the new branch and the new branch pushed to your GitHub fork.

When that is done, the new branch on your fork can be merged with the develop branch of ampache/ampache. 

Doing it this way will prevent file conflicts and duplicate commits.

wagnered:
Okay, here are the steps needed to get your changes into a new branch/PR
1. Checkout develop branch.
2. git remote add upstream https://github.com/ampache/ampache.git
3. git reset HEAD~2
4. git status // to see unstaged files. These  should be your original commited files for PR
5. git stash   // moves unstaged files so they aren't overwritten.
6. git checkout -b \<branch name>
7. git stash pop. // Brigs stashed files backito pay.
8. git add \<files>
9. git commit
10. git push -u origin \<branchname>

On your GitHub fork, change to the new branch and do a **new** PR.  Any questions please ask.

Greo83:
Okay, thank you for your detailled description. I'll try this later.
Yesterday I tried to rebase my fork with the latest changes and did some manual merging, but it seems that something has not worked as I expected. (I'm not so very familar with the Git processes. )

When I have my repo back in synch again I will test the changes and then create a new pull request.

Greo83:
I did the Git steps that you told me without any problem.
But I'm trying to set up a new test environment now, which always fails. Perhaps you have an idea for me?

First I tried to install ampache/ampache to a brand new Xampp installation. First call of install.php shows error "You may have reached this page because Ampache was unable to load the required dependencies". I was unable to find out the problem, dependencies seem to be fine.

Then I tried to use the official docker image. I'm able to start the install routine, but it fails in the step where it reads the configuration files - it claims that they are invalid or not readable.
Then I tried to overwrite the complete /var/www directory in the docker container with my new Git branch. Result: The same error as in the Xampp installation.
Last thing I tried was to use a new docker container and overwrite only my changed files. Result: Same problem as with the unchanged container.

Any help is very appreciated - I don't want to create a pull request for untested features.

wagnered:
Did you read the [Ampache install wiki](https://github.com/ampache/ampache/wiki/Installation#prerequisites)?  The develop version requires Composer to be installed and run from the repo's root folder.  I was able to duplicate your issue by not running composer before starting the install.  I suggest you  restart the XAMPP  based install after reading the wiki.

I can't help with docker(sp?) as it won't run on a non rooted Android.

Greo83:
Sure I did. But executing the composer threw some errors (sorry, I could have mentioned that). After some more research I found out that the latest version of composer meanwhile is something like 2.x. This version does not support the required packages any more. I was able to install composer 1.10.17, which then installed the depencies properly. Perhaps this should be added to the wiki?

Unfortunately now I stuck in the same point as in the unchanged docker version: The http://localhost/ampache/install.php?action=create_config skript claims that "Configuration files were either not found or unreadable." For the file config/ampache.cfg.php the lights for exists and configured turn to red. Creating the file manually crashes the Ampache server completely and disables all sites.

Can you help once more?

wagnered:
I remember  problem with 2.x being reported somewhere.  But you are correct, it should e added to the install wiki. BTW, any member can edit the wiki.

Anyway, 
> "Configuration files were either not found or unreadable."

At first glance, this seems like a permissions problem.  Apache must have permission to write to the Ampache folder.  Apache starts up as root but doesn't run as root.  Since I'm not familiar with XAMPP and I have no idea what your system configuration is I'm not sure what user/group is.

Greo83:
Finished! What an adventure...
At the end I installed a brand new Ubuntu 20.04 virtual machine from scratch and gave full privileges for everybody to the Ampache folder.
With that test environment I was able to finish the installation/configuration and was also able to do successful tests of my changes.

So I can close this pull request now. The new one is already created. I hope, I did everything correct - it was not so easy for me as a Git beginner.

[https://github.com/ampache/ampache/pull/2618](url)

PR #1849, opened 2019-02-02
Prevent spawning of refresh intervals on each nav or refresh

Psy-Virus:
Something that bugged me since ages was that every time you refresh the page or navigate, a new "refresh" interval was spawned through Javascript. This lead to many intervals running at the same time or firing shorter than set in the config.


lachlan-00:
Closing as this has been merged into core.

PR #1856, opened 2019-03-18
Disable video/channel on new install, Matomo.plugin.php, theme updates and php-cs-fixer for php7.3

lachlan-00:
Ampache should be focusing on the music and the musical aspects that work. I propose initially forcing allow_video to 0 and begin feature removal with channels.

Channels do not work well when compared to other streaming options and the ability to simply add an icecast stream to your ampache server negates the need for channels. Channels are also a regular issue that users are unable to easily set up or manage.

Ampache is a music server and should focus on the features relating to music.


lib/class/update.class.php
```diff
@@ -4023,9 +4023,14 @@ public static function update_380013()
     {
         $retval = true;
 
+        // disable video
         $sql = "UPDATE `preference` SET `value`=0 WHERE `preference`.`name`='allow_video'";
```
Psy-Virus: @lachlan-00, I think this should only be done on a fresh installation, as it would confuse users if the video function is suddenly not working anymore.
If my memory is not completely broken; there is a way to achieve this.

a7medo778:
php streaming has improved alot, and with channels msking the creation of "MUSIC" streaming channels a breeze without the dependency on 3rd party platforms like shoutcast/icecast, i think the removal of those would be more appropriate.
with an update to the channels underworking to bring true synced playback (from my current setup testing, channels works flawlessly and easy but suffers from unsynced playback between clients), this can be a feature for ampache as a music player instead of being considered as bloatware

lachlan-00:
> 
> 
> php streaming has improved alot, and with channels msking the creation of "MUSIC" streaming channels a breeze without the dependency on 3rd party platforms like shoutcast/icecast, i think the removal of those would be more appropriate.
> with an update to the channels underworking to bring true synced playback (from my current setup testing, channels works flawlessly and easy but suffers from unsynced playback between clients), this can be a feature for ampache as a music player instead of being considered as bloatware
> 
> TDLR, i am all for removing everything video, but channels are an audio feature that is just awesome

Over the years i've just seen channels have a really high rate of issues with general users. a lot of the open issues that stay open are related to channels failing. have you made changes to channels on your server?

it's music related so yeah that can come back if you really support it.

lachlan-00:
Now that i've created a hodgepodge of commits; here is a summary of changes:
* composer.json updated for php7.3, composer2 and added jquery 2.2
* ampache.cfg.php.dist updated to disable channels by default
* update.class.php allow_video & home_moment_videos = 0 for new installs
* ampache.sql allow_video & home_moment_videos = 0 for new installs
* install.lib.php allow_video & home_moment_videos = 0 for new installs, channels false
* Matomo.plugin.php add plugin from comments in /pull/1571
* show_login_form.inc.php added viewport to html to allow better mobile view.
* show_search_form.inc.php require allow_video to access video search
* show_stats.inc.php require allow_video to see videos in stats.

multiple files:
* use UI::get_logo_url() instead of strings
* use AmpConfig::get('theme_color') to respect light/dark theme icons.
* CSS updates and linting to make things a bit less clunky in login page

In my fork i've tested jquery3, api and subsonic updates from other commits. I've contacted the owner of power ampache to see if they can help with testing and if there are api changes he would like.

a7medo778:
I have it enabled and plan to use it in my next app update as radio source,
its far from perfect specially the synced playback issue, but its a much
better alternative than having to create another instalation instance for
icecast and mapping my catalog there.

I wish i can be helpful of the technical side but i am not that
knowlagdable tbh 😅

lachlan-00:
I think it's good that you're getting something out of channels. i ended up using icecast+liquidsoap as it was more reliable for me.

being able to put up a stream link i think is better than running the channel inside ampache but that's no reason to remove it.

The scary thing that needs updating is the web-player. i think jplayer hasn't been updated in about 5 years now....

lachlan-00:
i'm closing this pull to make one without 40000 commits

Issue #1861, opened 2019-03-24
Unable to add TWIT Podcast - Data too long for column 'copyright' at row 1


zerojay:
# Environment

 * Ampache version: 3.9.0
 * Web server + version: Apache 2.4
 * Server operating system: Ubuntu Server 18.04.2 LTS

 * Client type:
   - Web

# Settings

Settings don't matter here.

# Description

Attempting to add This Week in Tech podcast doesn't work, Ampache does nothing when attempted. URL: http://feeds.twit.tv/twit.xml

# Logs

2019-03-24 22:07:11 [jay] (Core) -> Registered post form add_podcast with SID aba62970baa5ff4daf0a5509a7054ed4 and expiration 1553468831 (3600 seconds from now)
2019-03-24 22:07:16 [jay] (session) -> 6ngn02ob7kv9re8k5if51h5337 has been extended to Sun, 24 Mar 2019 23:07:16 +0000 extension length 3600
2019-03-24 22:07:16 [jay] (session) -> Session created: 7c0811c5f1f51a4bb775bc5963e85dd8
2019-03-24 22:07:16 [jay] (Core) -> Verified SID aba62970baa5ff4daf0a5509a7054ed4 for post form add_podcast
2019-03-24 22:07:16 [jay] (Dba) -> Error: ["22001",1406,"Data too long for column 'copyright' at row 1"]
2019-03-24 22:07:16 [jay] (Dba) -> Error: ["22001",1406,"Data too long for column 'copyright' at row 1"]
2019-03-24 22:07:16 [jay] (Dba) -> Error: ["22001",1406,"Data too long for column 'copyright' at row 1"]
2019-03-24 22:07:16 [jay] (Dba) -> Error: ["22001",1406,"Data too long for column 'copyright' at row 1"]

----

Looking at the Ampache source and the feed, the reason this is happening is pretty obvious.

From the RSS feed:
<copyright>
This work is licensed under a Creative Commons License - Attribution-NonCommercial-NoDerivatives 4.0 International - http://creativecommons.org/licenses/by-nc-nd/4.0/
</copyright>

This is 166 characters long.
-----
https://github.com/ampache/ampache/blob/develop/sql/ampache.sql

CREATE TABLE IF NOT EXISTS `podcast` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `feed` varchar(4096) NOT NULL,
  `catalog` int(11) NOT NULL,
[...]

PR #1862, opened 2019-03-24
Increase copyright column size to fix issue #1861

zerojay:


lachlan-00:
that will only update new installs. This should go into a database update instead.

Check out lib/class/update.class.php and make your updates there.

First you need to give it a version number and a string explaining the db update
```
        $update_string = "- Fix copyright max size in the podcast table.<br />";
        $version[]     = array('version' => '380013', 'description' => $update_string);
```     

Then at the bottom you do the sql part
```
    public static function update_380013()
    {
        $retval = true;

        $sql = "ALTER TABLE `podcast` MODIFY `copyright` VARCHAR(255)";
        $retval &= Dba::write($sql);
        return $retval;
    }
```

Psy-Virus:
Hey @zerojay,
your PR looks reasonable and safe to merge. If you would change it from an "only on new installation" fix to a "Fix via database update" as @lachlan-00 described, I'd merge it.
Since the Laravel Branch is currently not developed, I guess there's no problem if I use my new permissions to merge also bug fixes and small enhancements, as long as they follow the standards @Afterster and @wagnered introduced.

zerojay:
Hi, sorry for the lack of response on this. I've been extremely busy. I've made the updates to the pull request as... well, requested.

lachlan-00:
@psy-virus I think you're the only person with push who has been active this year.

The update from @zerojay looks good to me. Big thumbs up for contributing.

Psy-Virus:
Looks good. Just one last thing: Get rid of the first commit. :)
As the Database gets updated while Ampache is installed the first time, the first commit (where you have changed the database directly) is useless.
We only need the latest one.
You have some options now. You can reset your local branch to the last commit before your additions by doing: `git reset <commit ID> --hard`, readd your changed `update.class.php` and do a force push to your repository like: `git push origin develop --force`.

Or you do what ever is written [here](https://sethrobertson.github.io/GitFixUm/fixup.html), that helps you solving this.^^

The 3rd option would be that I commit the changes from your repo. But in the history would be written that I committed it with your authorization.
I'd prefer one of the first options, as it keeps the history clear.

:)

lachlan-00:
Hey @zerojay 
Please make a new pull on the core branch
https://github.com/ampache/ampache/tree/core

because this is a DB update i'll do it separately with the other coding work and then you'll get the cred when merged back into dev/master

Thank you for patience with this one. Things are moving!

lachlan-00:
@zerojay you can now merge into the core branch. You'll probably have to close this one as i don't think it will let you change branches. But i'd really like to have your contribution recorded instead of me just putting it in.

https://github.com/ampache/ampache/tree/core

make your update 400001 and i'll merge asap.

lachlan-00:
Merged into my core branch and this will merge into the official branch soon so this pull isn't needed now.

Thanks for your efforts and you're now an Ampache Contributor!

PR #1863, opened 2019-03-26
Fix autoplay blocking ampache/issues/1858

lachlan-00:
load the ajax webplayer div into the default page so browsers don't think the new pages are untouched by the user.


lachlan-00:
Closing. Focusing on amuzak.

PR #1866, opened 2019-03-30
add a default profile picture for users

lachlan-00:
Currently there isn't a default user avatar.
This CC picture functions the same way as blankalbum.png.

Can't find an avatar? Fallback to this.


lachlan-00:
Closing. Going to focus on my fork instead

PR #1867, opened 2019-03-30
fix user sorting

lachlan-00:
The user list header column is in the wrong order and also doesn't sort my username. only fullname.

this commit fixes both


Psy-Virus:
Nice small catch :D
Looks good and works.
Thanks for that!

PR #1869, opened 2019-04-15
fix bug in search, add myplayed to search options

lachlan-00:
The myrating search didn't play well with others and i found the culprit. (me!)

I renamed myratings to myrating to match up with the regular rating and fixed the join part.

This fixes the ratings search and adds in a search for "Played (Mine)"
Whether you yourself have played the track instead of everyone.


lachlan-00:
pull/1873 is better than this one

PR #1873, opened 2019-04-28
Add artist and album play/rating searches to song

lachlan-00:
Close pull #1869,

This is a better search update.


lachlan-00:
Forgot I removed video in amuzak.
I'm going to focus on my fork instead.

Issue #1872, opened 2019-04-25
Multiple vulnerabilities in Ampache 3.9.1-develop 


TarlogicSecurity:
Hi!

We found multiple vulnerabilities that, chained, could lead to a full compromise. The product affected is Ampache 3.9.1-develop. Can you provide us a mail address where we can send you the details?

As a summary:
- SQL injection
- Account takeover via stored XSS + CSRF
- Few other security related issues

Best regards

vollmerk:
@TarlogicSecurity - I'm not longer involved in the project, but I care about it (that's why I still pay for the domain name personally) If you don't hear from someone quickly feel free to reach out to me at karl.vollmer@gmail.com and I'll see if I can help.

@Afterster - are you still involved, I haven't paid attention... I just pay the domain bill haha :)

lachlan-00:
I can help out with a pull but I don't have merge.

lachlan-00:
@vollmerk the only active person this year with merge has been @Psy-Virus

TarlogicSecurity:
Hi!

We just sent to you (@vollmerk ) a mail with the details. We hope this can be helpful for you.

lachlan-00:
@TarlogicSecurity I've started working on the security issues in Ampache. Thank you for bringing this forward and I'll be working fast as I can to bring Ampache up to scratch.

Here is my new codacy repo where i'll be working on the project
https://app.codacy.com/project/Ampache/ampache/dashboard
![image](https://user-images.githubusercontent.com/1305249/57112936-aeaa8700-6d85-11e9-8c7f-7d1117ee8d02.png)

VS the current codacy.
https://app.codacy.com/project/afterster_2222/ampache/dashboard
![image](https://user-images.githubusercontent.com/1305249/57113009-e7e2f700-6d85-11e9-9266-30e3a3bf4190.png)

TarlogicSecurity:
Hi!

We just received a mail from MITRE with the CVEs assigned:

- CVE-2019-12385 for the SQL Injection
- CVE-2019-12386 for the persistent XSS

Maybe you want to include them in the changelog when you release a new version.

Best regards.

lachlan-00:
Thank you, i'm bugfixing now so we can hopefully close these holes soon.

stebe:
Now that core is in develop, are these security issues resolved?

lachlan-00:
The issues @TarlogicSecurity found are fixed but I was just keeping it open till I can double check the list tonight

[...]

PR #1890, opened 2019-05-15
Core update 5

lachlan-00:


admin/access.php
```diff
@@ -29,7 +29,7 @@
 
 UI::show_header();
 
-switch ($_REQUEST['action']) {
```
Psy-Virus: Aren't those changes overkill, as we already validate the incoming data hard coded by with switches?

What I can see is that there is that we slow down the code execution by adding a function that doesn't improve the security in those cases.

The filter_input function would be interesting if we had some user controlled output, like a message or a post that's saved in a database, unfiltered/unsanitized. But in these cases I don't think it's an improvement, as code is only executed if the POST or GET request matches any case. If there's no match, the default case will handle the request or if there's no default case, nothing will happen.

lachlan-00: Thanks for reviewing.

I haven't compared load times between the branches yet. But I do agree that I haven't checked whether all the filters are needed.

There are xss and injection abilities in develop for pretty much every page right now.

Once I've made all the code changes in core part of my process will be on testing between core <-> develop as well as speed of those functions.

lachlan-00:
@psy-virus I see what you mean with the switch. The PHP manual leads me to believe that switches will only execute on the case, not the variable so because the cases are hard coded strings they shouldn't be vulnerable.

https://www.php.net/manual/en/control-structures.switch.php
```
 Only when a case statement is found whose expression evaluates to a value that matches the value of the switch expression does PHP begin to execute the statements.
```

Psy-Virus:
Precisely ^^
I believe that those automatic checks that you probably have run (or they: #1872) to find vulnerabilities are not made for those cases and create false positives.

But all in all, good job so far. It's nice to have someone who has, before all, the time and the motivation to push Ampache and is not afraid of learning something new.

:)

lachlan-00:
Thanks for the positive thoughts! I'll update my branch and I can filter the returning issues I know will be safe.

I'm really happy to push forward with all the "boring" things that haven't really moved over the years. It's allowed me to do a few things that have been bugging me and the Ampache project itself has been really important to me personally.

I've made a lot of the changes I wanted to make in the modernization project so I'm hoping in a month I should be able to start my testing plan to move core into develop.

PR #1891, opened 2019-05-16
adding year to several views/ gather art when update from tags (RE:PR)

noyse27:
adding column/row Year on:

- album
- song
- nowplaying
- recently played

making albums searchable with year from nowplaying year

gather art when updating album from tags

change last.fm scheme
from https to http as described in Last.fm scrobbler api
https://www.last.fm/api/scrobbling 

> As with all our write web services, requests must be sent as HTTP POST requests to http://ws.audioscrobbler.com/2.0/ with form urlencoded parameters in the body of the request. The text encoding must be UTF-8.


lachlan-00:
I've always thought about adding this. Thank you.
I'll send you a few changes soon once I get it on my server but this should be good to go with minimal effort

lachlan-00:
I haven't used last.fm for a while but let me do some work on that end as well to ensure it's okay.

Visually it works well. It would be good to replace intval() with (int) to keep it in line with the core branch. but the code isn't an issue.

Great first contribute!

lachlan-00:
Another tip would be to use switch instead of all the ifs in your get_by_year function but I'll do that in the core branch now it's in

noyse27:
thank you :)

lachlan-00:
Thank you for becoming an Ampache contributor! It's really appreciated by everyone here.

This code is in the core branch now and will merge into develop soon. Closing as it's not needed.

PR #1919, opened 2019-06-13
General Translation Update June 2019

Psy-Virus:
As usual. Updated translations from Transifex

Thanks to all the Translators that made this possible!

:)


lachlan-00:
I'm glad you know how transifex works!

I'm going to try and cut down on some of the duplicate strings. Especially those that are just different case over the weekend to make it a bit cleaner

PR #1920, opened 2019-06-13
Fixing some bugs related to Music Clips implementation

alejandroliu:
Hello,
This PR fixes the Music Clip implementation in Ampache.  Specifically it fixes an issue whereby Catalogs configured to contain "Music Clips" do not import any videos as "Music Clips", rather they are imported as a plain videos, without related artist information.

Also, it address an issue whereby the video title was escaped twice.

* * *

When Music clips are discovered and added, the method `insert` in the `clip` class is called with the metadata tags.  This method expects that the passed `$data` array to contain an `artist`
and `song` values.  Because these are inserted directly into the SQL table, these actually
should be integers representing an `artist id` and `song id`.

The metadata code that populates this `$data` array actually does not know anything about this, and instead passes a string for `artist` and does not populate `song`.

This PR will actually do the look-up to convert the `artist` string into `artist_id` and the song `title` into a `song_id`.   It is not really clear to me, how the `song` id is used, it it is used at all.

* * *

When the music clips are displayed, they were showing the URL for the artist page instead of the artist name.  This PR address that bug.

* * *
Lastly, titles of music videos where escaped twice.  So video titles containing special characters would not look right.


lachlan-00:
Thank you for your efforts! I don't use videos personally but I'll test this out asap. I'm also close to a big merge from the core branch so that might happen first.

lachlan-00:
Can you run php-cs on this and update?
```
./php-cs-fixer -v fix
```

alejandroliu:
I have additional changes related to viewing video thumbnails.  Would you prefer to add them to this PR or should I open a new PR?

lachlan-00:
If it's more changes in the video class you might as well combine them into this one

lachlan-00:
Thanks @alejandroliu please run the codestyle fix again and i will test on my branch asap.

Really great to have someone new looking at the videos part of ampache!

alejandroliu:
Really can't believe that I missed that one.  I thought I run trough code fixer already.

lachlan-00:
Hey don't worry, it's amazing what you miss sometimes. Thanks for updating and I'll play around with this tonight on my dev server

lib/class/video.class.php
```diff
@@ -255,7 +255,7 @@ public function format($details = true)
         $this->f_title      = scrub_out($this->title);
         $this->f_full_title = $this->f_title;
         $this->link         = AmpConfig::get('web_path') . "/video.php?action=show_video&video_id=" . $this->id;
-        $this->f_link       = "<a href=\"" . $this->link . "\" title=\"" . $this->f_title . "\"> " . $this->f_title . "</a>";
+        $this->f_link       = "<a href=\"" . $this->link . "\" title=\"" . scrub_out($this->f_title) . "\"> " . scrub_out($this->f_title) . "</a>";
```
alejandroliu: Is this at all necessary?

In line 255, we do: `$this->f_title = scrub_out($this->title);`

And in line 258 we then do: `$this->f_link = ... scrub_out($this->f_title)` ... essentially it means that we are doing:

`scrub_out(scrub_out($this->title))`

Also, I am not sure if it s indeed fixed in core because in my set-up, I would see the names of clips showing with the re-escaped characters.

lachlan-00:
You're right about the scrubbing, i didn't see the line above in the PR pages

Can you show me a screen of which characters are escaped?

lachlan-00:
Another thing could be your characters are not UTF-8 characters?

The default charset is:
```
site_charset = UTF-8
```
scrub_out returns the string based on your character set in the config
```
return htmlentities($string, ENT_NOQUOTES, AmpConfig::get('site_charset'))
```

lachlan-00:
using filter_var for scrubbing instead of scrub_out. things like umlaut, etc were being encoded.

lachlan-00:
This has been committed through the core branch and you are now an ampache contributer!

PR #1936, opened 2019-06-21
Update database for older users

lachlan-00:
This seems to have affected people using ampache since 3.6 when subcategories were introduced.

@apastuszak can you tell me what options are missing for you comparing your old preferences table to the new one?

I would also compare the two user_preference tables as well.


lachlan-00:
72/shoutcast_active seems to be the only other one that doesn't come in from a DB update as well. will check up tomorrow.

PR #1961, opened 2019-07-01
Stringsupdate

kuzi-moto:
Updated lots of stuff. Feel free to undo some if they are wrong.


kuzi-moto:
Oh yeah, meant to mention the changes to sidebar_modules.

I thought it was a bit redundant to have "Catalog Modules" as the title, and then also have "modules" in the item names. Thoughts?

lachlan-00:
excellent! i'll have a read and pull in.

lachlan-00:
It's a little redundant to have modules but it also seems to follow existing menus for other sections like in the admin page?
image

But i do think this might look good though.
image

kuzi-moto:
Ah you're right, I didn't think about the other sections. If we were to move modules to admin page, it might make sense to keep its previous form. Well feel free to change it however it makes sense.

PR #1962, opened 2019-07-02
Update install.lib.php for mysql legacy auth

lachlan-00:


lachlan-00:
Actually not going to work on mysql8 for the foreseeable. still pre-release and not important

PR #1963, opened 2019-07-02
Strings Update 2: Electric Boogaloo

kuzi-moto:
I know that Psy-virus was already notified to review strings, but I found some more...

If it's too late, that's okay!

By the way, I noticed some trends and tried to follow this convention:

UI::show_box_top - Title Case
AmpError::add - Sentence case
show_confirmation() - Sentence case - though this one is unique, because you can supply a title and a body, and sometimes there is no body included. So I guess it's on a case-by-case basis?

Anyway, if you agree then maybe we can add it to the Wiki so that new contributors can follow a standard. Also, with the errors, I appended "Error:" to all of them, since it seemed to make sense to do that and a lot of them already do, or some start with "Error" and no colon.


lachlan-00:
i'm happy with this. adding now.

PR #1975, opened 2019-07-15
.tag_actions to show on hover of .tag_container

kuzi-moto:
I'd like to propose a change to the Tag Cloud page, so that the tag actions only show when you hover over the button container. When you have a lot of tags, the actions are pretty distracting and make it more difficult to read the tags.

image

But, if we set the buttons to be hidden by default, the result is a much cleaner-looking interface:

image

Then, when you hover over a button its actions will show, allowing you to click on it as you would normally:

image

### DIFF
--- a/themes/reborn/templates/default.css
+++ b/themes/reborn/templates/default.css
@@ -174,7 +174,7 @@ input[type=button]:focus:active, input[type=submit]:focus:active {
 
 #ajax-loading {
     background-repeat: no-repeat;
-    background-position: right; 
+    background-position: right;
     display: none;
     font-size: 12px;
     left: 52px;
@@ -322,7 +322,7 @@ input[type=button]:focus:active, input[type=submit]:focus:active {
     width: 128px;
     height: 128px;
     background-repeat: no-repeat;
-    background-position: center; 
+    background-position: center;
     background-size: contain;
     margin: auto;
 }
@@ -1210,6 +1210,11 @@ span.fatalerror {
 
 .box-content #tag_filter .tag_actions {
     margin: 15px 0 0 -8px;
+    visibility: hidden;
+}
+
+.box-content #tag_filter .tag_container:hover .tag_actions {
+    visibility: visible;
 }
 
 .box-content #tag_filter .tag_button span {


lachlan-00:
It does look nicer. The whole tag page is really out of step with everything else but this looks like a stsrt

stebe:
Looks good and much cleaner. I'm also looking to be able to select multiple tags so I can delete them. Now I have to do it one by one and wait for the page to refresh each time.

lachlan-00:
Do the tags you delete come back or are they left over from renaming? If they're left overs I should look into whether the garbage collection does this.

kuzi-moto:
I just tested it out, and it seems that when you delete a tag, it will delete it from all the music as well. The only way to get it back for me was to update an album from its tags.

Currently you can only delete one tag at a time, where the page refreshes after each one. It seems what stebe wants is to be able to select more than one tag at a time, then delete them all at once? Though deleting one at a time doesn't seem to take long for me, the page reloads in less than a second, it would just be a bit tedious if you had lots of tags you wanted to remove.

I would really recommend that if there are many undesired tags, to use something like [beets](http://beets.io/) to re-tag music. It has a very nice way of handling genre tags. Like if you had tons of electronic music you could instruct beets to classify ambient dub, ambient industrial, dark ambient, and drone music as all just "Ambient", but I digress...

As a side-note, perhaps we could maybe change "Tag Cloud" to "Genres" perhaps? I've always thought that was weird, as I can't find any other examples of places where "tags" are used in place of "genres". Especially as the term "tags" could also refer to "ID3 tags".

lachlan-00:
tag/genre confuses me a lot but i don't think it would be a great benefit to changing it especially with all the pending string changes. it would also require changing the database table name to keep up.

kuzi-moto:
Ah, I suppose that makes sense. It's a pretty small critique anyway. Probably not much gain for all the work involved

lachlan-00:
just testing it out now, i think  changing the padding from top/left to top+bottom/right would make the hovering a bit easier

Current padding here
image

maybe this?
```
.box-content #tag_filter .tag_button {
    margin: 10px 10px 10 5px;
```

kuzi-moto:
I can't test it out, but that seems like it should work! I was also thinking, maybe it would make more sense to change the layout from a cloud? Like you said, it's a bit out of place from the rest of the interface.

I was thinking of something more like three columns making them into vertical lists? I think it would match better, and improve readability because currently it's kind of hard to browse through the cloud.

lachlan-00:
it might be easy enough to set each container area as a percentage  width so it forces a bit more alignment.

stebe:
> Currently you can only delete one tag at a time, where the page refreshes after each one. It seems what stebe wants is to be able to select more than one tag at a time, then delete them all at once? Though deleting one at a time doesn't seem to take long for me, the page reloads in less than a second, it would just be a bit tedious if you had lots of tags you wanted to remove.
> 
> I would really recommend that if there are many undesired tags, to use something like [beets](http://beets.io/) to re-tag music. It has a very nice way of handling genre tags. Like if you had tons of electronic music you could instruct beets to classify ambient dub, ambient industrial, dark ambient, and drone music as all just "Ambient", but I digress...
> 

This is usually after I update my music files to remove unwanted tags, are misspellings or incorrectly delimited that creates a tags I don't want. So if I want to remove 15 to 20 tags that are there, doing it in one shot would be a bonus, although at this point my catalogs are pretty clean so it would not be a feature I would be using much.

PR #1976, opened 2019-07-16
Only show the "Video" header on the sidebar when there are videos

kuzi-moto:
I think that the sidebar should only show the "Video" header on the sidebar when the user has a video catalog with videos in it.

If you have enabled video features in Ampache server settings, then you will get the "Video" section in the sidebar, even if you do not have any video catalogs. I don't think this makes any sense, as it distracts the user by cluttering the interface with something that appears relevant, but does nothing.

image

This is a fairly simple fix, adding a check to see if there are any videos in the database before showing the heading.

Change with video:

image

Change without video:

image

### DIFF
--- a/templates/show_search_form.inc.php
+++ b/templates/show_search_form.inc.php
@@ -40,7 +40,7 @@ UI::show_box_top(T_('Search Ampache') . "...", 'box box_advanced_search');
     } else {
         echo T_('Artists');
     } ?></td>
-    <?php if (AmpConfig::get('allow_video')) {
+    <?php if (AmpConfig::get('allow_video') && Video::get_item_count('Video')) {
         ?>
         <td><?php if ((string) filter_input(INPUT_GET, 'type', FILTER_SANITIZE_SPECIAL_CHARS) != 'video') {
             ?>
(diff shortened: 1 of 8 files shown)


lachlan-00:
I have video disabled usually unless testing so does the tag cloud just present an empty page with no video catalogs?

I wonder if we add video counts to all the allow_video sections as well?
It might be a simpler way of hiding video features if they aren't used without having to disab;e the whole feature in config.

kuzi-moto:
Actually I'm pretty sure it's broken for video. It just shows all the music tags.

If you think that's a better way, sure! I was just doing this one for the most simple fix :)

lachlan-00:
I would think a lot of people leave the default where allow_video was enabled but might not even care that the feature is there.

I think sidebar, stats pages and search are the 3 places off the top of my head that could probably have this change.

kuzi-moto:
Yeah, I could add it to each of the sections using the allow_video config setting. Alternatively, should video be disabled by default?

kuzi-moto:
I think this cleans up all the other instances of Video showing up in the interface.

lachlan-00:
cool! i'm going to enable allow video and have a browse around.

the visuals are probably the only reason that i disable that option so i think this is a good default to set.

Issue #1974, opened 2019-07-15
sql error generating playlist stats


ftoledo:
file: templates/show_stats_recent.inc.php
```
$sql    = Stats::get_recent_sql('playlist', $user_id);                                                                                                                                                          
$browse = new Browse();                                                                                                                                                                                         
$browse->set_type('playlist', $sql);                                                                                                                                                                            
$browse->set_simple_browse(true);                                                                                                                                                                               
$browse->show_objects();                                                                                                                                                                                        
$browse->store();     
```
this block call get_recent_sql and the generted query have error on "AND GROUP BY":
the another previous blocks just work.

```
SELECT DISTINCT(`object_id`) as `id`, MAX(`date`) FROM object_count WHERE `object_type` = 'playlist' AND  GROUP BY `object_id` ORDER BY MAX(`date`) DESC, `id`
```

Issue #1973, opened 2019-07-15
sql error al get_top


ftoledo:
miss space between userid and the AND statement 

> AND `user` = 1AND (SELECT COUNT

```
2019-07-15 10:25:12 [ragnarok] (browse.class) -> Show objects called for type {album} 
2019-07-15 10:25:12 [ragnarok] (stats.class) -> get_top SELECT object_id as `id`, COUNT(*) AS `count` FROM object_count WHERE `object_type` = 'album' AND `user` = 1AND (SELECT COUNT(`song_dis`.`id`) FROM `song` AS `song_dis` LEFT JOIN `catalog` AS `catalog_dis` ON `catalog_dis`.`id` = `song_dis`.`catalog` WHERE `song_dis`.`album`=`object_id` AND `catalog_dis`.`enabled` = '1' GROUP BY `song_dis`.`album`) > 0 AND `count_type` = 'stream' GROUP BY object_id ORDER BY `count` DESC  
2019-07-15 10:25:12 [ragnarok] (dba.class) -> Error: ["42000",1064,"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SELECT COUNT(`song_dis`.`id`) FROM `song` AS `song_dis` LEFT JOIN `catalog` AS `' at line 1"] 
2019-07-15 10:25:12 [ragnarok] (dba.class) -> Error: ["42000",1064,"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SELECT COUNT(`song_dis`.`id`) FROM `song` AS `song_dis` LEFT JOIN `catalog` AS `' at line 1"] 
2019-07-15 10:25:12 [ragnarok] (dba.class) -> Error: ["42000",1064,"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SELECT COUNT(`song_dis`.`id`) FROM `song` AS `song_dis` LEFT JOIN `catalog` AS `' at line 1"] 
2019-07-15 10:25:12 [ragnarok] (dba.class) -> Error: ["42000",1064,"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SELECT COUNT(`song_dis`.`id`) FROM `song` AS `song_dis` LEFT JOIN `catalog` AS `' at line 1"] 
2019-07-15 10:25:12 [ragnarok] (browse.class) -> Show objects called for type {album} 
[...]

Issue #1971, opened 2019-07-12
XML-API: flag action seems that it does not work.


matiasdelellis:
## Environment

* Ampache version: 4.0.0-develop (Commit: https://github.com/ampache/ampache/commit/7246ddc6da29920baedc69bca065bc35ebb06968 )
* Web server + version: Ngnix
* Server operating system: Debian
* Client operating system: Fedora 31

## Client type:
  * @pragha-music-player over XML-API

## Description

Some time ago I implemented the reproduction of Ampache music on Pragha
Now with the latest API additions, I'm trying to handle the favorites.

API:
 * https://github.com/ampache/ampache/wiki/XML-methods#flag

Call on Pragha:
 * https://host/server/xml.server.php?action=flag&type=song&id=100003&flag=1&auth=TOKEN

Api response:
```
<?xml version="1.0" encoding="UTF-8" ?>
<root>
	<success />
</root>
````
Result:
 * I open the browser, I enter ampache, and I do not see the favorite anywhere. I can add or remove many songs, but I never see it in the browser. :disappointed: 

## Logs

I do not have access to the server to do internal tests.. Please @ftoledo. You can test and append it here?

Thanks.

**Comment by matiasdelellis** [comment]
p.s: I try to get the user's favorites ..

API: stats

Description: (https://github.com/ampache/ampache/wiki/XML-API)
> stats
> 
>    type (Type of album stats to retrieve [newest|highest|frequent|recent|flagged|random])
>    username (Username filter, only if type is recent for now. Optional)

Example: 
 * https://HOSTS/server/xml.server.php?action=stats&type=song&filter=flagged&auth=TOKEN
 * https://HOSTS/server/xml.server.php?action=stats&type=song&filter=flagged&username=matias&auth=TOKEN

Description
 * I I do not pass the username (Which says it is not implemented), The api returns the favorites of all users, and there is no way to filter your own. If it is not 'implemented', it can be a valid answer, but please ... can it be implemented ???? :smile: 
 * As it is an optional parameter, I add it with the hope of fixing it later, but server response the favorite albums of all the users. which is wrong since i query for songs. :disappointed:

**Comment by lachlan-00** [comment]
It flags for me in the website. if you can't get it immediately over the api maybe it's cached?
```
https://ampache.org/server/xml.server.php?action=flag&auth=AUTH&version=350001&type=song&id=94761&flag=1
```

for stats it's not passing the user but the Userflag::get_latest method supports it so lets add it!

[...]

PR #1978, opened 2019-07-17
extend stats in api, bugfixes

lachlan-00:
fix for #1973
fix for #1974
extra code checks for #1971


lachlan-00:
@matiasdelellis i'll be merging this after i verify that stats doesn't break when using the old way compared to the new version but there is a lot more work to be done in the api.

matiasdelellis:
Great. Looks good .. :wink: 
Probably today or tomorrow i have access to the server to apply it and test.

Thanks

Issue #1990, opened 2019-07-25
Registration confirmation page says user will get activation code even if not required


kuzi-moto:
# Issue Template

## Environment

* Ampache version: 4.0.0-develop
* Client type: Any

## Settings

```INI
allow_public_registration = "true"
user_no_email_confirm = "true"
```

## Description

There are no checks in the confirmation page if `user_no_email_confirm` is set, so the message states the user will get an activation key, even though they can login after this point.

Another easy one, I can do this too.

lachlan-00:
You're assigned!

If you want to put all these three into one pull we can merge together.

kuzi-moto:
Nice, sounds good! I probably could have made one issue, but wasn't planning on running into so many bugs.

lachlan-00:
It's good to keep them separate as they are sort of different and the same so we'll need to make sure they all work as intended.

One thing i've found is that fixing bugs in ampache generally leads to new ones!

Issue #1988, opened 2019-07-25
User registration pages need style work


kuzi-moto:
# Issue Template

## Environment

* Ampache version: 4.0.0-develop
* Client type: Any

## Settings

```INI
allow_public_registration = "true"
```

## Description

When user registration is enabled, and you use the user agreement, the user agreement is overlapping with the text "Registration..." underneath the Ampache logo. Also, the checkbox to accept the User Agreement is centered and on it's own line from "I accept"

![image](https://user-images.githubusercontent.com/17068090/61897295-614d3a00-aedc-11e9-8483-7378814233ec.png)

Additionally, after a user has registered, it gives a super ugly page with minimal styling and some more overlapping text:

![image](https://user-images.githubusercontent.com/17068090/61897563-ecc6cb00-aedc-11e9-8623-be600fe41d6a.png)

This is a pretty simple and minor fix, so I can tackle this one sometime when I'm free.

Issue #1989, opened 2019-07-25
Registration allowed/no error messages when creating a new account and email is disabled


kuzi-moto:
# Issue Template

## Environment

* Ampache version: 4.0.0-develop
* Client type: Any

## Settings

```INI
allow_public_registration = "true"
;mail_enable = "false"
```

## Description

The issue here is that when enabling user registration, but not configuring the email settings, then someone can still register for an account. On registering they get a message stating they will receive an email with an activation code, but will not get it. Nor will the administrator see an error in the logs stating that someone tried to register an account.

It may be a good idea to instead show an error on the registration page, preventing someone from creating an account, show an error after creating an account, or at least mention that in the log file.

lachlan-00:
i'll put some error checking for these into your pull for the other registration issues

PR #1994, opened 2019-07-26
Fix registration page styles

kuzi-moto:
These fixes should solve the problems in #1990 and #1988, and I have to say, it looks beautiful now! Unfortunately it probably wont get much use but such is life.

Some shots of what it looks like:

image
image
image

### DIFF
--- a/register.php
+++ b/register.php
@@ -127,7 +127,7 @@ switch ($_REQUEST['action']) {
             AmpError::add('password', T_("Your passwords do not match"));
         }
 
-        if (!User::check_username((string) $username) || $username == null) {
+        if (!User::check_username((string) $username)) {
             AmpError::add('duplicate_user', T_("Error Username already exists"));
         }
 
--- a/register.php
+++ b/register.php
@@ -25,7 +25,7 @@ $_SESSION['login'] = true;
 require_once 'lib/init.php';
 
 /* Check Perms */
-if (!AmpConfig::get('allow_public_registration') || AmpConfig::get('demo_mode')) {
+if (!AmpConfig::get('allow_public_registration') && !Mailer::is_mail_enabled()) {
     debug_event('register', 'Error Attempted registration', 2);
     UI::access_denied();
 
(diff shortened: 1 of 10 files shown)


kuzi-moto:
One thing I wanted to look into further, is if we enable administrator activation, as well as email verification, if Ampache will actually require both before the user can sign in.

lachlan-00:
I'll enable registration and confirm.

Users would start disabled so enabling both would probably mean that whichever does it first would enable them?

kuzi-moto:
If both are enabled, would we want to make sure that both are done before they can log in? Otherwise, just throw a notification or something saying that they need to verify the email still. I can see reasons you would want to ensure valid emails.

Also, this needs a little tweaking. One of my changes to the css broke things once logging in.

lachlan-00:
It might be better to not enable the user when these are set then notify the admin somehow. (A PM +/or email)

So, User registers -> email sent to user -> user confirms -> admin notified to manually enable the user as they have confirmed their email.

kuzi-moto:
Okay, I have completed all my changes and pretty much overhauled much of the styles for the login, registration, and lost password pages. Here's a list of some tweaks:

* Generally simplified the CSS in default.css
* added `.alert-danger` class to actually change the color of an alert. Before the text sort of blended in with other text
* Looks much better on mobile devices
* Changed footer for lost password page to mach others
* Added color to asterisk on registration page
* Login page: made login button same line as "remember me" checkbox. Moved "Lost password" and "Register" buttons below the login.
* Did the same thing for register page, and accepting the user agreement.
* lots of little style tweaks to register page.
* Changed the confirmation text for registration depending on what was set in config.
* Possibly some other small things.

lachlan-00:
run cs over that and you're done.

I'll need to sit down tonight and get all this recent stuff merged as well as the extra checks in #1989

kuzi-moto:
I did, for some reason it's not showing here :/

lachlan-00:
okay that's no problem, i'll pull it onto my debian desktop at home when i add in the 1989 checks.

lachlan-00:
it's fixed in the repo so it's travis that's locked.

lachlan-00:
very slick looking now. The pages won't be shown unless mail_enabled is also true which means it must be manually enabled to allow registration.

PR #1998, opened 2019-07-30
Sorry, lots more updates

kuzi-moto:
Updated all the errors to remove "Error:" as it kind of looks nicer that way. Simplified some strings, combined others, grammar fixes, various spelling errors.


lachlan-00:
feel free to add to gitignore if you have other tools in use.

PR #2004, opened 2019-08-06
Allow cells in Debug page to wrap long lines

kuzi-moto:
Noticed on the Debug page, that if there is a really long line that it creates a scroll bar and pushes the width of the table larger than the page.

I think it should be okay to wrap very long lines so that it removes the scroll bar and keeps the page looking nice.


lachlan-00:
looks much nicer. :)

PR #2006, opened 2019-08-06
Fix header

kuzi-moto:
I cannot remember why I added this... but it looks like it breaks the header from one of the latest commits.


lachlan-00:
this one?
image

Do you know what it's doing? i can't really see anything out of the ordinary

kuzi-moto:
Yeah, for some reason I didn't see it right away but when logged in the search bar section is aligned to the left, rather than the ride side.

lachlan-00:
Ahh, now i see it.

want to stick it in #2004 and i'll merge it tonight?

lachlan-00:
closing, put the revert in the other open request editing default.css

PR #2010, opened 2019-08-10
viewport and page style fixes

### LINKED ISSUE #2007: Feature Request: Change viewport to device-width
## Description

I was thinking about a mobile interface for Ampache, and on looking into it, found that header.in.php has `<meta name="viewport" content="width=1024, initial-scale=1.0">`. To support mobile layouts, I think it needs to be changed to `<meta name="viewport" content="width=device-width, initial-scale=1.0">`. I think there will most like need to be a change or two to reborn, because probably screens less than 1024 will look strange.

Just wanted a second option before screwing around with it.

kuzi-moto:
So this originally started out as just setting all instances of viewpoint to device-width, but then got sidetracked. I also fixed css for an error page for when the config file is not parseable, moved install-doped.cs to insall.css, I also changed some strings, maybe should have left it for the strings update, but realized that  "Ampache" doesn't need to be translated, and that "for the love of music" should. And some other fixes 
This was originally to fix #2007.


lachlan-00:
I think Ampache should stay translatable.

Add the love of music to strings update though. that's an obvious miss.

kuzi-moto:
Okay, string changes have been removed (except a couple that are new, and I don't think would merge well with stringsupdate).

> I think Ampache should stay translatable.

I have added Ampache back in the translated strings. Just looking through translations, I don't think it actually ever gets translated.

lachlan-00:
Cool. Will test tonight!

lachlan-00:
Need to test out the install and error pages now but it looks good.

Logo on the login seems a lot lower than before. There's no footer?

kuzi-moto:
Send like the logo thing is from my last pull, most be some extra padding or something. This update shouldn't have any impact on login page as it already had viewport set to device-width. I'll take a look at it though.

As for footer, it still shows up for me.

lachlan-00:
Makes sense. I'll look too

lachlan-00:
padding top
image

kuzi-moto:
Ah, nice catch!

lachlan-00:
will test installs and errors tonight but this should be fine.

PR #2026, opened 2019-08-26
Cleanup multi-line strings

kuzi-moto:
Noticed some strings that should probably be cleaned up. Not sure if I did it in the best way, feel free to fix up.


lachlan-00:
Some hints to fix but it's fine. Good spotting.

PR #1957, opened 2019-06-30
strings update

lachlan-00:
code i used to grep the translated strings
```
grep -R -oh -E "T_\(.*\)" > ../regex-strings.txt
```
[regex-strings.txt](https://github.com/ampache/ampache/files/3537658/regex-strings.txt)

@kuzi-moto This is ready for you to check case and look over any weirdness i might have made. especially around any %s / $s strings.

It would also be good if you could check these for me and see if there's a better way to do it or at least hints for them.
```
T_($access)
T_($action['title'])
T_($f_name)
T_($fsubcat)
T_($key)
T_($msgid)
T_($original, $plural, $value)
T_($pref['description'])
T_($preference['description'])
T_($preferences['title'])
T_($result['error'])
T_("%1$s %2$s (Playlist)")
T_("%1$s %2$s (Smart Playlist)")
```

Good news is this not only adds a lot of new strings but reduces the total by about 200 strings while making it more natural

PR #2028, opened 2019-08-28
review of strings part 1

kuzi-moto:
first part of my review. Lots of changes.

### DIFF
--- a/admin/access.php
+++ b/admin/access.php
@@ -49,7 +49,7 @@ switch ($_REQUEST['action']) {
         $access = new Access(Core::get_get('access_id'));
         show_confirmation(T_('Are you sure?'),
                 /* HINT: ACL Name */
-                sprintf(T_('This will permanently delete %s'), $access->name),
+                sprintf(T_('This will permanently delete "%s"'), $access->name),
                 'admin/access.php?action=delete_record&amp;access_id=' . $access->id, 1, 'delete_access');
     break;
     case 'add_host':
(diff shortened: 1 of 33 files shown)


stebe:
I'm doing a quick look at this. 


` show_confirmation(T_("There Was a Problem"), T_("Couldn't delete this Artist."), AmpConfig::get('web_path'));`

Are the sentences in the correct tense?  Using "Was" would imply that the problem is gone but really it still exists. Therefore shouldn't the sentence be "There Is a Problem"?


Same in this sentence, shouldn't this be "There is an error...".

       ` printf (T_('There was an error trying to open "%s"'), $path);`

lachlan-00:
This is after you confirm the action. So it's past in the sense of your personal confirmation

Ie
are we doing this now?

*Problem
When I did this action there was a problem

kuzi-moto:
I think it should be "was" as it's referring to something that happened in the past, since the action has already ended. Think of if it said "there was a problem with the action", it wouldn't make sense to say "there is a problem with the action" since it's not currently ongoing.

lachlan-00:
In my regex for get_cat[e|a]gories it looks like you've got all the mispelled ones. so that's good and I like all these changes. :) i'll merge this so i can use it live and have a deeper look

@stebe please feel free to go over the original pull (https://github.com/ampache/ampache/pull/1957) and leave some comments or changes you might see. we're a few weeks from merging as i'll be away next week.

PR #2035, opened 2019-09-06
Add ability for themes to use SVG icons

kuzi-moto:
I realized that Ampache has hard-coded .png files for icons and that it would be unable to load any other icon type. This modifies ui.class.php so that if a theme has an svg instead of a png that it's loaded correctly.

I opted to read the svg as an object, so that the "title" and "desc" attribute could be edited, and any other nodes can be modified/added if needed in the future. Then, the xml data for the SVG is loaded directly in the DOM so that icons will be able to be modified by external style sheets.

I don't know if my code is any good, but have a look over. I was also thinking of automatically adding the icon name to the ID attribute for additional control with CSS. It looks like you can pass the "ID" attribute as a value using the `get_icon` function, but I can't see that it's ever actually used in-code. Probably makes sense to remove that, and just do it automatically based on file name - thoughts?

Also, noticed that `$hover_name` is never used, as none of the icons are passing names as arrays... when would that be used anyway?


lachlan-00:
This is a good idea!

kuzi-moto:
Thanks! There's a couple other things I'd like to change, but it's just about done.

lachlan-00:
I think you do ifs based on svg -> fall through to png/image.

Might as well frame this to move images to svg as the primary type. (not that i even know how to make one!)

kuzi-moto:
Yeah, that sounds like a good idea. They look so much better than pngs! I'm not sure how interested we are in using other icons, but Google has released a ton of icons as open source many of which work really well.

lachlan-00:
i'm happy with anything under creative commons really.

it might also be time to make a new theme with svg icons. keep the light/dark but just a new theme name to keep the old one without replacing it entirely.

kuzi-moto:
Yeah that's kind of what I've been doing. I'm writing the CSS from scratch, finding replacement icons, etc. I think I'm going to keep the same color scheme, or at least utilizing most of the same colors, and designing it mobile-first with media queries and flexbox.

I'm not sure how long it takes, probably a long time but maybe it will be good enough for a later release. I'll probably open an issue about it for discussion later on when there's more to show so that others can throw in some input.

kuzi-moto:
@lachlan-00 I think this is ready now.

lachlan-00:
no problem. let me add in some svg's tonight and play with it.

After merges this i think i'll make a draft pull for a theme update with a new theme name

lachlan-00:
This seems like something legacy? i don't know what this sprite file is so we might be able to remove it.
```
$bUseSprite = file_exists(AmpConfig::get('prefix') . AmpConfig::get('theme_path') . '/images/icons.sprite.png');
```

lachlan-00:
i spent a lot of time on the multi disk patches last night so this is next.

I've decided that the svg theme name is savage. (SaVaGe) ;)

kuzi-moto:
I wonder if someone meant to use sprite sheets at some point and just never finished. I'm fine with just removing it entirely, we can always reimplement if we think it's useful. If we do want to keep it, then we may want to put the alt/title attribute logic in there as well.

This seems to be working really nicely though, and I'm excited to mess with it more. It will give a lot more control over icon appearance as you can style pretty much everything in the SVG with CSS.

I was going to go with "reborn-responsive", but "Savage" sounds way better haha.

lachlan-00:
sprites look like part of a commit from about 4 years ago when the theme used them.
https://github.com/ampache/ampache/commit/25f10ee47ee31c399221117c9c74ca6659479321

I'll remove that and test asap today

kuzi-moto:
Great! I think this will be a nice change for the interface.

lachlan-00:
Almost done.
Just want to test fall through to /images folder some more and play around with the find part a bit more.

I'm thinking maybe move the PNG's to /images keeping the unique ones in reborn.

That way a theme doesn't need to have all the icons by default

kuzi-moto:
I added the lines back in. Plus a lot of other fixes. I noticed that there were a lot of images linking directly to the fallback location. Added a new function to take care of those, plus added ability to add a class since it's required for the expand/collapse buttons.

Themes can now change more icons!

lachlan-00:
I think the install templates need the direct link because the settings required from ampconfig::get don't exist when you use a subfolder in the webpath. (e.g. http://ampache/music will miss the music part until you've set it in the installer)

but that's easy enough to test. I can tell you're excited for this one to get in!

kuzi-moto:
Ah, that would make sense! Any reason that Ampache doesn't just use relative links?

Yeah, I put work on my theme on hold since I didn't want to make too many huge changes to Ampache and wanted to incorporate them into develop first.

lachlan-00:
Once this goes in i'll make a pull for you and mattish to play in with a theme folder to work on.

i'm really keen to make this easy for people to pick up and this work makes it easier to handle. this work should make multiple themes much more viable.

kuzi-moto:
Nice! Yeah I think anything that makes it easier to work with Ampache will only be good, and hopefully spur more people to work with it.

lachlan-00:
so sweet!
image

lachlan-00:
this should be finalised tonight when i get home.

PR #2045, opened 2019-09-14
remove inline style

kuzi-moto:
Noticed a spot where style is hardcorded inline. I don't think this a is a good idea as it makes Ampache less flexible. There are probably a bunch more, but this was first one I found.


lachlan-00:
chuck this into strings update and we'll do it from there seeing as it's almost done.

kuzi-moto:
Sounds good! Maybe I'll try to find a few more before I submit a pull there.

lachlan-00:
merged in stringsupdate

PR #2048, opened 2019-09-16
Sidebar tweak and translation add

kuzi-moto:
Feel free to edit as you think is necessary, but I wanted to replace the '<<<' & '>>>' for the sidebar with an icon so I moved it into the css, and since there is no icon added some logic to detect if theme doesn't provide an icon to fallback to the brackets.

Also, noticed that "Loading" wasn't being translated.

### DIFF
--- a/templates/header.inc.php
+++ b/templates/header.inc.php
@@ -478,7 +478,8 @@ $t_logout    = T_('Log out');
                 <?php
                 // remove when reborn has an icon for this
                 $icon_expand = UI::get_icon('expand', T_('Expand/Collapse'));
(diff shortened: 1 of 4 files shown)


lachlan-00:
will need to tweak a bit, this will probably have to be light/dark specific but even just moving to css will be better. i'll make an icon up of the current interface as a png

kuzi-moto:
Makes sense, I just wanted to get things working for myself at least, and put it on your radar. There are a lot of other styles hard-coded in Ampache that I'd like to move to the theme so that everything is customizable.

lachlan-00:
i don't like my icon, i'll just move it to light/dark for now.

PR #2050, opened 2019-09-19
JSON API 2019

lachlan-00:
updated branch from #1268 for current develop


AshotN:
Glad to see you still working on this. What's preventing the JSON API from being merged?

lachlan-00:
Mostly just that the new XML API hasn't stabilised completely and now that subsonic has an xml2json function I'm considering how it might be easier to do the same.

But being able to request json is important and I'll be merging it at some point

AshotN:
Do we have a page with all available API calls and future planned ones? It's kind of a mess with how many different services are being supported right now. The original reason I worked on the JSON API was to build a music client, I think that gave me a lot of insight as to the faults of the current API. I have forgotten most of the issues and experiences so maybe some organization is what we need to cleanup the Ampache API.

lachlan-00:
The wiki has pages for the methods but there is still work to do because I'm not actually sure what some of the inputs actually do yet.

I've been working on a python library for the API and if you'd like I can tag you in a new issue when I'm 90% sure I won't be adding new methods?

I'll be writing up a decent markdown doc with details when I'm finalising it as well

What's been good about writing a python lib has really helped me work out what's missing and I've already added a few things I thought were missing

AshotN:
Ok i see you created a list of the XML API https://github.com/ampache/ampache/wiki/XML-methods, which if I recall should have feature parity with the JSON API. I'll try to get reacquainted with the XML and JSON files and see if I can reduce redundancy. I want us to be able to expand the API without it being a huge headache of managing feature parity across a million different services.

I would also like to have a cleaner documentation, but I have to admit that API documentations is something I have always kinda slacked off on and been meaning to figure out how to do cleanly.

lachlan-00:
@AshotN i've just done a big update on api.class to use switches instead of format and based it on the current api in develop. There is probably a lot not working in json_data but i'm going to start working towards this pull asap. (after 4.1.0 though)

lachlan-00:
@AshotN i'm going to merge this when i can test the function of xml. (maybe sunday night)

json_data will need to be updated as wel go but i don't see why we can't merge this once i've verified xml

lachlan-00:
Tested and happy that it hasn't broken anything obvious in the API for XML.

after travis this is in. @AshotN you've probably made the biggest feature addition here in years!

PR #2059, opened 2019-09-30
Update issue templates

kuzi-moto:
Github has a new method of storing issue templates, so I changed the existing one, and added one for feature requests. These show up in the GUI as options to choose from, so it makes it look a bit nicer and its easier to use. Also, I re-ordered the sections, as I thought it would be nicer if the issue was at the very top, with all the supporting details underneath, so you don't have to scroll so much in some cases with large logs/configs before getting to the issue itself.

image


lachlan-00:
I don't know which branch this needs to be in but develop is probably right as the main branch.

PR #2085, opened 2019-10-14
remove old html from links

lachlan-00:
rel="nohtml" has no use/value in the code


lachlan-00:
this actually seems to be needed in areas

PR #2088, opened 2019-10-14
Various HTML fixes

kuzi-moto:
* Removed `language="javascript"` from `<script>` tag as it's not needed.
* Fix to ui.class where icons/images are given an id attribute based on their name. This would cause the same id to appear multiple times on a page. Now it is only assigned an id if manually specified.
* Removed tags from show_web_player_embedded, putting duplicate `<html>`, `<body>`, `<title>` tags in the markup.
* other small tweaks to make code easier to read.
* Fix lost password button on login page
* Fix incorrect id attribute in sidebar_home

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -277,7 +277,7 @@ abstract class Catalog extends database_object
      */
     public static function show_catalog_types($divback = 'catalog_type_fields')
     {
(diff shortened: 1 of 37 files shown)


lachlan-00:
i'll do the php-cs after the merge

kuzi-moto:
Oops, thanks!

PR #2097, opened 2019-10-16
Change rel="nohtml" to a class

kuzi-moto:
Resubmitted because somehow I had some extra files. But seems to be working!


lachlan-00:
Seems like a much cleaner way of doing it where rel seems to be outside of our actual need for using it

PR #2108, opened 2019-10-23
update from tag results as a table

kuzi-moto:
Noticed that the results page style after updating from tags doesn't match the rest of Ampache. Making it into a table seems to look much better.

Before:

image

After:

image


lachlan-00:
Much cleaner!

I'll merge this after I fox the upload problem in edge

PR #2135, opened 2019-11-07
Lots of HTML/CSS fixes

kuzi-moto:
Included are a lot of things I've noticed while creating a responsive theme and wanted to merge some of these changes before there are too many. A list of what has been done:

* Replaced direct paths to images with get_image/get_icon
* Changed #ajax-loading to remove inline styles. Now it adds/removes a `hidden` class instead.
* Removed some attributes from HTML that are no longer needed
* Added missing translations to cookie disclaimer page
* Removed `<html>`, `<title>`, `<body>` HTML tags that were being embedded within Ampache pages causing duplicates
* Moved theme stylesheet to bottom to take preference over other stylesheets.
* Changed header image to remove links to images from within CSS, and put the image into actual `` tags instead of on a header to produce better markup.
* Changed sidebar JS to put the `expanded` and `collapsed` classes onto the entire `<li>` tag instead of individual items to make things more simple.

The only thing left that I couldn't quite figure out was that expanding/collapsing the sidebar no longer animates as that was handled by javascript. CSS transitions require changing from absolute values.


wagnered:
Kuzma,

Have you checked out w3.css on the w3schools web site? It's a responsive css that doesn't require jquery or javascript.  I used it on the Laravel branch to create the exanding side panel with the dropdown lists. It's very easy to modify the direction and type of expansion/collapse.

It can be combined with the unique styles in Ampache's Css3 version, but Ampache's CSS seems "patchy" and unorganized and often conflicted with the in line w3.CSS I wanted to use.

-------- Original message --------
Subject: [ampache/ampache] Lots of HTML/CSS fixes (#2135)

lachlan-00:
So glad you've been finding all these issues. It's going to make theming a lot easier!

So the only issue left with this is the sidebar? I think the js for it looks for a specific class/id. I've definitely broken it before

kuzi-moto:
@wagnered Yes, I have seen the w3.css framework, along with several others. I was thinking about using them at some point, but I know that we already ship Ampache with bootstrap, and would hate to add another framework. I'd don't mind writing the CSS by hand, it's mostly that Ampache's source is messy, and often relies on ids, when we should be utilizing classes more.

> So the only issue left with this is the sidebar? I think the js for it looks for a specific class/id. I've definitely broken it before

I removed the JS code used for that, since I don't believe that javascript should be used for anything directly affecting the presentation of a web page, so that it is customizable by themes. It should be possible with CSS, just needs a bit of tweaking. Thinking about it, I should probably change the class for `#ajax-loading` to be active/inactive instead of hidden.

Also, it looks like it failed the CI check, but I ran php-cs-fixer on it.

kuzi-moto:
> So glad you've been finding all these issues. It's going to make theming a lot easier!

Thanks! And yeah, that's the plan. I think before making a new theme, I want to fix the source code up a bit. I want to separate all the PHP and HTML code, and then standardize all the classes, and add new ones so that it makes writing CSS for Ampache easier. A lot of the classes used seem to be inconsistent. It would be a lot better than trying to mess with the current structure I think.

lachlan-00:
I think the Travis conf needs to bump up now that I've added a higher version of php-cs-fixer. I'll do that when I get to work

kuzi-moto:
Oh, looks like somehow I didn't add my changes to the commit.

templates/error_page.inc.php
```diff
@@ -41,7 +41,7 @@
     <div class="navbar navbar-inverse navbar-fixed-top" role="navigation">
         <div class="container">
             <a class="navbar-brand" href="#">
-                <img src="./images/ampache-dark.png" title="<?php echo $t_ampache; ?>" alt="<?php echo $t_ampache; ?>">
```
lachlan-00: the error page needs to be hardcoded for install/test errors

templates/install_header.inc.php
```diff
@@ -48,7 +48,7 @@
     <div class="navbar navbar-inverse navbar-fixed-top" role="navigation">
         <div class="container">
             <a class="navbar-brand" href="#">
-                <img src="./images/ampache-dark.png" title="<?php echo T_('Ampache'); ?>" alt="<?php echo T_('Ampache'); ?>">
```
lachlan-00: same for this one

templates/show_install_lang.inc.php
```diff
@@ -23,7 +23,7 @@
 require $prefix . '/templates/install_header.inc.php'; ?>
         <!-- Main jumbotron for a primary marketing message or call to action -->
         <div class="jumbotron">
-            <h1 id="headerlogo"><img src="./images/ampache.png" title="<?php echo T_('Ampache'); ?>" alt="<?php echo T_('Ampache'); ?>"></h1>
```
lachlan-00: and this

templates/show_test_config.inc.php
```diff
@@ -34,7 +34,7 @@
         <div class="navbar navbar-inverse" role="navigation">
             <div class="container">
                 <a class="navbar-brand" href="#">
-                    <img src="./images/ampache-dark.png" title="<?php echo T_('Ampache'); ?>" alt="<?php echo T_('Ampache'); ?>">
```
lachlan-00: ;)

lachlan-00:
Besides the install/test/error pages that need to keep the hardcoded images. (they use the dark reborn theme) the sidebar will need to work either the old way or a new way. but other than that these are all good changes.

Travis is bumped now as well

AshotN:
Ampache does use a lot of IDs, really ID shouldn't be used that often. I also think a lot of CSS declarations need to be removed. Especially something like vendor prefixes for border-radius, something that has been supported without vendor prefixes for over about years.

lib/javascript/ajax.js
```diff
@@ -17,18 +17,18 @@
 
 // Some cutesy flashing thing while we run
 $(document).ajaxSend(function () {
-    $('#ajax-loading').show();
+    $('#ajax-loading').removeClass("hidden");
 });
 $(document).ajaxComplete(function () {
-    $('#ajax-loading').hide();
+    $('#ajax-loading').addClass("hidden");
```
AshotN: Any reason to do this? .hide() and .show() accomplish the same exact thing without needing a CSS deceleration.

lib/javascript/jplayer.playlist.ext.js
```diff
@@ -12,8 +12,8 @@ jPlayerPlaylist.prototype._createListItem = function(media) {
         }
         attrClass += "\" ";
 
-        var listItem = "<li" + attrClass + " name=\"" + $(jplaylist.cssSelector.playlist + " ul li").length + "\" " +
-                "data-poster=\"" + media.poster + "\" data-media_id=\"" + media.media_id + "\" data-album_id=\"" + media.album_id + "\" data-artist_id=\"" + media.artist_id + "\" " +
+        var listItem = "<li" + attrClass +
```
AshotN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals would make this readable. But I don't know if Ampache cares about Internet Explorer support.

lib/javascript/ajax.js
kuzi-moto: My idea was that all aspects of the web interface could be controlled through the theme. Styles applied with JavaScript are done directly to the element, making it hard to override, and change the behavior.

Then, if there are multiple elements that are being shown/hidden you can also apply the same styles to them, without adding additional selectors to a CSS rule.

Also, thinking about it, it might make more sense to toggle an active/inactive class instead.

templates/error_page.inc.php
kuzi-moto: Sounds good, will fix

lib/javascript/jplayer.playlist.ext.js
kuzi-moto: I'm not sure about this, would be nice to know if anyone still uses it.

kuzi-moto:
Okay, so everything should be fixed up now, and works the same way it did before. Let me know if there's any other changes you'd like to see.

lachlan-00:
Looks fine now.
Let's merge into edge and give it a whirl

PR #2144, opened 2019-11-09
Add random_albums API method

AshotN:
Returns random albums from the user.

### DIFF
--- a/lib/class/api.class.php
+++ b/lib/class/api.class.php
@@ -669,6 +669,28 @@ class Api
         echo XML_Data::albums(array($uid), $input['include'], true, $user->id);
     } // album
 
+        /**
+     * MINIMUM_API_VERSION=400001
+     *
+     * This returns a number of random albums.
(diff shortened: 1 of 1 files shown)


lachlan-00:
stats already gets random albums?

lachlan-00:
in master the api only searched for songs but in develop i put in album and artist
image

AshotN:
I did not notice stats, I guess the name of it really didn't give me the impression. I should've searched for ::get_random 

I guess we can close this out

lachlan-00:
We could alias it out but i did the same thing before i realised stats already did it for songs

AshotN:
I think an alias would be nice, wouldn't really require much code. 

I changed my code to `stats` worked exactly the same, cool. 
image

lachlan-00:
if something like that caught you up i think it's a good idea. being able to think of the action you want and actually find it would be beneficial for others.

PR #2148, opened 2019-11-09
General translation Update

Psy-Virus:
Updated translations from Transifex

:)


Psy-Virus:
Whoops, I guess I was too fast for Travis. Anyways Translation Updates don't tinker with PHP-Code, so it's safe, even if he failed.

PR #2149, opened 2019-11-10
Update api.class.php

4phun:
Hi,

With reference to our conversation earlier, here's my humble contribution to the XML API. Please check it and let me know what do you think.

Added new function ad_hoc_playlist.

### DIFF
--- a/lib/class/api.class.php
+++ b/lib/class/api.class.php
@@ -1188,7 +1188,7 @@ class Api
             $rule_count++;
         }
         if (array_key_exists("album", $input)) {
-            $album = New Album((int) $input['album']);
+            $album = new Album((int) $input['album']);
             // set rule
             $array['rule_' . $rule_count]               = 'album';
             $array['rule_' . $rule_count . '_input']    = $album;
@@ -1196,7 +1196,7 @@ class Api
             $rule_count++;
         }
         if (array_key_exists("artist", $input)) {
-            $artist = New Artist((int) $input['artist']);
+            $artist = new Artist((int) $input['artist']);
             // set rule
             $array['rule_' . $rule_count]               = 'artist';
             $array['rule_' . $rule_count . '_input']    = $artist->id;
--- a/lib/class/api.class.php
+++ b/lib/class/api.class.php
@@ -1205,13 +1205,17 @@ class Api
         }
 
         ob_end_clean();
-    
         XML_Data::set_offset($input['offset']);
         XML_Data::set_limit($input['limit']);
 
         // get db data
         $song_ids = Search::run($array, $user);
         shuffle($song_ids);
+
+        //slice the array if there is a limit
+        if ((int) $input['limit'] > 0) {
+            $song_ids = array_slice($song_ids, 0, (int) $input['limit']);
+        }
         
         // output formatted XML
         switch ($format) {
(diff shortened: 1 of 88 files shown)


wagnered:
Not pointing fingers, but I've seen several PR's that hadn't had php-cs-fixer
run before commiting changes and pushing to remote.

I've done it too.   Can't say how many times I had to revert local commits, run
php-cs-fixer and recommit.  The positive side is that only one commit is then needed.

4phun:
@wagnered True that! Thank you for the heads up.

Pull request have been updated, tests are passing now.

lachlan-00:
There's a little bit of overlap in this that i think is partially solved by advanced_search, stats and playlist methods (when you make a smart playlist.)

object_count is also probably the best place to check activity. user_activity in develop is a lot different compared to stable and it was also not previously mandatory.

Let me have a play with this one but i think we should also call it playlist_generate

lib/class/api.class.php
```diff
@@ -2357,4 +2357,107 @@ public static function democratic($input)
...
+     *
+     * @param array $input
+     * $input = array(mode            = (string)  "least_recent_played"|"most_recent_played"|"random"   // optional, default = "least_recent_played"
+     *                filter          = (string)  $filter                                               // optional, LIKE matched to song title
+     *                album_id        = (integer) $album_id                                             // optional
+     *                artist_id       = (integer) $artist_id                                            // optional
+     *                flagged_only    = (integer) 0|1                                                   // optional, default = 0
+     *                format          = (string)  "song"|"index"|"id"                                   // optional, default = "song"
+     */
+    public static function ad_hoc_playlist($input)
```
lachlan-00: public static function playlist_generate

lachlan-00:
Actually stats does almost all of it but i think we can tweak this to include some advanced searching as well.
image

lib/class/api.class.php
```diff
@@ -2357,4 +2357,107 @@ public static function democratic($input)
...
+        
+        // process parameters
+
+        $select       = array("song.id AS song_id");
+        $from         = array("song");
+        $where        = array();
+        $order        = array();
+        $bound_values = array();
+
+        if (in_array($input[mode], array("least_recent_played", "most_recent_played"), true)) {
```
lachlan-00: $input[mode] should be ($input['mode']

The input array is like a dictionary

4phun:
As we agreed earlier, I only meant this to be a conversation stater.

stat has "recent", but not its opposite "least_recent" a.k.a. "oldest". Not sure what would be the best name for it.

4phun:
Cool. As we agreed earlier, I only meant this to be a conversation stater.

> Actually stats does almost all of it but i think we can tweak this to include some advanced searching as well.

stat has "recent", but not its opposite "least_recent" a.k.a. "oldest". Not sure what would be the best name for it. If you think it's better to add that to stat, that works for me just fine.

lib/class/api.class.php
4phun: Ah, hard to switch my brain off of perl... But of course you're right.

lachlan-00:
like most things it's close but not terribly close. i'll do an edit on this pull and extend out stats as well. that way you get contributor status in the repo! ;)

I think it's good to keep this as sort of an alias method as i'll be using advanced search and stats to generate the xml.

Give a day or so to work on it and we'll put it in asap.

lachlan-00:
I like to think if it's not 'recent' it's 'forgotten'. i'll ping this pull again when i'm done.

4phun:
Great, thank you very much.

lachlan-00:
I'm going to try and do this one tonight to finish off the api changes.

4phun:
Great news, looking forward to see it in Ampache!

lachlan-00:
just working in edge on this now.

i've set it to use searches instead of sql and i just need to test all the options but it's working.

lachlan-00:
I've ruined this tree but will merge this now from https://github.com/ampache/ampache/pull/2183

in your local git do this to restore it after i've merge the new pull
```
git pull
git reset --hard upstream/edge
git push --force origin edge
```

4phun:
Cool! Thank you!

Issue #2147, opened 2019-11-09
[Feature Request] implement album_songs into album


AshotN:
The API right now has `albums` and `album_songs`. Instead of needed two API calls, could we add a boolean flag to `albums` to include songs in the output. I don't think this should require two API calls.

lachlan-00:
A lot of the other calls have include so we could add include=songs to do just that.

AshotN:
I'd implement it, but it's kind of a mess for me to create a pull for develop because my code is mixed with the JSON API in it. On my client I'm making two database calls, which is kind of annoying but I just wanna get a working prototype to show case here quickly.

lachlan-00:
I can get this done sometime quickly tonight/tomorrow. Its a pretty reasonable change to make for artists, albums, etc

lachlan-00:
It actually looks like it's already meant to do it

album and albums both have an include filter so i'll double check why it doesn't return anything extra
```
echo XML_Data::albums($albums, $input['include'], true, $user->id);
```

lachlan-00:
It is looking for an array but not actually making one. sending in a fix now.

Album, albums will return songs with include=songs

PR #2150, opened 2019-11-10
fix for #2147

lachlan-00:


lib/class/xml_data.class.php
```diff
@@ -489,9 +489,12 @@ public static function tags($tags)
      */
     public static function artists($artists, $include = [], $full_xml = true, $user_id = false)
     {
-        if (null == $include) {
+        if ($include == null || $include == '') {
             $include = array();
         }
+        if (is_string($include)) {
```
AshotN, line 495: `$include` is expected to be of type `array` as per the docs. Why hare we allowing someone to send in a `string`?

lachlan-00: i'll check where this is called from, back here was just breakfixing so it's probably simple to clean up

lachlan-00: explode on the input?
```
echo XML_Data::artists($artists, $input['include'], $user->id);
```

AshotN: I think that would make more sense, would get rid of a check. Reduce complexity.

lachlan-00: yeah, now that i've documented all these functions they need to actually send what we want.
doing it in #2347 as well

Issue #1515, opened 2017-03-29
c-pchart has a licensing clause that makes it potentially non-open source


bowlofeggs:
Hello!

I've been working on [packaging](https://fedoraproject.org/wiki/Ampache) Ampache for the [Fedora project](https://getfedora.org/) for some time. Fedora has a peer review process for adding packages to the distribution, and during [package review](https://bugzilla.redhat.com/show_bug.cgi?id=1425275) one of my peers noticed an odd clause in the readme file for c-pchart that mentioned that c-pchart can only be used under the GPL [if it is not being used in a commercial project](https://github.com/szymach/c-pchart/blob/2.0.4/README.md#license). This raised a red flag and so we consulted Fedora's legal list to ask whether this would be a problem:

https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/4MDAWLRSVR3QIIYFJLVKMEELS5OPBIMI/

The legal list ruled that c-pchart was not suitable for Fedora due to the remarks about making money and commercial use, as the GPL doesn't make any such restrictions, and adding restrictions like that does mean the software isn't truly Free Software. One of the comments was from [Richard Fontana](https://en.wikipedia.org/wiki/Richard_Fontana), a notable authority on the GPL.

I've spent the last couple of weeks trying to contact the author of the upstream to c-pchart to ask him to clarify his [licensing statements](http://www.pchart.net/license), but have been unable to get in touch. I sent him an e-mail and a message on Linked In, but he has not responded after about 2.5 weeks. I then contacted the c-pchart author about this issue to ask if he knew how to get in touch, but unfortunately he did not:

https://github.com/szymach/c-pchart/issues/35

Note that you can see the e-mail I sent to the upstream pchart author in that issue, that explains the issue a bit more than I've done here.

I would like to request that the Ampache project consider taking some action to remedy this situation, as I will not be able to add Ampache in its current form to Fedora. Here are some options I can think of:

1. Ampache could also attempt to contact the pchart author to ask for license clarification. It might help to get some more voices.
2. Ampache could attempt to replace c-pchart with a different library.
3. Ampache could make c-pchart optional through configuration, and Fedora could simply configure a boolean to be False. This assumes c-pchart isn't crucial to Ampache's core functionality.
4. Also depending on how crucial c-pchart is, Fedora could carry a patched version of Ampache that removes functionality that depends on c-pchart, and distribute the rest of Ampache.

What do you think? Please let me know if you can think of other options.

[...]

PR #2175, opened 2019-11-18
p-chart is released under gplv3

lachlan-00:
restore it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -42,7 +42,7 @@ Steps to reproduce the behavior:
 
 <!-- Keep only the affected clients in the list below. Add the client application name and version to the right of the client type used. -->
 
-### Client type:
+### Client type
 
 * Web
 * Ampache
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
         "tightenco/collect": "5.2.*",
         "ampache/ampacheapi-php": "^1.0",
         "dropbox/dropbox-sdk": "^1.1",
-        "szymach/c-pchart": "^3.0"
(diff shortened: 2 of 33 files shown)


wagnered:
@lachlan,

PChart is dual licensed; ampache.org cannot change that.

I don't remember the exact details, maybe you do, but someone wanted to create a package to include with one of the Linux distributions.  He was refused because of this dual license situation.  I think it might have been Fedora.

Until pChart is replaced by chart.js or other open-source charting library, I would like to suggest that pChart be  enabled on the admin tab with the user being required to acknowledge that for commercial use, a commercial license must be purchased, and point to http://www.pchart.net/license for details and to purchase a commercial license.

I suggest this to reduce the potential for ampache.org becoming involved in legal issues.



-------- Original message --------
Subject: Re: [ampache/ampache] p-chart is released under gplv3 (#2175)

lachlan-00:
It is definitely dual licensed still but has changed from MIT to gplv3 since that issue.

Graphs are disabled by default requiring config to enable so it could go in there.

I need to do more research in the reasoning behind the fedora packaging with dual licensing as I don't think a lot of the restrictions apply to Ampache itself.

wagnered:
The [Fedora docs](https://docs.fedoraproject.org/en-US/quick-docs/setup_rpmfusion/) might shed some light on that.

I think someone at pChart finally decided to look at the wording of the of the MIT is excludes any attempt to impose copyright or proprietary clauses to the conditions of usage.

lachlan-00:
I've reopend the issue #1515 and put it to debian maintainers of the old versions as well.

I think to avoid all this i'll separate c-pchart package from the base composer and detail the extra requirements to enable in the config and wiki. that way it's not included in the base install without multiple manual steps.

PR #2178, opened 2019-11-19
smarter tag search using a list / boolean_numeric

lachlan-00:


kuzi-moto:
Pulling an all-nighter @lachlan-00?

lachlan-00:
It's been good to put the extra time in this week with so many little fixes.

Although that was a particularly long session, that was more of an all dayer too

PR #2185, opened 2019-11-21
duplicate check for Api::playlist_add_song, playlist->has_access()

lachlan-00:


lachlan-00:
@AshotN there is now a check item for adding to playlists.
if you set check=1 then duplicates won't be added to the playlist

PR #2192, opened 2019-11-22
Html css fixes

kuzi-moto:
I have added a fix for the loading div, it was active all the time.

@lachlan-00, I have tested in Firefox and Chrome on Windows, and Safari on Mac. I could not identify any of the issues you were describing previously.


lachlan-00:
Let's do it again then!

lachlan-00:
Much better. Only issue left is the logo in mobile and the blue pointer arrow in the search header is the opposite of all the others.

I should be able to fix those easy.

Is the blue pointer arrow in the titles the reverse of the original way?

kuzi-moto:
There's an issue with the JavaScript I think. Arrows should be working correctly, it's that the sections are expanding when the class says collapsed. Also, the cookie doesn't seem to be saving properly, so they are always expanding on refresh (but technically they should be collapsing, not sure why they aren't). Looks like I just made a big mess of this thing.

lachlan-00:
Okay, well let's keep it in edge and we'll try to fix it all up this weekend

PR #2228, opened 2019-12-03
playlist_create suggested changes

manuelglez86:
This fix th ID not showing on the XML file when you create a new playlist with an API Client


manuelglez86:
i had a type error, let me correct

manuelglez86:
You already merged, on the xml data file on the if i put $created = true it should be $created == true thats right?

lachlan-00:
Don't worry about it, I'll be testing now

Issue #2140, opened 2019-11-08
Selecting `Add New...` does not prompt for Artist or Album on save


NatureMuncher:
## Description
Many of the songs in my collection are unknown so are stuck with the `Unknown (Orphaned)` labels for Artist and Album name. I am unable to update them.
### Describe the bug
I select Edit and then select the drop-down for `Add new...`. I click Save hoping for a prompt or something but nothing happens. The Artist and Album are simply blanked out.

### To reproduce
1. Open a song in the Unknown album and click Edit
2. Attempt to change Artist or Album details from Unknown to Add New... 
3. Click Save

### Expected behavior
Guessing that when `Add New...` is selected that either a popup would prompt for the new Artist/Album name, or when clicking `Save` another screen would prompt me to enter relevant details

## Environment

* Ampache version: Develop
* Web server + version: Apache/2.4.6
* Server operating system: CentOS Linux release 7.7.1908
* Client operating system: Fedora release 31 (ThirtyOne)

### Client type:

* Web

## Settings

https://pastebin.com/SvT80nzm

lachlan-00:
I can safely say I've never actually used this feature so there probably is an issue here. Will need to test

lachlan-00:
I didn't realise this was even a feature but will need to work out what it's supposed to be doing.

NatureMuncher:
How else would I set the artist/album correctly? Rename the file perhaps?

kuzi-moto:
> How else would I set the artist/album correctly? Rename the file perhaps?

I believe the best way would be to edit the ID3 tags on your music. There are many programs to do this, but Musicbrainz Picard is a good one that can auto-tag. My personal favorite is beets, which can be fast depending on how you use it.

lachlan-00:
What's annoying is that I have been through pretty much every line of Ampache and I don't think there is any facility of adding an artist outside of tags.

So why it's even there I'm not sure but we should also be providing this function as well

manuelglez86:
Yes it is in previous versions you can Add New artist (or Album) manuelly from the web interface, is usefull when you only need to edit 1 or 2 songs so you dont have to Verify the entire catalog, and I think is realated to #2212

lachlan-00:
it is related. i'll need to install the older versions and have a look as i've never used it before.

lachlan-00:
I just installed legacy-develop and it doesn't show up in the lists. is there an option for it?

lachlan-00:
[...]

PR #2241, opened 2019-12-13
Update tools.js - fix issue #2140

Rhythm-of-Time:
This will make the Add New... artist in Song editor work again, like described in issue #2140. 

Instead of changing "===" to "==" to avoid strict comparison and allow for automatic type casting, one could also implicitly declare the type by using double quotes and leave the strict comparison in place: === "-1". Maybe this is cleaner?

I noticed that this version of tools.js is using double instead of single quotes (unlike the old one). This may come handy when using JSON, but adding HTML becomes more difficult, because all double quotes contained in strings need to be escaped - see line 339. Hard to read...


Rhythm-of-Time:
For PHP 7.4, Travis CI build failed with the following error:

PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in ./modules/oauth/OAuth.php on line 126

I'll fix at the file and make another pull request...

Rhythm-of-Time:
Frankly, I have no idea why the tests keep failing. May this be related with the versions of php-cs-fixer.phar i used? First I tried the one from the download link in the Ampache Wiki, which complained about my local PHP version being above 7.2 (actually 7.3.11). Then I used v2.16.1, which did change the line endings of all files to LF. Any suggestions what else could be the problem with Formatting?

lachlan-00:
Thanks! Will test out and merge ASAP

lachlan-00:
thanks for becoming a contributor to ampache!

Rhythm-of-Time:
Thank you for fixing my github mess...

PR #2244, opened 2019-12-19
Extend information about local_web_path for reverse proxies.

cquike:
This PR simply adds some extra information to parameter local_web_path in the default configuration file which I would have found useful when setting up ampache behind a reverse proxy.


lachlan-00:
Thanks for adding. When I get home on Monday I'll add this and then update it in the wiki as well

Issue #2141, opened 2019-11-08
Unable to update album info


NatureMuncher:
## Description
GUI appears to accept input but changes do not show up upon refresh.

### Describe the bug
Changing Album artist to correct value doesn't seem to stick
### To reproduce

1. Go to Album
2. Click on Edit
3. Scroll down to Artist
4. Click incorrect value
5. Scroll to and click on correct value
6. Click Save
7. Go back to album description and see old Artist value still exists

### Expected behavior
The value should update to the new one

## Screenshot
![image](https://user-images.githubusercontent.com/1648745/68505545-1954e780-022d-11ea-9c19-ddacc0e8c5ef.png)

## Environment
* Ampache version: Develop
* Web server + version: Apache/2.4.6
* Server operating system: CentOS Linux release 7.7.1908
* Client operating system: Fedora release 31 (ThirtyOne)

### Client type:
* Web

## Settings
https://pastebin.com/SvT80nzm

## Logs
```
2019-11-08 13:30:51 [ampache] (session.class) -> Existing session found 
2019-11-08 13:30:51 [mike] (session.class) -> c2145892c23b1f05bf448c283ff27140 has been extended to Fri, 08 Nov 2019 14:30:51 -0600 extension leng
th 3600 
2019-11-08 13:30:51 [mike] (edit.server) -> Called for action: {edit_object} 
2019-11-08 13:30:51 [mike] (tag.class) -> Updating tags for values {Other} type {album} object_id {860} 
2019-11-08 13:30:51 [ampache] (session.class) -> Existing session found 
2019-11-08 13:30:51 [mike] (session.class) -> c2145892c23b1f05bf448c283ff27140 has been extended to Fri, 08 Nov 2019 14:30:51 -0600 extension leng
th 3600 
2019-11-08 13:30:51 [mike] (session.class) -> Session created: 53ced1c92250377107327a5baa2de508 
2019-11-08 13:30:51 [mike] (browse.class) -> Show objects called for type {song} 
2019-11-08 13:30:51 [ampache] (session.class) -> Existing session found 
2019-11-08 13:30:51 [mike] (session.class) -> c2145892c23b1f05bf448c283ff27140 has been extended to Fri, 08 Nov 2019 14:30:51 -0600 extension leng
th 3600 
2019-11-08 13:30:51 [mike] (ajax.server) -> Called for page: {browse} 
2019-11-08 13:30:51 [mike] (browse.ajax) -> Called for action: {get_filters} 
[...]

PR #2249, opened 2019-12-21
 Fix: Unable to update album info #2141 

Rhythm-of-Time:
New Albums / Artists can be created and existing artists or album can be assigned to songs.

Some fine-tuning to be done - a lot of debug messages still there, and maybe the Song Edit Dialog needs a little work if the meta tags for Artist and Song ID should be updated by AJAX to reflect changes made in the dropdown.

Sorry for the large quantity of commits - I am still practising using Github. Only tools.js and song.class.php are needed for the fix - that is, the last two commits.


lachlan-00:
I can squash and play around with the commits so don't work too much about that.

An easy fix is to copy the changes out of your repo and do a reset.
Then you copy the changes files back in after for a single commit

git reset --hard upstream/develop
git push origin develop --force

These commands may be wrong. I'm not home for another day

Rhythm-of-Time:
Thanks for the advice, using the command line is much easier than using Github Dektop (on Windows) which also creates troubles of it own sometimes. Didn't dare to use reset --hard, now I know how to deal with the files. Backing up outside of a repo is always a good thing...

lachlan-00:
okay, all tested now. see what happens when you contribute? you get the bug for it.

lachlan-00:
The commands above should work to reset your local repo to the correct commits. just make sure you pull first
```
// make sure the ampache upstream url is available
git remote set-url upstream https://github.com/ampache/ampache.git
git pull
```
```
// This will reset to Ampache\develop
git reset --hard upstream/develop
```
```
// This will overwrite Rhythm-of-Time:origin/develop to develop based on your reset
git push origin develop --force
```

Rhythm-of-Time:
Thanks. Worked like charm.

PR #2265, opened 2020-01-16
Fixed GROUP BY error in artist.class get_albums

KingOperator:
Hi @lachlan-00,

when using the current develop version I experienced an empty album list on the artist page. In the debug log the following error occurs:

` 2020-01-16 21:36:40 [xyz] (dba.class) -> Error_query SQL: SELECT MAX(``album``.``id``) AS ``id``, ``album``.``release_type``, ``album``.``mbid`` FROM ``album`` LEFT JOIN ``song`` ON ``song``.``album``=``album``.``id`` LEFT JOIN ``catalog`` ON ``catalog``.``id`` = ``song``.``catalog`` WHERE (``song``.``artist``='1234' OR ``album``.``album_artist``='1234')  GROUP BY ``album``.``prefix``, ``album``.``name``, ``album``.``album_artist``, ``album``.``release_type``, ``album``.``mbid``, ``album``.``year``  ORDER BY ``album``.``name``, ``album``.``disk``, ``album``.``year``
2020-01-16 21:36:40 [xyz] (dba.class) -> Error_query MSG: ["42000",1055,"Expression #2 of ORDER BY clause is not in GROUP BY clause and contains nonaggregated column 'ampache.album.disk' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by"] `

This error is independent of the used MySQL version (7.x or 8) and depends on the used sql_mode=only_full_group_by.

Please validate the provided fix and merge if it is ok for you.

Br

### DIFF
--- a/lib/class/artist.class.php
+++ b/lib/class/artist.class.php
@@ -307,7 +307,8 @@ class Artist extends database_object implements library_item
                 $sql_sort = '`album`.`name` DESC' . $sort_disk;
                 break;
             default:
-                $sql_sort  = '`album`.`name`, `album`.`disk`, `album`.`year`';
+		$sql_sort  = '`album`.`name`, `album`.`disk`, `album`.`year`';
+		$sort_disk = ', `album`.`disk`';
         }
 
         $sql = "SELECT `album`.`id`, `album`.`release_type`, `album`.`mbid` FROM `album` LEFT JOIN `song` ON `song`.`album`=`album`.`id` $catalog_join " .
@@ -315,7 +316,7 @@ class Artist extends database_object implements library_item
 
         if (AmpConfig::get('album_group')) {
             $sql = "SELECT MAX(`album`.`id`) AS `id`, `album`.`release_type`, `album`.`mbid` FROM `album` LEFT JOIN `song` ON `song`.`album`=`album`.`id` $catalog_join " .
-                    "WHERE (`song`.`artist`='$this->id' OR `album`.`album_artist`='$this->id') $catalog_where GROUP BY `album`.`prefix`, `album`.`name`, `album`.`album_artist`, `album`.`release_type`, `album`.`mbid`, `album`.`year` ORDER BY $sql_sort"; //TODO mysql8 test
+                    "WHERE (`song`.`artist`='$this->id' OR `album`.`album_artist`='$this->id') $catalog_where GROUP BY `album`.`prefix`, `album`.`name`, `album`.`album_artist`, `album`.`release_type`, `album`.`mbid`, `album`.`year` $sort_disk ORDER BY $sql_sort";
         }
         //debug_event('artist.class', 'get_albums ' . $sql, 5);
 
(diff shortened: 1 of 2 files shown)


lachlan-00:
Thanks, i have a separate mysql server to test on so i'll try it out and get it in.

Thank you for finding it. the queries have really exploded over the last year.

lachlan-00:
I think what i'll also do is set up my dev environments to use full grouping seeing as it's a mysql default.

lachlan-00:
i wonder if this line is backwards.
```
$sort_disk = (AmpConfig::get('album_group')) ? "" : ", `album`.`disk`";
```

If you set this code instead does it work?
```
$sort_disk = (AmpConfig::get('album_group')) ? ", `album`.`disk`" : "";
```

lachlan-00:
although it seems like just putting the disk part into 319 will work instead
```
WHERE (`song`.`artist`='$this->id' OR `album`.`album_artist`='$this->id') $catalog_where GROUP BY `album`.`prefix`, `album`.`name`, `album`.`album_artist`, `album`.`release_type`, `album`.`mbid`, `album`.`year`, `album`.`disk` ORDER BY $sql_sort";
```

KingOperator:
I've tested your proposals and they both work. It's up to you which one to choose. I would prefer the last one (1st solution: more code, 2nd solution: makes no sense from an function perspective).

But it looks like there is another problem with the grouping function. In the artist view I still see two entries for the same album (without disc suffix). In the album view both discs are listed which is ok. I'm going to investigate this further next days.

lachlan-00:
I think you've found the same thing. It probably has to be something like max disk in the select instead.

Group by disk will split them up again making the query do the opposite of what we want

KingOperator:
Hi lachlan,
the duplicate albums were caused by the static album.disk in the default case. I've fixed it by setting the album.disk in the default case depending on the album_group config. This makes the previous fix unnecessary.
Additionally I've found another SQL GROUP BY depending bug in the artist sort function. I've fixed it in the query class. The solution isn't the prettiest but it works.
Sorry for the failed travis build but I don't know what's the problem. Can you please tell me? I'm not experienced with it.
Let me know if the changes are OK for you and if I can further assist.

lachlan-00:
in your git directory you run this before a commit (don't worry i forget it a lot)
```
php ./php-cs-fixer -v fix
```

that runs all the scripts that travis checks.
https://travis-ci.org/ampache/ampache/jobs/643125027?utm_medium=notification&utm_source=github_status

the job log also shows the script output.

I'm glad you had a look and worked it out. it's great to see more people with an interest in the code.

lachlan-00:
btw i'll do a test on the current code. (check your spacing ;)) and i'm happy with your changes so far.

lachlan-00:
Reminder for me that mariadb can set the same grouping as MySQL with this query
```
SET @@SQL_MODE = 'STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,ONLY_FULL_GROUP_BY';
```

KingOperator:
Hi lachlan, thank's for the help with the php-cs-fixer. I saw you've already fixed the spacing. Can I further assist?

lachlan-00:
I just need to test the change on a full group server now.

PR #2284, opened 2020-02-01
Translation Update

Psy-Virus:
As usual new translations from Transifex


lachlan-00:
you usually merge right away, is there anything you need?

Psy-Virus:
grafik

No merge without permissions :D
Thought maybe it's a bug. But I still can't merge...

lachlan-00:
Okay, that shouldn't happen for you at all so let me fix it. I must have been too restrictive when allowing ashot and kuzi into the other branches

lachlan-00:
I've explicitly given the dev group develop and master (I think)

Sorry for stopping that!

PR #2336, opened 2020-03-26
Performance Fixes

Afterster:
I'm giving a hand to Musique-Libre association (https://dogmazic.net/).
Thanks to @luciolebrillante for his help, he did some work to improve performance in Ampache in order to manage the important database dogmazic has (55000 tracks, table with millions of lines). With his agreement, I am basing this pull request on its work in order to include them in the official repository. In that sense I will also generally work on fixing/improving the Ampache performance.

It is unfortunate that both myself first and @wagnered we stopped our work on moving Ampache to Laravel. I believe the current architecture will still remains for a while then... that's why taking a look at the performance after years of adding new stuff when I was the main active maintainer is a good penance :stuck_out_tongue_winking_eye:.

- [x] Lazy load images that may not be important on a few views. I didn't found a lot of place where it could be done.
- [x] Load javascript files as async or defer to improve responsiveness at first load. Right now I added defer attributes only to several javascript files. Not all scripts can be defered or loaded asynchronously of course.
- [x] Store waveform on disk instead of database, according to existing "store art on disk" setting (thanks @fufroma).
- [ ] Reduce overall resources size. Several resources are unnecessary big (2MB font, ...) and may be reduced easily to improve first time page loading.
- [x] Improve SQL request stats based on object_count table. This is a bottleneck on all Ampache installation with several users. It has always been, time to do something.
- [ ] Improve table entries cleanup. Some cleanup requests may take a long time to run if you have a big catalog. May be useful to optionally move such cleanup requests to a cron instead of running it that often at page load...

I'm submitting this PR already for discussion. Feel free to do some suggestion or link to another existing issue.


lachlan-00:
a cron task is actually a really really good idea! I've taken some garbage collection out but it's probably a simpler way of performing cleanup. There are still plenty of tasks that run way too much.

It would be good to hear more opinions on performance. svg images would also be something to look at to reduce sizes.

I'm so happy to have you back and digging into code! we're starting to get a big group again.

luciolebrillante:
Hello,
Running maintenance performance tasks and cleaning tasks are usually grouped together and delayed during a period where the application and the server in general is less requested.

I think in order to improve significantly the performance of the application, it is important to identify which tasks are required to have the application usable and which others tasks could be delayed.

For example, in the fix I did, I run the trending calculation each four hours and I displayed this result when a visitor wants to see it. Yes, it has 4 hours of delay but it avoid to calculate it for each time. You could, of course, recalculate it more frequently if you want.

An option in the menu could be interesting too.

lachlan-00:
I think we should provide an option to cache smart lists (updated by cron) to allow large lists to not require such long queries.

lachlan-00:
I can't edit this branch but i see 85 instances of garbage collection in the code which will help performance by moving out.

I'll start by moving those out into a cron.php

lachlan-00:
merged into the typing branch, i'll be working on this a bit more while i clean up the strict types
image

lachlan-00:
RE: Improve SQL request stats based on object_count table.

Should we extend the column in the song,album,artist table to allow server totals without counting the object_count table?

not all object_count queries could be dropped and the per-user information is pretty good to have available.

Afterster:
I created a new option named "cron_cache" that only compute object_count for now. I don't believe we should add a new column on album,artist... tables as it would remove the capability to know the statistic for a given time period.
@lachlan-00 you are working to move the cleanup on a cron then? If you are working on it I believe reusing the same cron_cache option may make sense. If the cron is not enable on the ampache installation, cleanup unrelated to explicit content deletion action should still be moved to a separate manual action.

lachlan-00:
i'll use AmpConfig::get('cron_cache') to allow the old methods to keep working as is. That will allow us to keep both options available.

The cron.inc file runs all the garbage collection only right now but it only takes about 20seconds to complete for me after an initial 3 minute run.
https://github.com/ampache/ampache/blob/typing/bin/cron.inc

only in typing branch for now but i'm actually pretty happy with this branch and just need to keep testing for type errors before i can merge it.

Afterster:
Good. Next time I will create a branch on the official repo directly it will be easier ;).
Because of this dependency, you can merge my work into your branch if you want to. Better to test btw, and probably merge your bin/cron.inc with my bin/compute_cache.inc? If so, I will take care of further changes (eg. Reduce overall resources size remaining) later.

lachlan-00:
I think that's a good idea if we merge into typing first.

I'm done with the sweeping code changes in the branch now so putting this on top shouldn't hurt and will just let me test both at the same time.

This is actually genuinely exciting to have you and @wagnerd around again.

lachlan-00:
Tomorrow i'll merge these changes into typing and then we will restart the remaining changes in a separate change.

I need to test every function I can as I've put strict typing on every class file but i think I've already gotten close to finishing that as i've been using it for a week without issue now.

lachlan-00:
rolled back a docstring that was blocking this. remaining tasks are to verify that it all works when merged into typing as well as respecting the config options.

Remaining tasks from this pull are:
* Reduce overall resources size. Several resources are unnecessary big (2MB font, ...) and may be reduced easily to improve first time page loading.
* Improve table entries cleanup. Some cleanup requests may take a long time to run if you have a big catalog. May be useful to optionally move such cleanup requests to a cron instead of running it that often at page load...

i'll run php-cs-fixer over the code when it's merged. ;)

AshotN:
Oh cool, @Afterster is back!

I've been inactive myself due to this virus causing my work load to increase. But I hope to get through everything I missed. 

I see a lot of your issues were with the UI, and I am glad someone is refactoring the Reborn interface. The new UI I've been working on is very limited so far, but I'd appreciate your feedback @Afterster.

lachlan-00:
Updating a flac album from tags is sub-second now with these changes. This is a big change that should really be championed in the next release.

I'm almost ready to merge in develop i just haven't had the time to sit down with all the covid/WFH+children stuff.

Afterster:
Great. Yeah covid-19 is disrupting our life. Positive note is that it gives me more time on the weekend to work on side projects ;)

Concerning the performance, page load is much faster on my side for instances with lot of usage (object_count entries). Like 4 time faster. It is promising.

x011:
[CON](https://www.google.com/search?q=con)vid19

PR #2343, opened 2020-04-03
add strict Typing

lachlan-00:


admin/users.php
```diff
@@ -167,8 +167,8 @@
         }
 
         /* Attempt to create the user */
-        $user_id = User::create($username, $fullname, $email, $website, $pass1, (string) $access, $state, $city);
-        if (!$user_id) {
+        $user_id = (int) User::create($username, $fullname, $email, $website, $pass1, (string) $access, $state, $city);
```
AshotN, line 170: Any reason this needs to be casted?

lachlan-00: Construct for user has an int parameter for id. And create returns a string.

There are still a few inconsistencies between ids with strings and ints that need looking at

AshotN: Can't create just return an int?

lachlan-00: I do think that everything id related should be int. That's what we store these as.

You did say it was going to take a while. ;)

lachlan-00:
@Afterster I've added compute_cache to the cron.inc file and expanded cron to include the build_cache options for pretty much everything.

In compute i swapped the echos for debug_events and fixed up the sql that was relying on a global variable. Playlists don't seem to have a decent way of caching and were broken in compute so removed for now.

This cron still takes about 2ish minutes to run and has sped me up multiple times faster than usual. (some things are 10x+ faster than before)

I'll write up the changelog and wiki regarding cron before the merge to really push this change to users. For now i've got to do work work instead of Ampache so this should all happen tomorrow.

pretty exciting!

lachlan-00:
@Afterster the "loading="lazy"" code in the header template breaks the search box.
I'm removing it for now but can you tell me what it's supposed to do?

```
-    itemhtml += "<img src='" + item.image + "' loading="lazy" class='searchart' />";
+   itemhtml += "<img src='" + item.image + "' class='searchart' />";
```

AshotN:
https://web.dev/native-lazy-loading/

What browser are you using that it breaks?

lachlan-00:
Regular Firefox and nightly. There's an error in the console for it until I remove it

lachlan-00:
That would explain it. It's only supported in 75 plus

lachlan-00:
I think my debian was back on like 73/4

AshotN:
If the browser doesn't support the attribute, nothing should happen. That attribute is probably not a good idea to use on Reborn since the interface is really wonky.

Afterster:
That's right, nothing should happen. But if it is unstable on some browser we shouldn't add it yet then. Not a big improvement anyway, that shouldn't be a blocker.

Great for the merge, looks promising :smiley:

lachlan-00:
okay, i've left it out for now. it's not going to affect us too much.

Found issues with podcast typing but just going through the ui now looking for errors

lachlan-00:
one thing i want to add is "last_cron_run" or something into ampache debug so we can see that things are actually running in the interface.

lachlan-00:
hey @Afterster and @AshotN this is going to merge now that i've squashed and documented in the wiki

I am really effing (Australianism?) happy with the performance increases and have tested everything i can. Lots of breakfixes to methods and so many codebugs from phpstorm but this has been a really fun miniproject to complete.

Next for caching i would like to cache smartlists but that's not important enough to hold this back now.

PR #2347, opened 2020-04-07
api subsonic playlist delete/add

DEvmIb:
please review.

### DIFF
--- a/lib/class/subsonic_api.class.php
+++ b/lib/class/subsonic_api.class.php
@@ -881,6 +881,12 @@ class Subsonic_Api
         $playlistId = $input['playlistId'];
         $name       = $input['name'];
         $songId     = (is_array($input['songId'])) ? $input['songId'] : explode(',', $input['songId']);
+        $songId  = array();
+        if (is_array($input['songId'])) {
+            $songIndexToRemove = $input['songId'];
+        } elseif (is_string($input['songId'])) {
+            $songIndexToRemove = explode(',', $input['songId']);
+        }
 
         if ($playlistId) {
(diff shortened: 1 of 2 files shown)


lachlan-00:
can you give me some detail about what it does? What situation causes subsonic not to send an array?

It seems off the top of my head that updatePlaylist calls need to enforce an array instead.
self::_updatePlaylist($playlistId, $name, array($songId));

DEvmIb:
when client sends an update for multiple songs 

`&songIdToAdd=&songIdToAdd=&songIdToAdd=`

 we get an array but $songsIdToAdd_count is not set to the count therefore the loop to adding songs never run. 

on removing songs the client send 

`&songIndexToRemove=0&songIndexToRemove=1&songIndexToRemove=2&songIndexToRemove=3. `


starting from 0 therefore  we need to use "isset" and +1.


 i hope you understand me right :)

lachlan-00:
Thanks for adding the explanation when i get typing merged into develop i'll start testing on this one. i haven't seen issues with add/remove but maybe i'm missing something so it'll be good to sit down with it and check.

lachlan-00:
sorry, didn't change my branch back so i've reverted that comment change

lachlan-00:
trying to make this a bit smarter by enforcing the input parameters before going into the function.

also less if's the better, will test asap

lachlan-00:
thanks for becoming an ampache contributer!

I'm getting errors in dsub but i think that's dsub related as the server isn't asked to update the playlist.

but now i can add and remove the tracks in any size and order. it's amazing that i've never actually done this!

PR #2348, opened 2020-04-10
Fix playlisttrack field in API

AshotN:
I'm not sure why the playlisttrack field had a whole function dedicated to it. Just in case I missed something I'd appreciate a review. 

This fixes a problem where a playlist contains the same song twice, the `playlisttrack` field would be the same for both songs.


lachlan-00:
i think this is a cleaner way to do it by far, when we pull a playlist does the array list in track order from the DB?
image

if that's the case then this simplifies the whole process and means we get the order from the db and then just dump the tracks in a clear order. (no dupes in the list, 0 tracks, etc)

lachlan-00:
I have had a look through it now and the SQL query sets the order using track so i actually don't see this function being needed.

it actually seems a bit useless so i think this should go in.

I think we take is a step further and include playlisttrack in all songs lists by taking out the parameter $playlist_data and just including it everywhere.

you've generated a list of songs so we should track them in order.

lachlan-00:
made some changes, let me know if you like that. I think that because the order is set with the sql this is a pretty easy way to do it.

also cleaned up the xml_data calls to match json, makes it a bit clearer/cleaner.

AshotN:
I think that’s fine as long as when we document the API we can make it clear playlist_track refers to the index of the returned songs.

lachlan-00:
Aaaaaand I broke something last minute in typing. Reverted and I'll work on fixing that today, then this

lachlan-00:
okay, merging now, api needed a bump to reflect this. we'll be working on a few more API things as well so this is a start of 4 build 5.

It is kind of odd that we never gave a list of track order correctly
one of the processes was putting object_id in stead of track! (i think my db is up to 9 million?)

PR #2352, opened 2020-04-15
Typing Round 2

lachlan-00:


lib/class/json_data.class.php
```diff
@@ -417,6 +424,10 @@ public static function playlists($playlists, $create = false)
      *
      * This returns a JSON document from an array of song ids.
      * (Spiffy isn't it!)
+     * @param $songs
+     * @param array $playlist_data
+     * @param boolean $user_id
+     * @return false|string
```
AshotN, line 430: I notice there are a lot of instances of `false|string` in the new typing update.

Wouldn't it make more sense to return `null` instead of `false` when nothing is found. 

And if an error occurs, we should throw the exception.

lachlan-00: That's definitely a change we should make. so may functions are allowed to do nothing when they shouldn't even be run.

PR #2353, opened 2020-04-16
Unnessary null check for XML tags function

AshotN:
The third parameter for `array_splice` defaults to `null` if not provided. So this was an unnecessary check as both calls are identical.

`function array_splice (array &$input, $offset, $length = null, $replacement = null) {}`

### DIFF
--- a/lib/class/xml_data.class.php
+++ b/lib/class/xml_data.class.php
@@ -422,11 +422,7 @@ class XML_Data
     public static function tags($tags)
     {
         if (count($tags) > self::$limit || self::$offset > 0) {
-            if (null !== self::$limit) {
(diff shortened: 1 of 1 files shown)


lachlan-00:
put some null in your null so you can null when you null

AshotN:
I just noticed this occurs in more than one place. I'll create a more comprehensive pull to address it.

lachlan-00:
that's going to help reduce complexity a lot

lachlan-00:
want to do it in this branch or start a new one?

AshotN:
Only a couple more occurrences. I'm refactoring the JSON API, and while doing so whatever issue I catch in the XML I'll just create a pull for. 

The JSON changes I'll put into a separate branch, as they are more breaking changes.

PR #2366, opened 2020-04-27
[Gravatar Plugin] Make sure https is used when force_ssl is configured

rsertelon:
I discovered the issue with my own ampache installation. I had mixed-content warnings, while behind a reverse proxy with `force_ssl = true` in my ampache config file.

Note: I tested on my own install, it works as expected.

### DIFF
--- a/modules/plugins/Gravatar/Gravatar.plugin.php
+++ b/modules/plugins/Gravatar/Gravatar.plugin.php
@@ -79,7 +79,8 @@ class AmpacheGravatar
     {
         $url = "";
         if (!empty($user->email)) {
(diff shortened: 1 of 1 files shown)


lachlan-00:
is it okay without the empty? That's a good thing to catch i was just talking about this option today.
```
if (AmpConfig::get('force_ssl') ||
```

rsertelon:
I reused the same condition as in the [init.php](https://github.com/ampache/ampache/blob/develop/lib/init.php#L74) file for consistency.

Testing for the case where the parameter is undefined in the config file will be less easy for me, since I only have my live instance which requires the force_ssl param to work :)

lachlan-00:
Thank you for clarifying! Welcome to contributor status!

Issue #2374, opened 2020-05-05
Subsonic API returns lyrics of wrong song


jleipsig:
## Description
When retrieving song lyrics with Subsonic API, the lyrics of another song are returned. You always get the lyrics of the same song, no matter what request you send to the API.

In 'getlyrics' of Subsonic_Api, a query is built with 'rule operator' set to index 5. This turns out to be the operator 'not equal'. Changing it to index 4 ('equal') solves the problem.

A patch for version 4.1.1  is attached
[Change_operator_for_query_in_getlyrics_from_'not_equal'_to_'equal'.patch.gz](https://github.com/ampache/ampache/files/4581187/Change_operator_for_query_in_getlyrics_from_.not_equal._to_.equal.patch.gz)

## Environment

* Ampache version: 4.1.1
* Web server + version: Apache 2.4.38-3+deb10u3
* Server operating system: Debian Buster
* Client operating system: Android

### Client type

* SubSonic

## Logs

apache.log
```
192.168.1.10 - user [04/May/2020:23:54:38 +0200] "GET /ampache/rest/getLyrics.view?u=user&p=enc:key&v=1.2.0&c=DSub&artist=Some+Artist&title=Some+Title HTTP/1.1" 200 2350 "-" "DSub"
```
ampache.log
```
2020-05-04 23:54:38 [ampache] (access.class) -> check_network rpc: ip matched ACL 192.168.1.10
2020-05-04 23:54:38 [user] (subsonic_xml_data.class) -> API success in function getlyrics-1.13.0
```

lachlan-00:
what a great find! glad to see that you were able to parse the rules/operator parts of the code.

lachlan-00:
merged your pull

lachlan-00:
@jleipsig could you tell me how you found the difference in operator and how we could show that information better?

i really struggled as a newbie to the code and i want to make it easier forothers.

wagnered:
@lachlan-00 wrote:

> @jleipsig could you tell me how you found the difference in operator and how we could show that information better?

I'm not sure he is 'watching' this comment, so I took it upon myself to investigate since I have nothing better to do :smile:

I believe that the index of the operators are set in `Search::set_basetypes()`.  Although the order of operators do not agree with @jleipsig's comment, he might have used the 'trial and error' method.

[...]

PR #2375, opened 2020-05-05
Fix for issue #2374: Subsonic API returns lyrics of wrong song

jleipsig:
Change operator for query in getlyrics from 'not equal' to 'equal'.
This way an SQL query is built that fetches song lyrics with Artist and Title **equal** to the given parameters instead of songs with Artist and Title **not equal** to the given parameters (which returns always the same song).


lachlan-00:
welcome to the Ampache contributor club!

Issue #2406, opened 2020-06-11
Song with long title doesn't get added


jleipsig:
## Description
Songs with a ridiculously long title don't get added 

### Describe the bug
When trying to add files to a catalog with a very long title, a database error occurs:
Data too long for column 'title'

### To reproduce
The title tag of the song must contain a string longer than 255 bytes.
MYSQL sql_mode STRICT_TRANS_TABLES must be set (default for my environment)

### Screenshots

![long_filename](https://user-images.githubusercontent.com/14284810/84424629-ae371d80-ac20-11ea-9e82-8912edeade23.png)

## Environment

* Ampache version: develop (5d82ead lachlan-00 github@dewaard.email on 6/10/20 at 12:57 PM)
* Web server + version: nginx/1.17.9
* Server operating system: Debian GNU/Linux 10 (buster)
* Client operating system: Debian GNU/Linux 10 (buster)
* Database: MariaDB 10.2.31

## Settings
MariaDB [db]> SELECT @@SESSION.sql_mode;
+-------------------------------------------------------------------------------------------+
| @@SESSION.sql_mode                                                                        |
+-------------------------------------------------------------------------------------------+
| STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
+-------------------------------------------------------------------------------------------+
1 row in set (0.000 sec)

## Logs
2020-06-11T18:36:55+02:00 [admin] (vainfo.class) -> Checking /(.+?)\/(.+?)\/([0-9]+?)\s\-\s(.+?)\..+$/ _ Array on Sufjan Stevens/Illinois/02 - The Black Hawk War.flac 
2020-06-11T18:36:55+02:00 [admin] (vainfo.class) -> /(.+?)\/(.+?)\/([0-9]+?)\s\-\s(.+?)\..+$/ matched Sufjan Stevens/Illinois/02 - The Black Hawk War.flac on Sufjan Stevens/Illinois/02 - The Black Hawk War.flac 
2020-06-11T18:36:55+02:00 [admin] (vainfo.class) -> Cleaning id3v1 
2020-06-11T18:36:55+02:00 [admin] (vainfo.class) -> Cleaning id3v2 
2020-06-11T18:36:55+02:00 [admin] (vainfo.class) -> Cleaning vorbis 
[...]

PR #2408, opened 2020-06-11
truncate song title to max 255 bytes, before insert into db

jleipsig:
fix for #2406

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -2289,6 +2289,7 @@ abstract class Catalog extends database_object
         } else {
             $string = trim(substr($string, 0, $max_length));
         }
+
         return $string;
     }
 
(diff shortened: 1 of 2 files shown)


lachlan-00:
we should do this for artist, album and title as well.

do you want to move sanitize_title to say lib/general and update the artist/album inserts as well?

happy to do it for you but i like people to have their contributions fully theirs.

jleipsig:
Moved the `sanitize_title` function to `Catalog::check_length`, because most other sanitizing stuff is also in Catalog. After consulting the MYSQL documentation and some more testing (both a bit too late :)), I changed the check from max byte length to max character length.
Maybe I took it a bit too far, but the check is now added for artist, album_artist, album, composer, label, language, catalog_number and barcode :)

`Song::insert` and `Catalog::update_song_from_tags` have a lot of (almost) duplicate code.  `Catalog::update_song_from_tags` cleans lyrics before storage, `Song::insert` doesn't, resulting in unnecessary updates.
Also, when catalog_number or barcode are changed in file tags, `Catalog::update_song_from_tags` doesn't update them.

lachlan-00:
I'll double check the database but if it covers it all then no problem.

lachlan-00:
Thanks for all this work you're a bloody star.
i'm starting to go through these all now

Issue #2409, opened 2020-06-13
metadata field list not removed when choosing another search option


jleipsig:
## Description
When choosing 'Metadata' as search option in song search (/#search.php?type=song) and then choosing another search option e.g. 'Any searchable text', the select list with metadata fields is not removed. When repeating this step, multiple select lists with metadata fields are created and not removed.

### Screenshots

![search-metadata](https://user-images.githubusercontent.com/14284810/84568390-7eeaf280-ad7f-11ea-900e-a365a398bf0b.png)

## Environment

* Ampache version: develop (78e53cdb lachlan-00 <lachlan.00@gmail.com> on 6/12/20 at 11:48 AM)
* Browser: Firefox 68.9.0esr (64-bits)

### Client type

* Web

lachlan-00:
another great find.

PR #2410, opened 2020-06-13
correctly remove metadata field (subtype) select list

jleipsig:
fix for #2409


lachlan-00:
I see what you've done for removing elements that end with subtype but this doesn't seem to be enough for me in firefox?
image

chromium browsers are fine. i'll just check on another pc in case it's cached something

lachlan-00:
must be my session, working correctly in ff dev

Issue #2412, opened 2020-06-14
search song by artist name returns only one song for each artist


jleipsig:
## Description
When searching in songs on artist name, only one song is returned for each artist that matches the search.

Go to search page #search.php?type=song and select Artist as search field, any operator (e.g. 'ends with') and any input text. Hit search. For the search in the screenshot, the build sql string in `Search::run` is:
`SELECT DISTINCT(`song`.`id`) FROM `song` LEFT JOIN `artist` ON `song`.`artist`=`artist`.`id` WHERE (`artist`.`name` LIKE '%Wilson'  OR LTRIM(CONCAT(COALESCE(`artist`.`prefix`, ''), ' ', `artist`.`name`)) LIKE '%Wilson') GROUP BY `artist`.`id`
`

### Expected behavior
I expect all songs for all matching artists to show up

### Screenshots
![search-artist](https://user-images.githubusercontent.com/14284810/84600536-306e4e80-ae7a-11ea-9ca2-55aafe2d5849.png)

## Environment

* Ampache version: develop (78e53cdb lachlan-00 <lachlan.00@gmail.com> on 6/12/20 at 11:48 AM)
* Web server + version: nginx/1.17.9
* Server operating system: Debian GNU/Linux 10 (buster)
* Client operating system: Debian GNU/Linux 10 (buster)
* Database: MariaDB 10.2.31

### Client type

* Web

lachlan-00:
I never noticed but you can't unsee after that.

PR #2413, opened 2020-06-14
remove group on artist when searching by artist in songs

jleipsig:
fix for #2412


lachlan-00:
does look unnecessary. must be from me copy/pasting blocks when cleaning up search

tests fine

Issue #2414, opened 2020-06-14
[Feature Request] Some search enhancements


jleipsig:
I would like to add some features to search to make it even more powerful.

- Search in lyrics. Other fields from song_data (comment, label) are also searchable.
- Search with regular expressions in text fields. MYSQL supports it, and it gives a lot of extra possibilities.
- Search for non-existent metadata. When searching for an empty string in a metadata field with operator 'is', there are never results because empty values are not stored. You could use this to search for songs that don't have the given metadata field at all.

![search_enhancement](https://user-images.githubusercontent.com/14284810/84602059-053d2c80-ae85-11ea-8ce9-bd1cba1f0946.png)

This search translates to SQL:
`SELECT DISTINCT(`song`.`id`) FROM `song` LEFT JOIN `song_data` ON `song`.`id`=`song_data`.`song_id` WHERE `song_data`.`lyrics` REGEXP 'colo[u]?r' AND NOT EXISTS (SELECT NULL FROM `metadata` WHERE `metadata`.`object_id` = `song`.`id` AND `metadata`.`field` = 28)
`

lachlan-00:
one thing i see here is dupe names of the operators. (this isn't from you the metadata operators are a merger of text and numeric which both have the same names)

should look at changing that on probably the numeric side
![image](https://user-images.githubusercontent.com/1305249/84618719-56e1c780-af16-11ea-937c-05e98403b757.png)

lachlan-00:
okay fixed that

PR #2415, opened 2020-06-14
add search in lyrics

jleipsig:
Proposal for feature request #2414

### DIFF
--- a/lib/class/search.class.php
+++ b/lib/class/search.class.php
@@ -210,6 +210,19 @@ class Search extends playlist_object
             'description' => T_('does not sound like'),
             'sql' => 'NOT SOUNDS LIKE'
         );
+
+        $this->basetypes['text'][] = array(
+            'name' => 'regexp',
+            'description' => T_('matches regular expression'),
+            'sql' => 'REGEXP'
+        );
+
+        $this->basetypes['text'][] = array(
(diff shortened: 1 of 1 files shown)


lachlan-00:
i should have something with tagged lyrics, testing now

kuzi-moto:
This would be a great thing to add, I know it's been requested before. I haven't looked at the code for this yet, but would this work in just the advanced search, or also in the global search at the top of the page? 

Additionally, does the global search have a priority, such as showing matches first by Artist, then Album, then title, etc.? If it does, might be good to make sure lyrics have a lower priority.

lachlan-00:
it's a text search so it would be possible to add.

global runs the search.php 
* the search from the header is hardcoded to 0/contains
* it looks at a glance that it only searches Artist, album, title

The results window loads if you search and lets you dig deeper

lachlan-00:
If you search anywhere it does one search for each Artist, then album, then song then outputs the results.

the ajax/JS waits for 2 characters then executes the search each keypress.

I think it's fine keeping it in advanced_search as even just these like searches can be heavy

lachlan-00:
@kuzi-moto  move into #2414 if you want, i've tested this out and the searches are working correctly.

i've also documented this in the wiki as well.

Issue #2418, opened 2020-06-16
TXXX custom tags not imported from id3v2


jleipsig:
## Description
When importing or updating songs with id3v2 tags, tags that are stored in the '#TXXX User defined text information frame' are not imported, except musicbrainz, replaygain and some other predefined tags.
Custom tags from flac files are all stored.
My config has `enable_custom_metadata = "true"`. In my preferences no fields are selected in 'Custom metadata - Disable these fields'.
There are no error messages in the log.

### Expected behavior
When i have `enable_custom_metadata = "true"`, I expect all tags to be stored in ampache that are not disabled in preferences.

### Screenshots

![afbeelding](https://user-images.githubusercontent.com/14284810/84834918-c947ca00-b032-11ea-9fba-277a0777cee3.png)
![no_tag](https://user-images.githubusercontent.com/14284810/84835153-3c514080-b033-11ea-81a8-a3f3ff1874a0.png)

## Environment

* Ampache version: develop (cacd6bda lachlan-00 <lachlan.00@gmail.com> on 6/16/20 at 6:55 AM)
* Web server + version: nginx/1.17.9
* Server operating system: Debian GNU/Linux 10 (buster)
* Client operating system: Debian GNU/Linux 10 (buster)

### Client type

* Web

## Logs
2020-06-17T00:38:47+02:00 [admin] (local.catalog) -> Reading 01 - Where is my precious.mp3 inside /var/www/music 
2020-06-17T00:38:47+02:00 [admin] (local.catalog) -> Memory usage: 2 MB 
2020-06-17T00:38:47+02:00 [admin] (vainfo.class) -> Checking /(.+?)\/(.+?)\/([0-9]+?)\s\-\s(.+?)\..+$/ _ Array on www/music/01 - Where is my precious.mp3 
2020-06-17T00:38:47+02:00 [admin] (vainfo.class) -> /(.+?)\/(.+?)\/([0-9]+?)\s\-\s(.+?)\..+$/ matched www/music/01 - Where is my precious.mp3 on www/music/01 - Where is my precious.mp3 
2020-06-17T00:38:47+02:00 [admin] (vainfo.class) -> Cleaning id3v2 
2020-06-17T00:38:47+02:00 [admin] (album.class) -> Album check created new album id 66 
2020-06-17T00:38:47+02:00 [admin] (local.catalog) -> Finished reading /var/www/music , closing handle

PR #2419, opened 2020-06-16
Txxx metadata

jleipsig:
Allow ID3v2 TXXX tags to be imported. Proposal for #2418


lachlan-00:
i must have the ufid frame because i get all the meta no problems. i'll do a verify on my catalog with this first before merging

lachlan-00:
adds in more fields so a lot would be missing from a lot of libraries.

PR #2422, opened 2020-06-17
Changed label wording on some preference settings

wagnered:
It' a matter of semantics, but on/off works for me.


lachlan-00:
@Psy-Virus there have been a few string changes and i think we'll almost be ready for master next month.

What i find funny is that we thought about language so much but because we had conditioned ourselves to this language it wasn't noticed.

kuzi-moto:
Nice, and Ampache becomes a little more coherent.

Psy-Virus:
> 
> 
> @Psy-Virus there have been a few string changes and i think we'll almost be ready for master next month.
> 
> What i find funny is that we thought about language so much but because we had conditioned ourselves to this language it wasn't noticed.

All right, I'll update the translations in the next minutes and again in the first few days next month.

Psy-Virus:
@lachlan-00,
is there already a planned Master-Upgrade date?
There are still some people actively translating and I'd like to wait a bit longer to let them finish more strings. But if you have a deadline, I'll submit and merge the latest translations before it.

lachlan-00:
> 
> 
> @lachlan-00,
> is there already a planned Master-Upgrade date?
> There are still some people actively translating and I'd like to wait a bit longer to let them finish more strings. But if you have a deadline, I'll submit and merge the latest translations before it.

no i'd rather wait for better translations than release earlier.

PR #2423, opened 2020-06-18
Fixed transfer of local db user to step two during installation

wagnered:


wagnered:
Sorry about the mispelling.

lachlan-00:
took a bit to work out the logic there but that's a good change.

Issue #2425, opened 2020-06-18
[Feature Request] sort metadata field list on advanced search alphabetically


jleipsig:
## Describe what you don't like about Ampache, or what it is missing
The list of metadata fields on the advanced search page is sorted on id of the metadata field. An alphabetically sorted list would be nice.

lachlan-00:
i've been playing around with this a bit more and i see that meta fields aren't removed when they're missing from the song anymore.

ie. i removed the tag from a song and it didn't remove from the database unless the song was removed totally.

jleipsig:
Yes, that's already on my list of metadata things. Maybe you prefer to pick this one and/or one of the following?:

1. The select list 'Custom metadata - Disable these fields' on 'preferences.php?action=admin&tab=system' doesn't work. `scub_in` translates it to string 'Array'.
2. The name of preference 'Custom metadata - Define field list' doesn't make clear it's a comma separated list of fields that will be disabled. 
3. Fields defined in `disabled_custom_metadata_fields` and `disabled_custom_metadata_fields_input` are not disabled in advanced search.
4. Sometimes metadata tags are stored under different names in id3v2 and vorbis etc.  (f.e. see https://picard.musicbrainz.org/docs/mappings/). It would be nice to have some sort of tag mapping to store these values in the same field. For example store 'TAG_NAME' as 'Tag Name'. Because ampache is case insensitive, maybe it's better to store all tags in one format (lowercase) and pick a default way to present them (ucfirst/ucwords/uppercase/...) overruled by the case for defined tags in your tag mapping.

PR #2426, opened 2020-06-18
case insensitive sort on metadata field list in advanced search

jleipsig:
proposal for feature #2425 

sorting is done in javascript. Because the subtypes are stored as objects in `var types` via search-data.php, it's not possible to do the sorting on the server side.

### DIFF
--- a/lib/javascript/search.js
+++ b/lib/javascript/search.js
@@ -185,7 +185,7 @@ var SearchRow = {
         if (type.widget[0] == "subtypes") {
             var $select = SearchRow.createSelect({
                 name: "rule_" + targetID + "_subtype"
-            }, type.subtypes);
+            }, type.subtypes, undefined, true);
             $(this).after($select);
         }
         else {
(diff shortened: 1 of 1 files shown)


lachlan-00:
So to match subtypes like
Label
LABEL
label?

It would be good to possibly lower case meta inserts as well to avoid creating dupes like that

lachlan-00:
Oh, sorry alpha sorted!

lachlan-00:
it actually just needs to get sorted here in search.class
image

jleipsig:
That was my first thought also. But when this becomes a javascript object, the sorting gets undone

jleipsig:
When sorting on the server side, the structure of the json data of the subtypes has be be changed from an object with { key1 : value1, key2 : value2 } to an array like [ {key1 : value1}, {key2 : value2} ]

lachlan-00:
so should it go back further to the original query part as well? the matedata fields are selected in Repository.php (not sure if there are other tables using this)
image

jleipsig:
That would be a great solution to also get the metadata field sorted on the preferences page. For the presentation on the advanced search, there is still something needed in the translation in lib/javascript/search-data.php arrayToJSON($search->types)
The sorting is not preserved in javascript, when presented as an object (see previous comment).

lachlan-00:
ahh, i'm with you now. i'll sort and you JS here.

Issue #2429, opened 2020-06-20
songs with negative or big track numbers are not imported 


jleipsig:
## Description
When adding a song with a negative or big (>65535) track number, the insert fails
When updating songs, big numbers are (sometimes) truncated to 5 digits.

### To reproduce
Give the file a big (>65535) or negative tracknr, e.g. with kid3-cli:
kid3-cli /home/me/music/test2.mp3
kid3-cli> set 'Track Number' 975287
kid3-cli> save
kid3-cli> get
MPEG 1 Layer 3 170 kbps 44100 Hz Stereo 6:35
  Name: test2.mp3
Tag 2: ID3v2.3.0
  Title         Test 2
  Artist        Test
  Date          2020
  Track Number 975287

Do import/update in ampache

kid3-cli /home/me/music/test2.mp3
kid3-cli> set 'Track Number' -10
kid3-cli> get
MPEG 1 Layer 3 170 kbps 44100 Hz Stereo 6:35
  Name: test2.mp3
Tag 2: ID3v2.3.0
  Title         Test 2
  Artist        Test
  Date          2020
  Track Number -10

Do import/update in ampache

### Expected behavior

I would expect the track to be added.
Because the import fail is 'silent', it's hard to find out what tracks are not imported because of errors.

## Environment
* Ampache version: develop (580b1417 Manuel <psy-virus@users.noreply.github.com> on 6/20/20 at 12:55 PM)
* Web server + version: nginx/1.17.9
* Server operating system: Debian GNU/Linux 10 (buster)
* Client operating system: Debian GNU/Linux 10 (buster)
* Database: MariaDB 10.2.31

### Client type

* Web
* Ampache

## Logs
2020-06-20T16:57:12+02:00 [admin] (dba.class) -> Error_query SQL: INSERT INTO `song` (`catalog`, `file`, `album`, `artist`, `title`, `bitrate`, `rate`, `mode`, `size`, `time`, `track`, `ad
dition_time`, `year`, `mbid`, `user_upload`, `license`, `composer`, `channels`) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) 
2020-06-20T16:57:12+02:00 [admin] (dba.class) -> Error_query MSG: ["22003",1264,"Out of range value for column 'track' at row 1"] 
2020-06-20T16:57:12+02:00 [admin] (song.class) -> Unable to insert /home/me/music/test2.mp3

[...]

PR #2430, opened 2020-06-20
Negative tracknr

jleipsig:
sanitize track numbers, to prevent import fails
  negative track number are stored with an empty value
  big track numbers (>65535) are truncated to last 4 digits

### DIFF
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -2303,6 +2303,7 @@ abstract class Catalog extends database_object
     public static function check_tracknr($tracknr)
     {
         $retval = ((int) $tracknr > 65535) ? (int) substr($tracknr, -4, 4) : (int) ($tracknr);
+
         return ($retval >= 0) ? $retval : null;
     }
 
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -2298,13 +2298,13 @@ abstract class Catalog extends database_object
      * Check to make sure the tracknr fits into the database: unsigned, max 65535
      *
      * @param string $tracknr
-     * @return int null if tracknr is negative
+     * @return int
      */
     public static function check_tracknr($tracknr)
     {
-        $retval = ((int) $tracknr > 65535) ? (int) substr($tracknr, -4, 4) : (int) ($tracknr);
+        $retval = abs((int) $tracknr);
 
-        return ($retval >= 0) ? $retval : null;
+        return ($retval > 65535) ? (int) substr($retval, -4, 4) : $retval;
     }
 
     /**
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -2295,14 +2295,14 @@ abstract class Catalog extends database_object
 
     /**
      * check_tracknr
-     * Check to make sure the tracknr fits into the database: unsigned, max 5 digits
+     * Check to make sure the tracknr fits into the database: unsigned, max 65535
      *
      * @param string $tracknr
      * @return int null if tracknr is negative
      */
     public static function check_tracknr($tracknr)
     {
-        $retval = (strlen((string) $tracknr) > 5) ? (int) substr($tracknr, -5, 5) : (int) ($tracknr);
+        $retval = ((int) $tracknr > 65535) ? (int) substr($tracknr, -4, 4) : (int) ($tracknr);
         return ($retval >= 0) ? $retval : null;
     }
 
--- a/lib/class/catalog.class.php
+++ b/lib/class/catalog.class.php
@@ -1813,7 +1813,7 @@ abstract class Catalog extends database_object
             // fall back to last time if you fail to scan correctly
             $new_song->time = $song->time;
         }
-        $new_song->track    = (strlen((string) $results['track']) > 5) ? (int) substr($results['track'], -5, 5) : (int) ($results['track']);
+        $new_song->track    = self::check_tracknr($results['track']);
         $new_song->mbid     = $results['mb_trackid'];
         $new_song->composer = self::check_length($results['composer']);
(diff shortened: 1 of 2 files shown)


jleipsig:
proposed fix for #2429

wagnered:
If there is no other way, I'll reluctantly agree with your solution. But I think there is.

I realize that error catching and debugging wasn't given much thought during Ampache's early days, but please give this problem some more thought. All the information does exist in the $params variable and in the $sql variable. A function could be created to examine the contents of those variables and issue a debug event with the suspect info. 

There is also PDOStatement::debugDumpParams` which might work out with server-side include(?), but doesn't seem promising.

We appreciate your enthusiasm in improving Ampache.

lachlan-00:
Oh you better believe I'm putting in debugDumpParams in the code! I think it would be good to put that in for errors instead of the current debug at the very least

What are some examples of track numbers being so large as well? I can't say I've seen anything over 999 in my life.

jleipsig:
I had several tracks with negative track numbers that didn't get imported. For a long time I didn't notice that they didn't get imported. After I found out it would be  easy to fix the tags and don't bother. But I wanted to help others with the same messy tags.
I never encountered the big numbers, but that was already in the code, so I kept it ( in an 'improved' way)
No problem if this is not the right fix. Now I'm going to finish my glass of wine. It's almost 3 am ...

lachlan-00:
Haha, don't worry about that. Wine on the weekend is prime coding time.
I tend to work during the week in between my regular work so I'll start this tomorrow.

Keep going though. The amount of improvement we've had in the past week is amazing!

kuzi-moto:
Would it be safe to assume that track numbers larger than say, 1000 and lower than 0 are likely errors? Maybe there could be some logic to assume the correct track number? If metadata is incorrect, look at the track title, or just increment +1 from the last track, then any other tracks that were incorrect would at least have a sane track number (even if it were incorrect).

Alternatively maybe Ampache could alert the user to the error in the interface to allow manual intervention.

jleipsig:
Guessing the correct value could help, certainly in albums with successive track numbers. But I would prefer to leave the value as is or give the user enough information to correct it. F.e. I have one file '903_artist_title_year.mp3' with track number -121. It's in a folder with some singular tracks. Personally I would prefer to give it an empty value or -121, not a guessed value of 121 or 903.

I agree it would be very helpful to alert the user on errors (preferably without halting the import/update process), and supply enough information to allow some corrections.

The ID3v2 spec doesn't say the TRCK frame should contain a _positive_ number, although the description 'the order number [...] on its original recording' doesn't leave much room for negative values.
But I prefer the import/update to be forgiving towards funny data. Another solution for this negative track number issue, could be to alter the table field song.track into a signed number, so you don't need checks/guesses in the code. Maybe MEDIUMINT, to prevent the loss of existing values > 32767 and be a bit more lenient for some 'freak' track numbers possibly existing in this universe. It's just one byte / track extra storage.

(BTW thanks for the quick and constructive comments, that's what keeps it going. Wine on the weekend less so...)

lachlan-00:
moving to signed MEDIUMINT over just flipping the table column to unsigned would cover any existing 65k track numbers (i know there wouldn't be many but we should be compatible to all existing databases) and allow negative ints,

Moving to that would cover pregap tracks like these.
image

but my only question then is the storage would increase from 2byte to 3bytes per track number?
https://dev.mysql.com/doc/refman/5.7/en/integer-types.html

That's a negligible number in the scheme of things but something to consider. After googling around i think supporting negative numbers is valid for albums.

wagnered:
I don't think it matters much how large an integer to accept, as long as the upper limit is documented or just truncate the positive number. 

As for a negative number, what about getting the absolute value, then determine  if it is within the documented limit, or just  truncate it maybe with a warning.

lachlan-00:
I would like to go the other way and allow negative numbers as they do exist in commercial albums
https://en.wikipedia.org/wiki/List_of_albums_with_tracks_hidden_in_the_pregap

jleipsig:
Should I make a new pull request with the database change and close this one?

lachlan-00:
I always vote with what albums do as what ampache should do.
If there are negative ints as a possible value i think we should support them.

My vote for what to do is a database update to signed int and truncate existing values that are higher first. @wagnerd what's your opinion on this? 

Apparently 300 is the (current) highest number of tracks on an album so i think we keep it the same size and view any data higher than the 32k limit as invalid.
image

We should probably keep check_tracknr as some kind of limit function. testing upper/lower values (default to 0 for lower and require an upper limit) The added data checks have made the import process a lot stronger. something like:
```
check_limit($upper, $lower = 0)
```

We should close this one off and move back to #2429; as it's been a good discussion but the pull doesn't actually solve the original issue. @jleipsig you've really brought on some good discourse and made some really valuable changes here. :)

wagnered:
it sure would be nice to know the purpose of a negative track or album number. Couldn't find any reference to the subject in a web search.  The only reason I can think of is that people make mistakes so let the user eventually find it himself/herself or because "Everyone else is doing it". 😃

Actually, I like the PR in which @jliepsig uses the absolute value which seems to adhere the best to id3.org's i  But it doesn't fix the premise that people make mistakes. I just don't think an architectural change should be made to Ampache unless absolutely necessary. 

I thought this PR would take care of #2429. What am I missing?

lachlan-00:
negative track numbers are used for pre-gap / unlisted tracks. which is commonly use as -1 instead of 0

wagnered:
Ok, I always thought that track 0 was meant to be an introduction or verbal explanation of the album content, not an actual song.  I'm guessing that a negative numbered track would never be found on an original cd or album purchased and downloaded online.

Thank you for the information.  I don't know what it means, but if that kind of data is significant, then definitely convert it to signed integer.

PR #2440, opened 2020-07-05
Fix problem when an artist has a dash in its name, like AC-DC.

woodspire:
This fix will correctly grab the song name from the "artist" "song number" "song name" string.

### DIFF
--- a/lib/class/webdav_directory.class.php
+++ b/lib/class/webdav_directory.class.php
@@ -74,7 +74,7 @@ class WebDAV_Directory extends DAV\Collection
     {
         // Clean song name
         if (strtolower(get_class($this->libitem)) === "album") {
-            $splitname = explode('-', (string) $name, 3);
+            $splitname = explode(' - ', (string) $name, 3);
             $name      = trim((string) $splitname[count($splitname) - 1]);
             $nameinfo  = pathinfo($name);
             $name      = $nameinfo['filename'];


lachlan-00:
so webdav tries to identify songs by splitting the text?

Changing this would also just start affecting artists with the spaces (' - ') 
```
Aga - Wilk - 6 - Driving Me Mad.mp3
```

I also see that if i have a various artist album it's empty when i go to the artist themselves.
image

This all needs to be done better.

lachlan-00:
I haven't looked into this but is there a reason why it doesn't use the folder path instead? that would avoid making simple mistakes like this

lachlan-00:
oh and a third problem is webdav plays aren't recorded for the user.

lachlan-00:
@woodspire if you keep the ' - ' and take off the 3 at the end this should cover most cases. as well

```
$splitname = explode(' - ', (string) $name);
```

lachlan-00:
in webdav file add in a stat insert
image

(haven't worked out if the user is stored anywhere but it would be good to at least record that stat)
```
Stats::insert($this->libitem->type, $this->libitem->id, 0, 'Ampache WebDAV', null, 'stream');
```

lachlan-00:
So the only thing i see left is that the directory structure includes artists and albums without songs but i don't know enough about SABRE\DAV to work that one out but we can ignore that as it's not really affecting this pull

lachlan-00:
closing old pulls.

this is fixed in recent versions and uses the filename instead of a weird combined list

PR #2449, opened 2020-07-14
Spotify

wagnered:
Added Spotify for grabbing both album and artist images.


lachlan-00:
it looks like your develop is a bit behind, am i able to update it for you?

lachlan-00:
scratch that, i've updated the branch.
Does spotify have an API site we can point them to?

wagnered:
The url to sign up for an account: https://www.spotify.com/us/

Then there is the dashboard  https://developer.spotify.com/dashboard/ where one can sign up, get the id and secret and documentation on the api.

wagnered:
Another thing, I've been pulling changes from ampache/ampache. I don't understand why I've been getting merge conflicts  and apparently some files not being updated.  I haven't made any changes in the develop branch. So there shouldn't be any conflicts.

lachlan-00:
> 
> 
> Another thing, I've been pulling changes from ampache/ampache. I don't understand why I've been getting merge conflicts and apparently some files not being updated. I haven't made any changes in the develop branch. So there shouldn't be any conflicts.

yeah it seems really odd, i'll be able to fix it up tomorrow, i'm just putting kids to be now

lachlan-00:
your update seems to have fixed it, i'm rebasing now.

lachlan-00:
okay @wagnered this is at parity now. looking much better, i'll check it out tomorrow. :)

wagnered:
Found a bug already.

I assumed that the first item in the object would be the info on the artist being searched for.  Evidently that is not always the case.

The search uses both artist and album name so there shouldn't be a large number of entries.

Please wait for my ok to merge

lachlan-00:
removing echonest breaks the site as it's in about 10 other places. It's only for the song previews i think i'm going to put it back in the composer files for now and remove it after this.

got artist search working well but the album search will only return the first item. i'll look at this a bit more

wagnered:
Ok I think it's ready.  It found images for artists from the 50's and 60's rockabilly and dowop era.  Many artists I haven't heard of

lachlan-00:
i agree, it's done a good job so i'll wait for travis ad we're good

lachlan-00:
image
here's one issue i didn't see from scrutinizer

wagnered:
What found that conflict? And why isn't PHP complaining?

lachlan-00:
aliases and it's functioning as needed

lachlan-00:
just an issue in scrutinizer, i aliased it to avoid any confusion.

it didn't seem to cause a conflict but there was an Ampache::Session and a Spotify::Session in the art class.

I think if a situation arose where we needed to call new Ampache:Session it would have been overwritten by spotify. No issues with the operation though.

https://scrutinizer-ci.com/g/ampache/ampache/inspections

wagnered:
About echonest, I don't understand how it is working.  I tried pinging traceroute and couldn't get any response. Also tried the developer URL in the web browser and got a time-out waiting for a response.  Their website  says "forbidden".  When I tried to go to their rss sigup page it gave me 404. 

Also from wikipefia:
`
The Echo Nest is a music intelligence and data platform for developers ... The API was shut down on May 31, 2016 and developers were encouraged to use the Spotify API instead. The Echo Nest ...
`
The above is why I was convinced that the echonest package could be removed.

So you are sure it still works?

lachlan-00:
i've never used it but i'll get working on the removal now.

Issue #2446, opened 2020-07-10
UPnP localplay sessions are broken


ahoy:
#  Description

### Describe the bug

In the UPnP localplay code, sessions are used to store information about the players and playlists, and probably other things. The code creates an `api` session, but uses an `$sid` variable to set the key. It appears many years ago a `stream` session was added, and the metadata changed. The session type should be `stream` to be able to set the key with `$sid`.

In the log below, I added some debugging info. Notice that the session key is an md5sum, and not the string that starts with upnp_pls... On the last line, the session is accessed via the string again, but it is not found.
```
...
(session.class) -> Unable to read session from key upnp_pls_http://10.0.3.112:1400/xml/device_description.xml no data found 
(session.class) -> Session create: data=array ( 
(session.class) ->   'type' => 'api', 
(session.class) ->   'sid' => 'upnp_pls_http://10.0.3.112:1400/xml/device_description.xml', 
(session.class) ->   'value' => '{"upnp_playlist":[],"upnp_current":0}', 
(session.class) -> ) 
(session.class) -> Session created: 912e00a0101a01c0e8ab6088588a1532 
(upnpdevice) -> readDescriptionUrl: http://10.0.3.112:1400/xml/device_description.xml 
(session.class) -> Unable to read session from key upnp_dev_http://10.0.3.112:1400/xml/device_description.xml no data found 
...
```

### To reproduce

1. Add a UPnP localplay client
2. Click on "Show playlist"
3. Check debug logs

### Expected behavior
Sessions should be created, and then be retrievable.

## Environment

* Ampache version: develop
* Web server + version: Apache 2.4.43
* Server operating system: Freebsd 11.3-RELEASE-p3
* Client operating system: Ubuntu/UPnP player device

## Settings

If you need settings, let me know. This bug shouldn't depend on settings.

lachlan-00:
can you tell me what you use as a client so i can set it up? i've personally never used this (only mpd)

ahoy:
Unfortunately I was using my Yamaha receiver, since it was what I'd like to send music to. I'm not sure about something that would run on a computer, as I've never tried it. 

I'm setting up my own fork to try to offer some patches, but I'm new to working with github.

I'm not sure about the right way to use Session, so this is what I did to get this working:
https://github.com/ahoy/ampache/commit/f427cde32301aaeba4b75480eb51950b7b47fbc0

lachlan-00:
[...]

PR #2451, opened 2020-07-15
Update session code to use 'stream' instead of 'api'

ahoy:
The UPnP localplay code expects to retrieve session information using named keys.
The Session::create and Session::exists functions now use 'stream',
which allows the $sid to be set.
Should address #2446.

### DIFF
--- a/modules/localplay/upnp/UPnPDevice.php
+++ b/modules/localplay/upnp/UPnPDevice.php
@@ -73,7 +73,7 @@ class UPnPDevice
         $this->_settings['descriptionURL'] = $descriptionUrl;
 
         Session::create(array(
-            'type' => 'api',
+            'type' => 'stream',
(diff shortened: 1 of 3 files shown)


lachlan-00:
Thanks for taking the time to make this. I always love making new contributors!

As I said before I think that this is all correct (and also works for you) so I'll try and test it out asap.

lachlan-00:
I installed bubbleupnp to see what i had on my network but i found my kodi already has upnp active.

So i put your change in and connected to my ip http://192.168.x.x:1221 as a upnp instance.

all working no problem!

ahoy:
No problem, glad to help!

lachlan-00:
welcome to the contributor club!

if you have anything to add to the Upnp wiki page (which is a few minutes old) please feel free to add to it

https://github.com/ampache/ampache/wiki/Upnp

thanks again!

PR #2452, opened 2020-07-15
Replaed deprecated license desigator AGPLv3 with appoved designator 'AGPL-3-or-later'

wagnered:


wagnered:
Got a problem here. I ran php-cs-fixer on my local and found no problems.

So I'm guessing my .php_cs file needs updating.  But again this begs the question:  I just updated from ampache/Ampache last eve, why wasn't it updated?

I want to suggest that a version number be added to .php_cs.

lachlan-00:
i don't know why name and type are indented further in but the foreach in api needs a space.
image

i think you might need a hard reset on your develop branch after this. it's been really odd compared to develop

lachlan-00:
i did a rebase so that should fix it but locally you'll need to reset to make any more changes

wagnered:
I've never needed to do a hard reset. What commit do I do it to?  Do I do it before or after I do another pull?

wagnered:
Never mind,  it was a mess.  I decided to delete my local and fork then refork and reclone.

Shoud be all set now.  Thanks for your assistance.

PR #2470, opened 2020-07-27
DO NOT MERGE, using pull as a way to identify new issues between 4.1.0 and 4.2.0 in scrutinizer

lachlan-00:
there are a few left over new issues and i need to identify these before merging into master


lachlan-00:
issues now well below master.

PR #2478, opened 2020-08-02
change http://www.gnu.org and http://fsf.org to https

ibmibmibm:


wagnered:
I hope you used `sed` and `find` to change all those files: Takes less than 2 minutes to edit all of them. 🙂

Thanks

lachlan-00:
just checked the text and it's all https

ibmibmibm:
> 
> 
> I hope you used `sed` and `find` to change all those files: Takes less than 2 minutes to edit all of them. 🙂
> 
> Thanks

Yes, I was using `find` and `sed` 👍

PR #2482, opened 2020-08-05
Improve performance by loading .mo gettext files rather than .po

cquike:
When loading every single page the localized .po files are loaded. However I managed to improve the performance by loading the .mo files, which are machine readable and therefore optimized for machine parsing, rather than the .po files which are meant for human readability.
Using a Raspberry Pi 2 as an ampache server, the load times of simple pages were before in the order of ~0.9s (as reported in the page footer). After this change, load times have been reduced by ~0.4s on average, which is a big boost. The test has been done with es_ES, de_DE and ja_JP localizations, all showing similar performance improvements.

### DIFF
--- a/lib/i18n.php
+++ b/lib/i18n.php
@@ -31,11 +31,11 @@ use Gettext\Translator;
 function load_gettext()
 {
     $lang   = AmpConfig::get('lang');
-    $popath = AmpConfig::get('prefix') . '/locale/' . $lang . '/LC_MESSAGES/messages.po';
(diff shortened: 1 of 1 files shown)


cquike:
Of course, for this to work I have assumed that the .mo are kept up to date with respect to the latest .po files, which I guess is taken care of when importing from transifex.

lachlan-00:
@Psy-Virus would probably know that?

Psy-Virus:
Makes totally sense. The mo-files are always updated but have never been used, cool 😄. I never thought about looking for this...

Good Job

@lachlan-00, can be merged if you don't plan anything else with it.

:)

Issue #2455, opened 2020-07-17
Inconsistency between "Recent played" in the web interface and the RSS feed


cquike:
## Description

 The front page shows a "Recent played" list which contains the tracks recently played by the user and users who want to share their recently played information (setting 91 in preference table). If RSS is enable in the global configuration an RSS icon appears nearby.
 If the user doesn't want to share the recently played information, the RSS feed (which is not authenticated) doesn't contain the tracks from the user, even if the RSS link was retrieved from an icon heading the list of user played tracks. This is quite confusing for the user, who would expect the RSS feed to contain exactly the content of the recently played list.

### To reproduce

1. Have a ampache server configured with at least two users (i.e. user1, user2) 
2. Configure user1 not to "Share Recently Played information" while user2 does share.
3. Play some tracks with both users.
4. As user1, in the main ampache page will appear a list of recently played tracks which contains both the tracks played by user1 and user2. Close to the "Recently Played" list tittle an RSS icon appears with a RSS link.
5. Add the RSS to a RSS viewer
6. The RSS viewer shows the tracks only from user2, while the link was obtained while logged as user1.

### Expected behavior

 I am not sure what the expected behaviour should be. On one side I would expect the RSS feed to contain the list of tracks played by user1. But that's obviously in conflict with the user1 setting "Share Recently Played information" to false, since the RSS feed is not authenticated.

Possible options that I see:

1. For those users the RSS feed should be authenticated.
2. For the users create a feed which includes a random token in the URL specific to that user. That avoids having an authenticated but anyone who has the link can access the user information.
3. Do not show the RSS feed for those users.
4. Explain close to the icon that the feed contains only tracks from users who want to share their recently played tracks.
5. Do nothing and leave the users baffled :)
 
## Environment

* Ampache version: 4.1.1
* Web server + version: Apache 2.4.38
* Server operating system: Debian 10
* Client operating system: Any

### Client type

* Web
* Akregator RSS client

## Settings

 use_rss = "true"

lachlan-00:
the index and rss both use Song::get_recently_played() without a user and that will grab everyone. So what you see in the RSS should match the ui.

But what i do see is that it looks like the web-ui will take into account permissions and will only allow site admins to view the info or people who match the current user.
```
$is_allowed        = Access::check('interface', '100') || Core::get_global('user')->id == $row_user->id;
```

Seems like the Ampache_RSS class doesn't do a lot of this at all. it also looks pretty malformed in the raw data so will need to work on this.

[...]

Issue #2484, opened 2020-08-06
API key can only be reset by administrator but the icon to reset appears in the user settings


cquike:
## Description
 
 According to the API documentation (https://github.com/ampache/ampache/wiki/Ampache-API) only the administrator can generate (or reset) the API key. However, when a normal user is logged in and goes to the Account preferences, a reset icon appears near to the API Key label (see screenshot). However clicking on it leads to an empty page, presumably because the user doesn't have the permission to do it.

### To reproduce

1. As a normal user go to Preferences - > Account.
2. Click on the icon right to "API Key" which is labeled as "Generate a new API key"
3. An empty page is shown.

### Expected behavior
 
 I see two possible solutions:
 1. The "Generate a new API key" is not shown when the user is logged in.
 2. The user is allowed to change the API key. I am not sure why this is not allowed. A user is actually allowed to change the password, so why not the API key? Which is the rationale behind?

### Screenshots

![api_key_reset](https://user-images.githubusercontent.com/17937361/89518190-da16ef80-d7da-11ea-8138-ff1dd8909e5f.png)

## Environment

* Ampache version: 4.1.2
* Web server + version: apache
* Server operating system: Raspbian
* Client operating system: Firefox on Debian 10

lachlan-00:
I saw that today. (Too late for changes on 4.2.0)

What I think we should do is make it a permission so it only shows up if you assign permission to that level.

Ie allow content managers to reset API key and then only show to users content manager+ and admins

cquike:
I see that  in the latest version 5.6.0 this is not an issue anymore, so I will close the it.

PR #2486, opened 2020-08-06
Recently played RSS feeds fixed for users who don't share played info

cquike:
This PR  will fix issue #2455 by adding a unique token to the RSS URL. The RSS token allows identifying which user has asked for the recently played history.
The original ticket showed the problem for users who have their privacy setting to disable, however I see no reason why this is not applied in general. That way, for users who decide to turn their privacy setting off the feeds generated with the RSS token would still work after the user changes the settings.
  
The RSS token can be reset by the administrator in a similar fashion to the API key (but see issue #2484), which will modify the behavior of existing feed URLS, since now ampache cannot identify the user. Still, the feed will work as before fixing this problem, i.e., by showing all public recently played items but not the ones from the user. 

The database schema has changed to add the new rss token column. The updater takes care of the change and modifies the user table accordingly.

Additionally, a different bug has been fixed: previously the body of the feed contained a " time ago" text that informed when the item was played. However, some RSS clients get confused by body changes and interpret this as a new feed item every time they update the feed, showing a very confusing feed timeline with repeated items. Anyway this information was redundant, since it is just a fancy way to show the time, that is also part of the feed item that RSS clients usually show.


lachlan-00:
That's actually a really intelligent way of handling this. Looking forward to checking this one out

lachlan-00:
running up now should be merging soon

lachlan-00:
Done, love it! merging after travis

PR #2487, opened 2020-08-06
Fix non-locallized subcategories in preferences

cquike:
The preferences page shows a subcategory that until now has not been localized. The problem is the string was passed to ucwords _before_ being localized, and since that's the msgid, then it couldn't be found in the catalogue. Now ucwords is applied to the localized string instead.


Psy-Virus:
Makes sense too.

Why ever we need preferences to be in uppercase. But the Author probably thought something by doing so...

:)

lachlan-00:
now that's a good catch

PR #2488, opened 2020-08-07
Added translation of settings that cannot be changed by user

cquike:
The preference settings that the user doesn't have the permissions to change had a hardcoded string ("Disabled", "Enabled) so no translation is shown in the UI.
This PR fixes that, by adding the corresponding T_().
I have executed gather-messages.sh -g to generate a new messages.pot which I have included in the PR. However it is not clear to me whether the .po and .mo files need to be regenerated (with gather-messages.sh -ma I guess) or whether this is done automatically by transifex once a new messages.pot is there. If the former, I can easily run the command and remove the obsolete strings (there are two unrelated obsolete strings) to this PR.

### DIFF
--- a/lib/preferences.php
+++ b/lib/preferences.php
@@ -137,9 +137,9 @@ function create_preference_input($name, $value)
 {
     if (!Preference::has_access($name)) {
         if ($value == '1') {
-            echo "Enabled";
+            echo T_("Enabled");
         } elseif ($value == '0') {
-            echo "Disabled";
+            echo T_("Disabled");
         } else {
             if (preg_match('/_pass$/', $name) || preg_match('/_api_key$/', $name)) {
                 echo "******";


lachlan-00:
take out the change in locale/base/messages.pot and let @Psy-Virus handle that.

also in an aside i think we take out gather-messages now that we use transifex completely. I did the same thing when i started by updating the pots but it affects syncing between transifex

cquike:
I have taken messages.pot out of the commit.
Thank you for your support!

lachlan-00:
It's the weekend so I'm not getting to this for a bit but I want to really heap some praise on you for finding some really good changes that hadn't even registered for me.

They might not feel like much but I really appreciate every single pull and comment from you and everyone else here.

Having so many eyes and opinions on the code makes my life easier and Ampache better. :)

cquike:
Thanks so much for the praise. That's much appreciated! 
Ampache is a super useful software and It is a pleasure to contribute to its development. And the team is great, which makes it fun!
This week I had some family break and I could contribute indeed more. Unfortunately my contribs will be sparse in time, but I still hope to continue "regularly".
Thanks!

Psy-Virus:
@cquike,
> @lachlan-00: "take out the change in locale/base/messages.pot and let @Psy-Virus handle that."

What @lachlan-00 said! 😜 

@lachlan-00,
> also in an aside i think we take out gather-messages now that we use transifex completely.

The gather-messages script does the merge-, clean-, database- and what not update tasks for me. It's not perfect but does it's job. Don't discriminate it, it has feelings too 😢 🤣 

As of now, everything in the locale directory has a regular use. And I will care about everything that has to do with it, if needed.... Except the copyright stuff 😄 
And if there's more time to dig into it again, I will update the tool and make it more user friendly. But this will probably be the time when I can't work on Ampache any more. So pray that this day will never come 😛

Psy-Virus:
Oh, before I forget it, a big thanks for the commits, @cquike!

:)

lachlan-00:
sorry for lagging on this one. i'll get onto the rss one next

PR #2521, opened 2020-09-03
Fixed inconsistency in the display of Tags (Genre).

cquike:
Most of the UI uses the term "Tags" to show the genres of the songs (for instance albums.php?action=show&album=????).
However, when displaying the song details (as in song.php?action=show_song&song_id=????), the tags were shown under
the term "Genre".
Additionally, also editing a channel details show the term Tags rather than Genre.

There are two additional places where the term Genre is still used: in the catalogues and broadcasts properties. In the first one it appears as part of the file name to parse. Since the filename is an "external interface" I have left it as is. I have no experience with broadcasts at all, so I don't know whether it makes sense to also change it there or not.


lachlan-00:
@ampache/project-team @wagnered @Psy-Virus  it's time! do we want to keep using tag or do we make the change into genre?

I'm used to tag but it's probably the weirdest holdover from old Ampache and maybe the language isn't really correct? I understand the usage of tag and why it's there but i can also argue against it.

Seeing as this is happening during the 5.0.0 window it's probably the best time to make a call on it.

wagnered:
@Actually, **genre**, or maybe **content** as defined by id3.org  would be the proper nomenclature.  **Tag** is the generic term for all the metadata in a file, music or otherwise.  Audio/Multimedia files aren't the only ones that might contain metadata (tags).

There are sections of the code that I'm still not sure what **tags** mean, especially redoing the writing all tags to the file.  The distinction is definitely blurred.

kuzi-moto:
Seems like part of the problem is that often music can be 'tagged' with a genre such as on [Genius](https://genius.com/Genius-tags-music-genres-international-annotated), or [Soundcloud](https://help.soundcloud.com/hc/en-us/articles/115003446007-Adding-a-genre-and-tags-when-uploading-a-playlist). Then, as @wagnered mentioned, tags also refer to the metadata for many different types of media.

Since Ampache is often using both id3 tags, as well as genre tags, that when referring to genre we use either "genre" or "genre tags", and just "tags" should specifically refer to the metadata.

cquike:
Thank you for the discussion! It looks like I understood it wrongly, thinking that Genre was the "old" term and Tags the "new".
Anyway, it seems that there is an inconsistency that needs to be solved in one way or another. I should have probably created an issue rather than a PR, sorry for that.
And it looks like the issue is not as simple as naming it Tag or Genre. I guess the question is: should Ampache support tags which are not genre tags?

wagnered:
>  I guess the question is: should Ampache support tags which are not genre tags?

Not sure I understand the question.  The word "tag" is just a vague term for an individual piece of metadata(album, artist, etc). We can stop referring to those individual pieces of metadata as tags and  the word "tag" should be stopped  being used as a synonym for "genre".  That would clear up any confusion I have about it.

Sorry, about the proselytizing.  I'm old  and set in my ways😇

cquike:
Thank you for clarifying. That helps, since I have almost no experience with id3.
Your proposal makes sense to me. It would mean showing always "Genre" rather than tags in the UI. And I guess that if it is done properly also the database tables (tag, tag_map, tag_merge) and the naming of the code (templates/*tag*, lib/class/*tag*,...)  should be changed, right?

wagnered:
I would say yes. Showing Genre on the ui would be consistent with most other apps.  Changing db tables to genre, genre_map, and genre_merge should only furter clarify the intent of the code. As would renaming in the files.

lachlan-00:
Altering the database tables will definitely have to happen as well for sure. Il do that outside of the strings

wagnered:
Wow, that's the first and only time I've seen genre and style delineated like that.  

> Style is also relevant here as people tend to talk about genre when they mean style.

No wonder, are there even any taggers that delineate the two?

> So far I'm thinking artist/album/song/radio/API data is covered by genre.

I don't think they define content type.  Artist/album can be part of the metadata of course.  

Radio and API in my mind are transient "vehicles" of media transport.

lachlan-00:
I think style is used after genre was defined by id3 with a specific list so instead of cluttering genre, style is like a subset of genre

Psy-Virus:
Well, tag is not equal to ID3-Tag.
ID3-Tag is just the name of the tag-data collection used in "modern" audio files like APE and Vorbis-Tag or FLAC-Tag and should not be discussed here.
Genre is the right terminus for the specific kind of music, like rock, hard rock, rock ballad, metal, death metal; trance, techno, hard techno, house, hard-style, jump-style, dubstep and so on.

"Style" is more like a collection of genres. For example, if music is based on an e-guitar, drums, e-bass, one or more singers we call it rock, no matter if it's hard rock or a rock/metal ballad. Even metal is often considered as rock or vice versa.
The style of techno, house, trance and so on would be electro / electronic music. Personally, I also would describe the most of the nowadays produced music as electronic music as electronic beats and synthesized tones/melodies dominate it.

The terminus "tag" is more like a definition of what describes a song. Music can be funny, melancholic, happy, dark and what not. A "tag" can also be used to make collections of different styles into groups, like "Film Score", "Disco Music", "Birthday party of 2020".

I think we should make it so that a user can tag a song how he/she/it (should be politically correct, I guess :D) wants it to be tagged. That would require a new table in the DB where we have the "tag" and the associated song-IDs for this tag as fields.
But there we have to be careful on the performance impact as there can be thousands of song IDs per tag. The only ways, I know of saving an array to a database field is by serializing an array, by json_encode or a simple sign separated list (1,34,94,[...]) via implode and explode functions. My researches have shown that json_encode and json_decode is the fastest and we can use it right away to put the heavy task of building the tables and the html on the client side without former conversion through PHP on the server side.
The Tag-Cloud should be renamed into Genre-Cloud as what's listed there are only genres of the (ID3/Vorbis/FLAC)-Tag-Field "Genre" if I see it right.

Let me read your thoughts

:)

wagnered:
> Music can be funny, melancholic, happy, dark and what not. 

And in id3v2.4 these fit into the TMOO frame(mood).  Yet another IMHO valid classification.

> I think we should make it so that a user can tag a song how he/she/it (should be politically correct, I guess :D) wants it to be tagged. 

An admin can pretty much do that now, except for adding additional metadata (I think I'm getting to hate the word "tag"☹️).  But to paraphrase what someone stated earlier: Ampache ain't meant to be a tagger app. The "write id3 to file" feature never did work correctly except on a few most basic metadata, and I'm finding it tough trying to find the most efficient means to rewrite the metadata that is never imported into the db, but should be included when updating the file, but that's a different story.

> The Tag-Cloud should be renamed into Genre-Cloud as what's listed there are only genres of the (ID3/Vorbis/FLAC)-Tag-Field "Genre" if I see it right.

I concur, although I just assumed it would be.

Psy-Virus:
> And in id3v2.4 these fit into the TMOO frame(mood). Yet another IMHO valid classification.

> An admin can pretty much do that now, except for adding additional metadata (I think I'm getting to hate the word "tag"☹️). But to paraphrase what someone stated earlier: Ampache ain't meant to be a tagger app. The "write id3 to file" feature never did work correctly except on a few most basic metadata, and I'm finding it tough trying to find the most efficient means to rewrite the metadata that is never imported into the db, but should be included when updating the file, but that's a different story.

I didn't mean to write anything into the File-"tags" but just into the database, as I'm not a fan of doing such work on Ampache too. For tagging I use MP3Tag, what's much faster and suitable for those operations. We should not try to reinvent the wheel if there are tools that do this job already perfectly fine.

lachlan-00:
I actually really like the idea of the tag cloud becoming more like a mood board. it needs a big rework though and outside of this you've actually given me some good ideas on what to do with it in the future.

"i'm feeling Dark", "i want to party" That sort of thing would make discovery/browsing a mood cloud a lot easier.

But back to this, it seems like we all agree on Genre.

I'm going to update the pull with a few extra string changes i found and we'll get it merged.

lachlan-00:
okay! Genre it is, it's time i started updating the API functions so this will be the beginning of breaking changes in Develop for the API while we work on Ampache 5

tsointsoin:
Implementing styles would be a very welcomed addition, could it be doable ?

PR #2534, opened 2020-09-12
API fix: playlist_create optional "type" parameter

BLeeEZ:


lib/class/api.class.php
```diff
@@ -1481,7 +1481,7 @@ public static function playlist_songs($input)
      */
     public static function playlist_create($input)
     {
-        if (!self::check_parameter($input, array('name', 'type'), 'playlist_create')) {
```
usox, line 1484: I guess, this is ok because type is optional according to the documentation.

usox: But: If the type-Key is not defined in the input-array, Line 1488 will raise a warning because of the missing key.

Possible solution:
```
$type = $input['type'] ?? 'private';
```

usox (changes requested):

usox (approved):

lib/class/api.class.php
lachlan-00: use isset and that's fine with me
$type = (isset($input['type'])) ? $input['type'] : 'private';

lachlan-00:
thanks for making the changes. it's really just to keep the style of those ternery operators the same and keep tests inside brackets

lib/class/api.class.php
usox: @lachlan-00 just fyi, `??` is the null coalescing operator (>= php7) which checks if an expression evaluates to null and not the short form of the ternary operator (`?:`).

lachlan-00: just keeping it consistent ;)

PR #2536, opened 2020-09-14
Update json_data.class.php

AshotN:
Why is `id` begin cast to string when it's number? 

Fix the include not working in Album.

There are some type errors that still need to be resolved.

### DIFF
--- a/lib/class/json_data.class.php
+++ b/lib/class/json_data.class.php
@@ -150,7 +150,7 @@ class JSON_Data
                     ));
                 } else {
                     array_push($JSON, array(
-                        "id" => (string) $id,
+                        "id" => (int) $id,
                         "name" => $data['name']
                     ));
                 }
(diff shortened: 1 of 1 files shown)


usox:
Well, using `string` as type for ids is not a bad idea (at least as parameter type for external communication). Changing those parameters to `int` will remove some flexibility from the api (maybe, in future someone comes up with the idea to use uuids as id) - just sayin'.

lachlan-00:
XML requires the id to be strings to generate valid xml. Because of that i've made it a rule to do all object ids as a string and i'd like to keep JSON the same as XML.

We also have one string type; Smart playlists in the api are sent with a "smart_" prefix which would then be different if we went to ints.

lachlan-00:
I also just realised we don't have a users api call to get a list of users. that seems a bit of an oversight

AshotN:
I can revert the string->int casting then. But should JSON be held back due to XML.

AshotN:
I've made the requested changes, if you could review. If everything is alright, please squash and merge :)

PR #2538, opened 2020-09-15
Gather spotify improvement

wagnered:
Finally!


wagnered:
Hold up on the merge!  There are some Codacy issues I want to take care of.

wagnered:
Codacy did find a problem with the default css file.  I will check any Codacy messages I get to see if it's fixed.  

Codacy finds only one css problem at a time.  So far I had to do three  separate fixes/pushes and got another css error.

Some other complaints about direct use of  _REQUEST without stripping slashes is bogus if just using `isset`.

As Codacy is usually ignored, I looked for a command line CSS validator and found a number of online services, but I didn't think Ampache should go that way.  I found one cli app so far: a python module.  It's a wrapper for cURL to send the file to  the W3C CSS Validation Service.  I'll check it out.

lachlan-00:
i was looking at the settings and is seems like there are a few duplicates from the $_GLOBAL checks.

.gitignore
```diff
@@ -46,3 +46,4 @@ phpcs*
 *.v11.suo
 *~
 composer.phar
+spotify-test
```
usox, line 49: Is this necessary?

lachlan-00:
haven't had time to go test but just updated the branch and indentation/spacing in a few pages.

is there more you want to do?

wagnered:
well is other way to keep my own test files from being nagged by git?

wagnered:
I haven't finished checking Codacy.  Anyway what's your hurry?

.gitignore
wagnered: I think so, or I wouldn't have added it.  If you think I should polish it and add it to git history, given time, that is a possibility.

Also, isn't that the purpose of ,gitignore?

usox: > Also, isn't that the purpose of ,gitignore?

Just curious. The .gitignore file of a project should contain just project-specific entries - mostly generated stuff like composer.lock, the vendor-directories or user-specific versions of certain files (e.g. a phpunit.xml or the ampache-config file). The goal is to keep the .gitignore as small/clean as possible.

Temporary/test files should be treated like custom editor settings/artefacts (like .idea, vim swap files, ...) and should be placed in the users global .gitignore which has to be configured in the users global .gitconfig.

Examplewise in linux:
```
/home/<username>/.gitconfig
....
[core]
    excludesfile = /home/<username>/.gitignore
....
```

wagnered: >  The .gitignore file of a project should contain just project-specific entries -

And that's what I consider spotify-test.

> ...should be placed in the users global .gitignore which has to be configured in the users global .gitconfig.

Point well taken will make the change. And you should have led the conversation with that instead of the authoritarian attitude.

wagnered:
> haven't had time to go test but just updated the branch and indentation/spacing in a few pages.

Thanks, but Isn't php-cs-fixer capable of fixing those indents?

wagnered:
Ok, it looks like a "go" now.  A number of Codacy complaints might be false:. I think `scrub_in` sanitize s input.

.gitignore
usox: > And you should have led the conversation with that instead of the authoritarian attitude.

If I sounded authoritarian, I apologize.
I'm used to ask straight questions instead of playing the `know-it-all` and telling everyone, what to do. Many people (including me) tend to forget removing temporary entries in .gitignore. So my intention was just to point it out (in my clumsy german way).

lib/class/art.class.php
```diff
@@ -1103,8 +1103,11 @@ public function gather($options = array(), $limit = 0)
 
             return array();
         }
-        $config  = AmpConfig::get('art_order');
-        $methods = get_class_methods('Art');
+        if ($limit == 0) {
+            $limit   = (is_null(AmpConfig::get('art_search_limit'))) ? 5 : AmpConfig::get('art_search_limit') ;
```
usox, line 1107: Wouldn't it be a idea to define a public class constant for the default value (5) instead of having this `magic int` in several places? Any later change can simply be done by changing the constant.

config/ampache.cfg.php.dist
```diff
@@ -597,6 +597,18 @@ sociable = "true"
...
+;  Narrow the search. 
+;  POSSIBLE VALUES:  artist,(year:1991 or year:1991-2000)
+; POSSIBLE  VALUES: empty string("") or commented out for no filter
+; DEFAULT: none;
+;spotify_art_filter="artist"
+
+; Art search limit
+;  Limit the total images returned
+;  DEFAULT: 5
+art_search_limit="15"
```
usox, line 610: Was it your intention to set the default in the dist-config to 15? I guess, there's a `;` missing? Or did I get this wrong?

wagnered: Thanks, no, it wasn't. When I copied it over to the .dist file, I forgot to comment it.  Thanks for the heads up.  Taking care of it now.

lib/class/art.class.php
wagnered: I may be wrong, but I thought that access to the various gather functions is directed through ` gather($options = array(), $limit = 0)` thus the  "magic" number is not actually needed for the default parameter value in the other functions.  If not, perhaps all calls **should** be directed through this "gateway".

But then the functions called on by gather  would need to be private to prevent bypassing by external functions.  Would the constant then be advantageous?

usox: You're right with your assumption, that it `should` be that way. If someone re-uses the code in future, this may change (maybe by some classic copy-paste).

As long as those numbers represent the default value, a constant is imo very advantageous. Tomorrow, some one else needs to define this art-search limit in another place - that would be the third occurrence (after lines 1107 and 1196).

If one then decides to change the default value, one has to _know_ these 3 occurrences and has to change them by hand. If one has a constant, one can change it with a single line of code.

Just an idea:
In an ideal world, every default value from the config would have a constant and all those constants would be placed in the same place (an `Enumeration`, but php does not support real enum datatypes :( ).

It would look as follows:
```
final class ConfigDefaults {
  public const ART_SEARCH_LIMIT = 5;
  public const SPOTIFY_ART_FILTER_LIST = "artist";
...
}
```
Some time, the default config file could be generated using these defaults, too.

wagnered: There are more $limit=5 in the other class functions than those two.

I did a grep andl found only `gather($options = array(), $limit = 0)` called externally.  The defaults for all the other functions called by this function should not need a default value for $limit.

But I like the idea of descriptive class constants.  Perhaps this could be added to the wiki coding standards page.

wagnered:
Hold up again. Something isn't working right.

lachlan-00:
take your time, i've got a lot of day-job things to do this week so will be a few days before i can test

lib/class/art.class.php
```diff
@@ -1103,8 +1108,11 @@ public function gather($options = array(), $limit = 0)
 
             return array();
         }
-        $config  = AmpConfig::get('art_order');
-        $methods = get_class_methods('Art');
+        if ($limit == 0) {
+            $limit   = (is_null(AmpConfig::get('art_search_limit'))) ? ART_SEARCH_LIMIT : AmpConfig::get('art_search_limit') ;
```
usox, line 1112: @wagnered Sorry to bug you again, but using `ART_SEARCH_LIMIT` won't work as it's a class constant and not a global one. :(

self::ART_SEARCH_LIMIT or static::ART_SEARCH_LIMIT will do the trick (I personally prefer `static::` because of late static binding https://www.php.net/manual/de/language.oop5.late-static-bindings.php)

wagnered:
as `public function gather($options = array(), $limit = 0)` is the gateway for all art searches, I  don't see why it matters.  I would prefer that all functions called by this function be private, ensuring that this function remains the gateway as it seems to have been previously.  As far as I can tell, the art class is the only class to search for art and process the  results.

Could you give me an example of where else in the code `art_search_limit` might be functionally and semantically  used?

usox:
> as `public function gather($options = array(), $limit = 0)` is the gateway for all art searches, I don't see why it matters. I would prefer that all functions called by this function be private, ensuring that this function remains the gateway as it seems to have been previously. As far as I can tell, the art class is the only class to search for art and process the results.

I fully agree with you - those methods should be private. I would go further and move the complete art-search code into a single module having a single entry point (the public gather function) to encapsulate the logic. But this is future work.

> Could you give me an example of where else in the code `art_search_limit` might be functionally and semantically used?

No, because I would have to search for `5` in the complete source code - maybe, it's hidden somewhere :) Just kidding.

There have been a lots of discussions about magic numbers/string in the past decades and I just want to refer to the existing  wikipedia article regarding this issue:  https://en.wikipedia.org/wiki/Magic_number_(programming)

wagnered:
> No, because I would have to search for 5 in the complete source code - maybe...

Maybe you'll not find it  because  the default limit doesn't need to be passed in.😇  

>  I would go further and move the complete art-search code into a single module having a single entry point (the public gather function) to encapsulate the logic. But this is future work.

That should take care of Codacy's complaint about too many methods.  Sadly it has the same complaint about gather_spotify which I am about to attack.

I would like to take that on, also I would like to take the first step by privatizing the called functions while working on the gather_spotify complexity issue.

I won't wait for the PR to discuss: I'll start a discussion in the [developer team](https://github.com/orgs/ampache/teams/developers/members) discussion group.

@lachlan-00,  please add @usox to the roster.

usox:
@wagnered just fyi, I wrote a quick proof-of-concept regarding splitting up this `gather-art` methods. I you're interested to take a look, you'll find the code in here
https://github.com/usox/ampache/tree/split_gather_art/src/Module/Art/Collector

wagnered:
Looks like you are implying  that the refactoring wait until a future version

I will be modifying the get_art dialog to allow interactive setting of the search limit and html validation.

lachlan-00:
added

lachlan-00:
what are the possible filters in the config? just year and artist? if there's a link to the api we could add a bit more info about that.
```
 POSSIBLE VALUES:  artist,(year:1991 or year:1991-2000)
```

wagnered:
Those are the only two I've incorporated into the code.  Spotify also provides a filter for:  "only albums released in the last two weeks," and "lowest 10% popularity,"

[Spotify developer API](https://developer.spotify.com/documentation/web-api/reference/search/search/)

lachlan-00:
cool, good to know, i'm going to add this one now and have a run

lachlan-00:
could we filter by album year automatically too in the config? like year:ALBUM

and when we put artist,etc into the config that says search album and artist right?

wagnered:
The config switch `spotify_art_filter`  is ignored when searching via the dialog, but not when "Gather Art* is applied.

Let's just say, that when the artist filter checkbox is checked, the artist input field should not be empty.  Let's also say that validation check still needs to be applied.  Simple HTML5 validation won't do it.  I'll have to add JavaScript.

First, my develop branch developed a problem after checking out and trying source-changes.  I get the error: "[ampache] (core.class) -> 'Gettext\Translations' not found!".  I had php-gettext included when I built php7.4.10. Oh well!

lachlan-00:
got a lot of work on but i'll get there!

lachlan-00:
okay @wagnered it's all working well and i put the translations in for the new text and set a limit on the art page (1-50).

Because i added your other css pull first it seems like this one is deleting a lot of that in default.css? should i restore the deleted rules in this pull as well?

wagnered (commented):
I'm still working on the validation logic.

I suggest you hold off or I'll end up overwriting your changes or possibly merge conflicts. " It's not over until the fat lady sings".

Or comment on the change(s) you want and I'll add them.

lachlan-00:
no problem, i'm happy with the actual process and it looks a lot better now so it'll be easy to merge in then when you're done.

wagnered:
I'm pretty sure it's ready for merging now.

lachlan-00:
will checkout the branch and merge in a few hours

AshotN:
Please also squash the commits into one.

lachlan-00:
had to jiggle the css again, i think i took some part out when doing the cel_hover bit.

all good now.

PR #2562, opened 2020-10-02
Source changes

wagnered:
Added null coalescing operator to return value in `public static function get_global($variable)`.


lachlan-00:
your branch @usox so merge when you want

usox:
Hi @wagnered 

Thanks for the fixes. As your PR contains a weird commit (the first one) which seems to contain a merge of all commits since sep/17, I decided to cherry-pick the two other commits directly from your fork to keep the base branch clean. Therefor I close the PR without merging it.

wagnered:
That bothered me,. So how do I update my fork from upstream without getting all those extra merges?  I thought that because those "extra" merges came from Ampache/Ampache, they would be ignored by upstream in follow up PRs.

What is the timeline for this branch to become the default?  There are a bunch more modifications required, centered around updating file tags from the DB. These bugs have always existed. Of course, that could be because there is no policy as to how it should be done.  I wanted to get familiar with the architecture of this branch,  so I started the changes here.  If the timeline for publishing this branch is far in the future, I can adapt these changes to the develop branch and continue the work on  the changes.

usox:
> That bothered me,. So how do I update my fork from upstream without getting all those extra merges? I thought that because those "extra" merges came from Ampache/Ampache, they would be ignored by upstream in follow up PRs.

Well, I strictly follow this howto https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork.
As long as you use "feature branches" (https://martinfowler.com/bliki/FeatureBranch.html) and don't modify the main branch ("source-changes"), you're good.

> What is the timeline for this branch to become the default? There are a bunch more modifications required, centered around updating file tags from the DB. These bugs have always existed. Of course, that could be because there is no policy as to how it should be done. I wanted to get familiar with the architecture of this branch, so I started the changes here. If the timeline for publishing this branch is far in the future, I can adapt these changes to the develop branch and continue the work on the changes.

Well, honestly I don't know - maybe @lachlan-00 has an idea. This branch marks the start of a new base architecture and so, I guess there's still lots of work (at least in the template-section). But I also think, we can't wait forever until "it's done" - so, some work will be postponed until after the release - I already have lots of ideas and you also came up with some good points (e.g. the question, which kind of dbal/orm/.. ampache uses in future, etc..).
Maybe we will continue the "main work" at least until the end of the year and check in January how the release can be planned. But the release should not take place in March - that's when I'm on vacation :smile: 

I also began backporting some changes I made to develop, so you're not alone...

lachlan-00:
Jan is a good length. I'm going to do 4.2.3 tomorrow and update the api pull. Were starting to diverge a lot so it will need to be soon that we decide to make the break.

I think as long as for develop people the change is documented and warning of at least a month/a beta release in dec to give people time to adjust we should be good.

PR #2568, opened 2020-10-05
Replaced '/' with '?' the proper URL query separator as defined by ietf.org

wagnered:


wagnered:
I'm closing this until I can get a better understanding.

PR #2570, opened 2020-10-07
Returned correct functionality to 'Cancel' button of 'show get art' dialog

wagnered:
This works whether AJAX load is enabled or not.

### DIFF
--- a/templates/header.inc.php
+++ b/templates/header.inc.php
@@ -105,34 +105,7 @@ $t_logout    = T_('Log out'); ?>
             var jsSaveTitle = "<?php echo T_('Save') ?>";
             var jsCancelTitle = "<?php echo T_('Cancel') ?>";
         </script>
-
-        <?php
-        if (AmpConfig::get('ajax_load')) {
-            $iframed = true; ?>
-        <script src="<?php echo $web_path; ?>/lib/javascript/dynamicpage.js"></script>
-        <?php require_once AmpConfig::get('prefix') . UI::find_template('show_html5_player_headers.inc.php'); ?>
-        <script>
-            function NavigateTo(url)
-            {
-                window.location.hash = url.substring(jsWebPath.length + 1);
-            }
-
-            function getCurrentPage()
-            {
-                if (window.location.hash.length > 0) {
-                    var wpage = window.location.hash.substring(1);
-                    if (wpage !== 'prettyPhoto') {
-                        return btoa(wpage);
-                    } else {
-                        return "";
-                    }
-                }
-
-                return btoa(window.location.href.substring(jsWebPath.length + 1));
-            }
-        </script>
-        <?php
-        } else { ?>
+ 
         <script>
             function NavigateTo(url)
             {
@@ -144,8 +117,7 @@ $t_logout    = T_('Log out'); ?>
                 return btoa(window.location.href);
             }
         </script>
-        <?php
-        } ?>
+ 
         <script>
             $.widget( "custom.catcomplete", $.ui.autocomplete, {
                 _renderItem: function( ul, item ) {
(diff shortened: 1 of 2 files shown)


lachlan-00:
would this header script be used anywhere else?

wagnered:
I assume you are referring to the script I removed? 

First, whatever exists after a hash mark is **not** sent to the server. It's meant to go to a location on the same page: usually an id.

Second, the removed script was only valid if AJAX  load was enabled. That seemed strange and without any comments.

Third, according to the grep I did on "NavigateTo, there are about 18 found elsewhere, but they are used to form the href part of an anchor and wouldn't use the script I removed.

BUT revisiting the changes, I noticed I had removed  `$iframed`  which appears to be used only by the html5 player logic. So, I'm going to step through the pre-change header code and see what's going on.

Thanks for bringing it to my attention.

lachlan-00:
glad to have you on top of this!

wagnered:
I assume you are referring to the script I removed? 

According to the grep I did on "NavigateTo, there are about 18 found elsewhere, but they are used to form the href part of an anchor and don't need to use the script I removed.

BUT revisiting the changes, I noticed I had removed  `$iframed`  which appears to be used only by the html5 player logic. So, I'm going to step through the pre-change header code and see what's going on.

Thanks for bringing it to my attention.

wagnered:
Unless you or @usox find something else, it should be good to go.

I wanted.to get this out of the way as it's slowing down my work on writing id3 tags.

usox:
@wagnered tbh I don't know whether the removed hash-behaviour if crucial for the gui to work or not. I see the application updating just the hash when I'm using the gui, though. So I'd guess, it is actually necessary for the ajax-mode to work. :shrug:

wagnered:
> @wagnered tbh I don't know whether the removed hash-behaviour if crucial for the gui to work or not. ... So I'd guess, it is actually necessary for the ajax-mode to work

I don't like to guess, so I empirically tested the NavigateTo statement in "show_playlist.inc.php" with `AJAX load` disabled. and it worked correctly. the fact that it works correctly with  `window.location.hash = url.substring(jsWebPath.length + 1);` which in my case is document root,  implies something **undocumented** is happening internally for a full URL assigned to location.hash to work.   And because it's use here contradicts documented intent and purpose, it should not be used in an undocumented manner as it is likely to change. 

And it bothers me(IBM) that Firebug light doesn't have the capability of stepping through JS code: I would like to see exactly what is happening when a full URL is assigned to location.hash. Although I'm not sure it would be helpful in this case.

usox:
If something explodes, we can apply a revert any time - so lets get it merged so @wagnered is no longer blocked :smiley:

wagnered:
Thanks, but I'd rather get ampache's insane coding properly corrected.  I can merge the fix into the current feature branch and.continue my work.

I just encountered another insanity while stepping through the code to write the selected  image.  In the location header being returned is `http://127.0.0.1:4000/albums.php?action=show&album=3#artists.php?action=show&artist=1`.  The hashed fragment is not on the page that the location is being directed to.  Because the correct information is displayed, the browser must be ignoring the hashed part.

wagnered:
Again, my big fingers got in the way. Sorry. But I'm so frustrated by the general attitudes that I was tempted to save it close

wagnered:
Sorry again, little screen, big fingers.  But I'm so frustrated by the general attitudes that I am tempted to keep it closed.

usox:
Yay, this will save me some work in my gui refactoring poc.

wagnered:
I want to apologize for my outburst.  Frustration had been building from many directions:  From having to do all the debugging and development on a six inch screen to a not-so-successful cataract surgery.  They all came together and collapsed on top of me with the inadvertent re-closing of the PR.

Again, I'm sorry.

lachlan-00:
It's not a problem.

PR #2583, opened 2020-10-17
Feature: Import Artist images from file tags.

wagnered:


lachlan-00:
ah cool, this is a pretty good idea. the image 8 is artist and 3 is front/cover?

lachlan-00:
just looked it up here. https://en.wikipedia.org/wiki/ID3#ID3v2

that's going to help a lot of obsessive taggers. i should start putting artist in mine as well.

wagnered:
yea, all that good id3 info for every version can be found at id3.org.

I've been entertaining the idea of importing other picture types such as $13, the band or artist's  logo,  $0F photo from performance.

wagnered:
There is an image "kind" column that is hard-coded to 'defaut' which might be used in some way.  Also, there seems to be plenty room for more images on the artist or album page even on my 6 " display.

lachlan-00:
performance would be good for boots?

I've only ever put front on my files but would the other types be popular?

wagnered:
I believe that a favorite band's logo might be popular.  An audiophile of classic music might have a favorite conductor.

wagnered:
Most taggers offer only covers. Easy tag will mass insert any type of valid  images, but does not search for anything but front covers I think.

Issue #2587, opened 2020-10-23
After upgrade to 4.2.0 all new played songs are not counted


cquike:
## Description

Before ampache was upgraded to 4.2.0 a list of songs, albums or whatever view which showed a song included the proper number of times the song was played. After ampache was upgraded to 4.2.0 all new song reproductions are no longer shown.

### To reproduce

This is what I did

1. Check for the number of times a song has been played
2. Upgrade to 4.2.0
3. Play that song again
4. The number of times the song was played is not increased

## Environment

* Ampache version: 4.2.0
* Web server + version:  apache 
* Server operating system: raspbian
* Client operating system: Linux

### Client type

* Web

cquike:
I think I have discovered why this is happening: after the upgrade I enabled the "Cache computed SQL data" option to try fixing #2555 but I didn't enabled yet the cron execution of bin/cron.inc.
After manual execution of  bin/cron.inc then the song played count appears properly. But *only* for the songs played before the cron execution.

cquike:
I have created a PR #2589 that fixes the problem.

PR #2589, opened 2020-10-23
Fix the played count when the cron_cache option is enabled

cquike:
As reported in #2587, if cron_cache option is enabled, then the number of times an object has been played is not properly shown in the UI until the bin/cron.inc script is executed.
This commit simply uses the object_count database always, as it used to be. It doesn't seem to be any advantage on using the cache_object_count as far as I can see, since it still needs to do a similar DB query.

### DIFF
--- a/lib/class/stats.class.php
+++ b/lib/class/stats.class.php
@@ -209,14 +209,10 @@ class Stats
             $threshold = 0;
         }
 
-        if (AmpConfig::get('cron_cache') && !defined('NO_CRON_CACHE')) {
-            $sql = "SELECT `count` AS `object_cnt` FROM `cache_object_count` WHERE `object_type`= ? AND `object_id` = ? AND `count_type` = ? AND `threshold` = " . $threshold;
-        } else {
-            $sql = "SELECT COUNT(*) AS `object_cnt` FROM `object_count` WHERE `object_type`= ? AND `object_id` = ? AND `count_type` = ?";
-            if ($threshold > 0) {
-                $date = time() - (86400 * (int) $threshold);
-                $sql .= "AND `date` >= '" . $date . "'";
-            }
+        $sql = "SELECT COUNT(*) AS `object_cnt` FROM `object_count` WHERE `object_type`= ? AND `object_id` = ? AND `count_type` = ?";
+        if ($threshold > 0) {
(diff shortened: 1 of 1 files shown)


cquike:
Note that there is other place in stats.class.php which also has logic depending on whether cron_cache is on: function get_top_sql. I think it will suffer from the same problem: that function will get obsolete data until bin/cron.inc is executed. However, since it is about getting "top queries" I guess it might be acceptable to have a delay (equal to the cron frequency). Probably that should be discussed since the queries in this case are quite different, so I don't know whether there is a performance improvement using cache_object_count.

lachlan-00:
the cache is used to stop larger sites from grinding to a halt by counting the object_count table on each page load. for single user sites or newer sites without a lot of data there won't be a difference. but it also stops a lot of extra processes running in the interface which does have a little speed increase.

I think i'll put in a tooltip on the setting and a warning when cache is enabled and hasn't been run for a long time like the way the update warnings pop up does it.

cquike:
Hi,
thank you for the explanation. It is true that in my case, being a site with few users and only 15K songs the query COUNT(*) on object_count is quite similar in performance to get the 'count' column on cache_object_count. So my argument that these two queries have similar performance is not be true since they might scale different depending on the site.
However, the point here is really to fix #2587. The problem is that since the cache_object_count is not updated until the cron job is run, the information displayed is _wrong_. This happens even if cron is run quite often.

My typical use case is: I have a subsonic app (D-Sub or Ultrasonic) connected to some loudspeaker in the room while at the same time working on my laptop. It is quite convenient to then check the ampache web page (rather than the mobile phone, which who knows where I left it :) to see the song being played or last played. The surprise is that the song appears in the recently played songs but ampache claims (# played) that the song has been played 0 times (if this was the first time the song was played). And that information remains to be wrong until cron is executed. Note that I have the setting "Show # played" set to on, while the default is off. So the bug is not so evident if the setting is off.

fufroma:
@cquike you can close this PR, it won't be merge as a similar [PR was merged](https://github.com/ampache/ampache/pull/4410). It solve your issue and preserve load for big site.

You'll find it in next major Ampache release, Ampache 8 :)

cquike:
Sounds good!
Thanks for fixing #2587!

PR #2594, opened 2020-10-29
Write metadata to flac and ogg files.

wagnered:
This needs to be revisited: specifically the custom TXXX user fields. Most of them are computer calculated such as replaygain, thus don't need to be modified and written back to the file.
 
Also Ampache's handling of the license might be better,  Jamendo writes the license URL in the copyright(TCOP) frame.  This would seem more appropriate in the song or song_data table


wagnered:
Don't know if it was the merge conflict but after resolving , a problem remained.  Will do again after it is resolved.

PR #2595, opened 2020-10-29
Wagnered write id3

wagnered:
Retrying, the github way


lachlan-00:
lol, you're quick i was writing a comment on the closed one

i don't think many other tools use TCOP, at least picard/musicbrainz and discogs don't fill them so that probably wouldn't cause any conflict.

volume and peak level would be calculated as well but don't know the frams off the top of my head

wagnered:
Not sure why a conflict unless it was because the PR was cross-branch.

Actually, not sure why "licence" was added to Ampache.  Jamendo actually (helps) publish original artists.

lachlan-00:
> Actually, not sure why "licence" was added to Ampache. Jamendo actually (helps) publish original artists.

i think a lot of sites like dogmazic and netlabels use them when publishing tracks. i've gotten < 5 releases myself that were given license details in the files.

Not very common for a personal user but at least fairly common on bigger sites.

wagnered:
Although according to id3.org they don't use it correctly.

I'll have to check out some Dogmazic tracks and see how they add the license/copyright.  Dogmazic may only publish artists, not produce them, so have no part  in the content of the file tag.
Anyway, if the user selected license is to be written as copyright, perhaps TYER could prrpended to the string.

wagnered:
I downloaded one at random (Don't know much French.) Must have been self-produced:  id3v1 tags were converted to id3v1 and written to PRIV frames. The  license type was written in the COMM frame, no URL

.Did you notice that getID3 "cheats" by using platform apps to write flac and off files? 

Oh, I just remembered. This PR upgrades getID3 to 2.0-dev.  It uses namespaces and it conforms to psr-4 standards. .

wagnered:
Anyway, the first letter of the classes is upper case, I need to somehow make this "transitional" by using aliases or it will be a breaking change.

"The difficult I can do right away.  The impossible takes a little longer.🙂"

oops!

wagnered:
Question:. Would it really be a breaking change if it only requires a `composer install` or a `composer update` command be executed?

I'll leave it up to you.  It's ready now, but I'd appreciate you giving it a workout before merging, just in case.

lachlan-00:
I've 'broken' develop over the last few months and this next release is going to be pretty majorly breaking so it's not a huge deal for me. (i'm trying to get to work on a warning/notification banner for this type of thing as well as cron warnings when enabled but not run) 

I'll get onto your PR branch and check things out. i won't be able to do a lot until Monday but i'll at least get on the branch today and go over it.

lachlan-00:
i'm lost from the merges this weekend. has this been put in from your merges now?

lachlan-00:
I think this got merged by the other merges on the weekend, not needed now it's in.

Issue #2453, opened 2020-07-15
Changing album name requires once for each song in album.


wagnered:
That's because the changed album name is given a new index number.  Off hand, I don't understand why that is necessary, especially for a single album: some of my albums have as many as 30 songs.

wagnered:
Also seems to me that only `Album::update_field($field, $value, $album_id)` would be necessary for the task, but I don't see the entire picture.

lachlan-00:
the Album::check function does the creation based on all the fields it checks. (title, year, etc, etc) so if you change tags they don't belong in the original album that was created.

My understanding is that just because one song changes it doesn't mean all the others will as well, like if you want to split up an album that's duplicated you don't want to automatically rename all the other songs in that album.

Ampache uses song as the anchor for everything, album, artist, genre tags, labels, etc so we can't safely assume that a change on 1 song equals a change on all of them. I've reduced the amount of changes that happen on each update a lot though, what used to happen was each time an album was updated it would cycle through each song again. (10 song album would update the 9 other songs as it cycled through so that would be 100 updates for 10 songs.)

wagnered:
That all may be. But if only changing the album name, the album id doesn't need to change does it? The songs and other stuff stays with it.

BTW If the album id is changed when the name is changed, all the songs should be brought along with the new id at the same time. Not one at a time as it is done now.  Also as it is now, when the last song is gone, there will remain an empty album in the catalog.

I'm looking into it. I want to be able to change the name and remove the prefix if the altered name no longer contains one.

PaulWalkerUK:
I suppose it's a matter of perspective. From the point of view of editing an album to change the title (e.g. correcting a spelling mistake), then you could argue the ID and all the songs aren't changing, it's just an attribute of the existing album that's changed.

From the point of view of editing indivisual songs to change the name of the album, then I can see why ID's might need to change (some songs have the original album, some have the new one = two albums). But if you were making the same change to _all_ songs on the album, then we're back to the first scenario

wagnered:
@PaulWalkerUK thanks for the clue. I gotta look closer at the 'Add New' request from the song's drop-down album list.  Hopefully it will provide a way to differentiate between that and just simply changing the album name.

wagnered:
@PaulWalkerUK thanks for the clue. I gotta look closer at the 'Add New' request from the song's drop-down album list.  Hopefully it will provide a way to differentiate between that and just simply changing the album name.

lachlan-00:
[...]

PR #2597, opened 2020-10-31
Edit album

wagnered:
Ref:  #2453


wagnered:
Conflict should be fixed by now.

PR #2600, opened 2020-11-01
Player's current song color updated

Simounet:
I there,
I updated the current song's color to be more visible. It seems it was like that back in the time.
ampache-current-song-color

Have a nice day and stay safe.


lachlan-00:
much clearer to see

PR #2607, opened 2020-11-03
mailauth is an integer and not boolean

gramakri:
the "true"/"false" values in ini file is parsed as 1/0 and thus strict equality check won't work


gramakri:
Since we parse the config file using INI_SCANNER_NORMAL mode, we get 1/0 instead of boolean true/false.

lachlan-00:
now that's a good find!

lachlan-00:
enjoy your contributor status!

PR #2609, opened 2020-11-08
Login page centered content

Simounet:
Hi there,
Another quick login page style update.

Before:
ampache-before

After:
ampache-after

Stay safe and have a nice day.


lachlan-00:
don't mind that, looks good on mobile and web

PR #2612, opened 2020-11-08
jPlayer fullscreen shortcut disabled

Simounet:
A long time annoying issue removed. I love to use the well known shortcut `CTRL+f` to find something on the page but jPlayer bind the `f` key without checking if `CTRL` was pressed too. With the useless fullscreen view, we don't need to bother anymore.

Have a nice day.


lachlan-00:
that is another really good annoyance.

Issue #2527, opened 2020-09-08
[Feature Request] - Total time


4phun:
Hi, I think it would be a great addition to the web UI to show the total time for playlists, albums, and wherever else it may apply.

lachlan-00:
at some point album needs to have a total time column and playlists do have this calculated on the fly but i think we should start updating these into columns

theTendo:
Hello, I see this issue is in the <<In progress>> category, however, there is no one assigned to it. If it is avaiable I would like to take care of it.

lachlan-00:
> Hello, I see this issue is in the "In progress" category, however, there is no one assigned to it. If it is avaiable I would like to take care of it.

it's back in triage/todo list so feel free to take it on.

theTendo:
Okay, I will work on this.

theTendo:
> playlists do have this calculated on the fly
You mention time length already is calculated for playlists, but I can't seem to find it nor in the UI nor in the code. Could you point me to what you mean?

lachlan-00:
Most things are calculated from the song table, it's the only table with a time in it so albums and playlists, etc all calculate this from a song length.

In playlist.class we get the songs and sum the total time
![image](https://user-images.githubusercontent.com/1305249/96932673-9642a500-1502-11eb-884b-da3dbc1ffd9e.png)

in album.class we do the same in "_get_extra_info"
![image](https://user-images.githubusercontent.com/1305249/96932792-cb4ef780-1502-11eb-8ab9-fc7aeb5e8acd.png)

theTendo:
I looked at the way **# Items** is shown and did the same for **Duration** using the `get_total_duration()` function you mentionded. The UI looks like this with this addition:

![commit _7671b2e](https://user-images.githubusercontent.com/2078392/97000388-f18e8a80-1536-11eb-932a-726f83e2beb9.png)

4phun:
Hi, this is cool! Can we get it on the playlist screen too? Where the songs are displayed.

theTendo:
After a bit of thought I decided the best place for the **Duration** info would be adjacent to the **Item Count** info. However, the item count is displayed by `list_header.inc.php` which is generic to all lists and doesn't have an easy way to access the duration function. So in the end I added the changes to `show_playlist_medias.inc.php` with the following result:

![Screenshot_20201026_181313](https://user-images.githubusercontent.com/2078392/97206089-923fad00-17b8-11eb-9bda-6518baa630da.png)

[...]

PR #2614, opened 2020-11-10
#2527 Playlist duration added to UI

theTendo:
Initially, it was added as a column in `show_playlists.inc.php`, however, after user feedback, it has been changed to be shown inside `show_playlist_medias.inc.php` for a better look.

note 1: It works both for normal playlists and smart playlists.
note 2: In case the duration is >86400 seconds (a day) the hour format takes in all the extra hours i.e. 99728s = 27:42:08.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -22,6 +22,7 @@ This means Ampache now **requires** php-intl module/dll to be enabled.
 * Disable API/Subsonic password resets in 'simple_user_mode'
 * NEW plugin: 'Personal Favorites'. Show a shortcut to a favorite smartlist or playlist on the homepage
 * Run garbage collection after catalog_update.inc 'clean' or 'verify'
+* Added duration when browsing playlists and smartlists
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


lachlan-00:
Thanks for posting this one, i'll be able to test and merge after work!

lachlan-00:
for large lists are these hour minutes seconds still?
image


only other thing left to check is the duration for searches is setting the duration after displaying, i can probably do that but have had to go home early today for family stuff.

theTendo:
> for large lists are these hour minutes seconds still?
> image

Yep, the example from your screenshot would represent 1006 hours, 46 minutes and 0 seconds

> only other thing left to check is the duration for searches is setting the duration after displaying, i can probably do that

Okay, nice 👍

> have had to go home early today for family stuff

No problem, family is more important than this :)

lachlan-00:
okay, we're done here. i see where it's happening now.

when you 'browse' a list you run a search but we just keep an array of objects instead of the search/playlist. so we need to set a duration at the browse time instead of relying on the search/playlist count as the objects will probably not match.
image


last_duration in this respect is actually a second search after the original browse.

Thanks for doing the leg work on this it was actually really good to follow up and work on the code in this area. Now you're the 101st Ampache contributor!

PR #2618, opened 2020-11-13
Support image per song

Greo83:
Song-related images can be saved in the database additionally to album-related images.
See original discussion here [https://github.com/ampache/ampache/pull/1847](url)

PR #2620, opened 2020-11-14
support for image per song

Greo83:
Optionally save and show the picture of a song related to the song instead of its album. Use this feature if your music collection contains more single songs than full albums.

Original discussions:
[https://github.com/ampache/ampache/pull/2618](url)
[https://github.com/ampache/ampache/pull/1847](url)


wagnered:
Almost.  You forgot to run `php-cs-fixer -vvv fix` before pushing to your fork.☺️

Otherwise it looks quite clean.

Here is a ref for the proper way to [sync your fork](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork).

I'll give your code a try as soon as I can get to a stopping point in my current project.

lachlan-00:
Something that is hard to remember is to run the codestyle checkers. (i wish you could script that as part of a commit)
```
./php-cs-fixer -v fix && ./phpcbf --standard=.phpcs.xml lib/class
```

I'll be able to work on this during the week, it's been a rough house week here.

wagnered:
So far so good. Now how do I test it?
Really need more explanation. 

Are the images dispayed on the song page of the web UI?  What happens to the existing cover/artist images in the db?

  Do I need to have non artist/cover images embedded in the file tag?  I ask this because it affect the updating onf the tags which must be written in it's entirety each time.

lib/class/art.class.php
```diff
@@ -948,8 +948,17 @@ public function generate_thumb($image, $size, $mime)
      * @param string $type
      * @return string
      */
-    public static function get_from_source($data, $type = 'album')
```
wagnered, line 951: why did you remove the default value for parameter 'type'?

arts.php
```diff
@@ -160,7 +160,12 @@
         // Prevent the script from timing out
         set_time_limit(0);
 
-        $image      = Art::get_from_source($_SESSION['form']['images'][$image_id], 'album');
+        $show_song_art = AmpConfig::get('show_song_art') ?: false;
```
wagnered, line 163: The ternary operatioin `?:` isn't necessary and also isn't fool proof.  If you look at the function definition ` AmpConfig::get('show_song_art')` it has a second parameter  which can set a return value.

Just a reminder for the next PR you do.

lib/class/art.class.php
lachlan-00: did a search on it and don't see any calls with a null type, but it's still a fallback i would keep

Greo83:
> So far so good. Now how do I test it?
> Really need more explanation.
> 
> Are the images dispayed on the song page of the web UI? What happens to the existing cover/artist images in the db?
> 
> Do I need to have non artist/cover images embedded in the file tag? I ask this because it affect the updating onf the tags which must be written in it's entirety each time.

Imagine you have three songs/files. Standalone songs. Each song has one picture in its ID tag. In fact the picture shows the title of the song, so it is absolutely unique. The album tag is empty.
Ampache's standard behaviour in this scenario then is: When importing the first file it creates a new album entry called "orphaned". This new album gets the picture of the first song. The next files are attached to the same album - "orphaned". The pictures are not imported because the album already exists.
When the songs are shown in Ampache, no matter if in the web player or playlist or whatever, all three songs show the picture of the first file. This is because all of them are linked to the "orphaned" album and Ampache always shows the picture of the album.

When I started to investigate this I had the idea to handle the album "orphaned" in a different way. But then I found out that all pictures are linked to the albums in the SQL database. 
Because of this I created the option to additionally (!) add the picture with a relation to song. For our scenario above this means: You still have the album "orphaned" with the first picture, everything as before. But additionally the picture of each file is also saved in the database, with the relation "song". When playing a song, showing a playlist and so on Ampache now shows the song-related picture instead of the album-related picture.

This feature is interesting for people who have mostly stand-alone songs instead of full albums in their library.

lib/class/art.class.php
Greo83: My problem was that the function always used the default value 'album', although I called the function with the parameter value 'song'. My parameter was ignored, it always was overwritten by 'album'. I had a long debugging session until I realized this.

But to be honest I'm absolutely new to PHP - maybe the reason was different?

lachlan-00: i'll check it out, it's possible that something is being sent null

wagnered: Well here is the [php authority](https://www.php.net/manual/en/).  It's always good to have a default value: no one knows how it migt be accessed in the future.

image.php
```diff
@@ -51,7 +51,7 @@
 
 // FIXME: Legacy stuff - should be removed after a version or so
 if (!filter_has_var(INPUT_GET, 'object_type')) {
-    $_GET['object_type'] = 'album';
```
lachlan-00, line 54: what does this change do? i'm assuming it's setting a default type for art?

arts.php
lachlan-00: these are changed into different operators based on need now, the false isn't required in the way it's written

lib/class/art.class.php
lachlan-00: There's code below to set album/song based on the config. i'm not sure when this would ever fire but there must be something not format or set before the call somewhere outside this
```
public static function get_from_source($data, $type)
    {
        if (!isset($type)) {
            $type = (AmpConfig::get('show_song_art')) ? 'song' : 'album';
        }
```

wagnered:
>  Each song has one picture in its ID tag. In fact the picture shows the title of the song, so it is absolutely unique. The album tag is empty.

How are those images added to the tags?

What makes an image unique is a combination of the assigned picture's id as denoted by id3.org and the description of the image.  It matters not if an "album" frame exists.  The picture id associates the image with the cover, artist, conductor.  Granted, Ampache does a poor job of differentiating between artist an cover.

I'm concerned about the writing of the images back to the file.  The getID3 library adheres to the id3.org specs for id3v2.3 which  Ampache uses. If the embedded images don't conform, they will be lost upon tag rewrite. 

I think what you want can be done more transparently by using the `kind` column. Another config switch might not be necessary.  There are so any now that it take an undergrad degree to fully understand Ampache 😄.

I would like you to give it some thought.

image.php
Greo83: To be honest I don't remember. I changed that line 1 1/2 years ago in a lot of debugging and try and error sessions. When I review that now I would agree. I think this default setting should not be hard coded, but depending on the config file.

Greo83:
> How are those images added to the tags?
> 

What do you mean with that? I checked some files. The image type of the ID tag is "front cover" mostly.

> I'm concerned about the writing of the images back to the file. The getID3 library adheres to the id3.org specs for id3v2.3 which Ampache uses. If the embedded images don't conform, they will be lost upon tag rewrite.

Okay, I didn't try that. I use Ampache as a read-only tool. Is the object_type field really passed back directly into the ID tag?

>
> I think what you want can be done more transparently by using the `kind` column. 

Theoretically it should work, yes. But then all the database statements that Ampache calls have to be changed. This seems to be more effort, isn't it? Additionally, I think from a database point of view this is not correct, because the image is assigned to the song - not to the album. As far as I understood all the SQL and PHP code around the object_type field everything was designed to attach images also to different other tables than the album table. It just has not been developed yet.

wagnered:
The fact that you want to use cover art (id=3) generically  does complicate the issue.  I want the spec to be that "song"art could have any ID except that of cover and artist.  This will make the design  much less complicated.  The ID of your current embedded images coud be altered relatively easily with easy tag or eyeD3 on Linux.  I'm placing this limitation on the image spec because I shall be implementing image display on the song page and perhaps the song row page if I can squeeze it on.    Your PR  doesn't do this. 

>  But then all the database statements that Ampache calls have to be changed.

Not necessarily.  But  familiarity with a project in total helps one see solutions that affect the current development as well as future possibilities.  

But the near-future architecture will include a database managent system (dbms) that will make SQL statements  more manageable.

lib/class/catalog.class.php
```diff
@@ -1460,14 +1460,22 @@ public function gather_art($songs = null, $videos = null)
...
+            }
         } else {
             $searches['album']  = array();
             $searches['artist'] = array();
+            if ($gather_song_art) {
+                $searches['song'] = array();
+            }
             foreach ($songs as $song_id) {
                 $song = new Song($song_id);
                 if ($song->id) {
```
wagnered, line 1480: The objects were already created/returned on line 1471. To be consistent, I suggest that song IDs be returned and objects created here.

wagnered:
I'm feeling more comfortable with the PR because it seems to apply only to API access

I think that because Art::gather cycles through every entry in `art_order` there is an assumption that the `art_order` setting includes only "db" and "tags".  How would adding spotify to the list affect the results?

AshotN:
> Something that is hard to remember is to run the codestyle checkers. (i wish you could script that as part of a commit)
> 
> ```
> ./php-cs-fixer -v fix && ./phpcbf --standard=.phpcs.xml lib/class
> ```
> 
> I'll be able to work on this during the week, it's been a rough house week here.

Doesn’t GitHub actions do this?

lachlan-00:
> 
> 
> > Something that is hard to remember is to run the codestyle checkers. (i wish you could script that as part of a commit)
> > ```
> > ./php-cs-fixer -v fix && ./phpcbf --standard=.phpcs.xml lib/class
> > ```
> > 
> > 
> > I'll be able to work on this during the week, it's been a rough house week here.
> 
> Doesn’t GitHub actions do this?

it looks like it might actually... i'll check it out.
image

lachlan-00:
@Greo83 i'm going to start going through this to merge now.

usox:
> > Doesn’t GitHub actions do this?
> 
> it looks like it might actually... i'll check it out.

I could provide working workflow configs/examples if you wish.

lachlan-00:
> > > Doesn’t GitHub actions do this?
> > 
> > it looks like it might actually... i'll check it out.
> 
> I could provide working workflow configs/examples if you wish.

I think this would help docker as well because I could build a zip/gzip for building the image for develop

AshotN:
Correct me if I’m wrong. But my understanding is this will add a image for every song. Wouldn’t an image override option be easier? 

So if the album has an image, but a specific song within that album has a different image, that different art alone would be stored. 

To avoid needless duplication.

usox:
> > > > Doesn’t GitHub actions do this?
> > > 
> > > 
> > > it looks like it might actually... i'll check it out.
> > 
> > 
> > I could provide working workflow configs/examples if you wish.
> 
> I think this would help docker as well because I could build a zip/gzip for building the image for develop

Oh, I guess I got you wrong. You're talking about automatically fixing the code after pushing a commit (serverside at github), right?
I'm not a big fan of automatically altering peoples commits. A failing build of a PR shouldn't be a big problem for the developers here (My first PR failed, too). Maybe, we could add a CONTRIBUTING.md at root level of the repo to better communicate our rules...

AshotN:
> > > > > Doesn’t GitHub actions do this?
> > > > 
> > > > 
> > > > it looks like it might actually... i'll check it out.
> > > 
> > > 
> > > I could provide working workflow configs/examples if you wish.
> > 
> > 
> > I think this would help docker as well because I could build a zip/gzip for building the image for develop
> 
> Oh, I guess I got you wrong. You're talking about automatically fixing the code after pushing a commit (serverside at github), right?
> I'm not a big fan of automatically altering peoples commits. A failing build of a PR shouldn't be a big problem for the developers here (My first PR failed, too). Maybe, we could add a CONTRIBUTING.md at root level of the repo to better communicate our rules...

Just set it up to fail instead of modifying the commit. But for formatting it shouldn’t really be an issue.

kuzi-moto:
> > > > > Doesn’t GitHub actions do this?
> > > > 
> > > > 
> > > > it looks like it might actually... i'll check it out.
> > > 
> > > 
> > > I could provide working workflow configs/examples if you wish.
> > 
> > 
> > I think this would help docker as well because I could build a zip/gzip for building the image for develop
> 
> Oh, I guess I got you wrong. You're talking about automatically fixing the code after pushing a commit (serverside at github), right?
> I'm not a big fan of automatically altering peoples commits. A failing build of a PR shouldn't be a big problem for the developers here (My first PR failed, too). Maybe, we could add a CONTRIBUTING.md at root level of the repo to better communicate our rules...

I swear we had it somewhere. Obviously it's not easy to find, so it definitely should exist in the CONTRIBUTING doc.

lachlan-00:
k, happy with this now, will merge once it's built in travis

lachlan-00:
i migrated to travis.com from the org so i'm not waiting for that.

Issue #1547, opened 2017-05-26
Can't seem to get Ampache DLNA to work


drder66:
Apologies for noob unfamilarity, but I'm having trouble getting the DLNA interface to work (independently of the DAAP/RSP interface in my other ticket)

I'm trying to follow the instructions:

> You must enable UPnP Backend in System settings.

Ok, done

> You must also enable Enable url rewriting in Streaming settings.

I think this is done, but what user level to use ? I've tried ticking the "all" box, and "user" and "guest" and "admin" all to no avail.

> If your installation is within a subfolder '/ampache' you must take care for your Rewrite Rules within '[ampachepath]/play/.htaccess'.

It is in /ampache, so I modified .htaccess thus:
IfModule mod_rewrite.c
    RewriteEngine On
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteCond %{REQUEST_FILENAME} !-s
    RewriteRule ^art/([^/]+)/([^/]+)/([0-9]+)/thumb([0-9]*)\.([a-z]+)$ /ampache/image.php?object_type=$2&object_id=$3&auth=$1&thumb=$4&name=art.jpg [L]
    RewriteRule ^([^/]+)/([^/]+)/([^/]+)/([^/]+)(/.*)?$ /ampache/play/$5?$1=$2&$3=$4 [N,QSA]
    RewriteRule ^([^/]+)/([^/]+)(/.*)?$ /ampache/play/$3?$1=$2 [N,QSA]
    RewriteRule ^(/[^/]+|[^/]+/|/?)$ /ampache/play/index.php [L,QSA]
/IfModule

> PHP Socket mod must be enabled and PHP XMLReader extension available

Now you're losing me. Do you mean PHP socket modification, or PHP socket module ? How would I know which and whether they were enabled ? And from what I can find out the XML reader has been default enabled in the binary releases for some time. 

> You should broadcast the server periodically calling http://localhost/ampache

/upnp/?btnSend=Send+SSDP+broadcast or php bin/broadcast.inc with a cron (every 1-2 minutes recommended).
This seems pretty ugly as it's filling either my apache logs or cron logs or both. For now I've been using the /ampache/upnp/?btnSendAuto=Send+SSDP+broadcast+every+second link to do this, which has duly been filling apache logs up with 

[...]

PR #2633, opened 2020-11-29
Upnp changes

drder66:
This is my contribution to fixing issue #1547 (DLNA not working)
Covers many issues, described in the commit log.
Tested across a variety of devices, none of which worked properly (or at all) before these changes
Apologies if I haven't got things right, my first Github submission


lachlan-00:
upnp was i think widely broken before you've had a look at this, really great to see how much effort has gone into this!

lachlan-00:
fyi all i'm doing is working through the scrutinizer issues (which is why i'm committing over the top.)
https://scrutinizer-ci.com/g/ampache/ampache/inspections

becuase of the age of the code i think it'll be good to keep some of the old code commented out as you've done it so rearranging and making things match up with the style i've been keeping.

lib/class/upnp_api.class.php
```diff
@@ -991,13 +1191,317 @@ public static function _videoChilds($prmPath, $prmQuery, $start, $count)
...
+        for ($i=0; $i < $actualsize; $i++) {
+            $actualtokens[$i]=trim($actualtokens[$i]);
+            if ($actualtokens[$i] != "") {
+                $nospacetokens[$h++] = $actualtokens[$i];
+            }
+        }
+
+        // now put together tokens which are actually one token e.g. upper hutt
+        $onetoken    = "";
+        $h           = 0;
```
lachlan-00, line 1221: can we expand this variable name out to something?
i'm not familiar with the hutt word outside star wars?

drder66: Yes, that's not a very helpful comment.

// now put together the tokens which are not going to be search qualifiers

Would express the intent better

composer.json
```diff
@@ -41,6 +41,7 @@
         "blueimp/jquery-file-upload": "10.2.0",
         "carhartl/jquery-cookie": "1.*",
         "cboden/ratchet": "0.4.*",
+        "clue/multicast-react": "dev-master",
```
usox, line 44: HI. Could you specify a version constraint instead of using the master in here? That would be great

lachlan-00: it looks like you willed it into existence. they only just release 16 hours ago
https://github.com/clue/reactphp-multicast/releases/tag/v1.1.0

lachlan-00:
@drder66 i'm up to getting this merged now, are you able to pull my updates and check it out?

Was there any specific program/test you used to try all this out? it'd be good to actually write stuff down in the wiki if you have any process you followed

lachlan-00:
i was also going to move the ssdp.service into examples and write up a upnp.md file in the docs folder so anything you can write about it would be really helpful for others

drder66:
Is there anything else you need me to do?

lachlan-00:
I'll be home by the weekend to check it all out again. Should be fine, as long as I can browses the server it's ready

PR #2641, opened 2020-12-05
Add live streams to getindexes API method

4phun:
Hi,

I would like to support Radio Stations in my client. This pull request adds them to the get_indexes API method. Please consider for inclusion in the next version.

Thank you

### DIFF
--- a/lib/class/xml_data.class.php
+++ b/lib/class/xml_data.class.php
@@ -365,7 +365,7 @@ class XML_Data
      * we want
      *
      * @param  array $objects (description here...)
-     * @param  string $object_type 'artist'|'album'|'song'|'playlist'|'share'|'podcast'|'podcast_episode'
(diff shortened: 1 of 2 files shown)


lachlan-00:
Sounds good. I'll test and merge on Monday

lib/class/xml_data.class.php
```diff
@@ -480,6 +480,14 @@ public static function indexes($objects, $object_type, $full_xml = true, $includ
                 case 'video':
                     $string .= self::videos($objects);
                     break;
+                case 'live_stream':
+                    $live_stream = new Live_Stream($object_id);
+                    $live_stream->format();
```
lachlan-00, line 485: don't think the format is needed as f_name is just scrubbed just finalising now.

lachlan-00:
actually lets just keep the format

4phun:
Cool, thank you

lachlan-00:
found a bug in the json while adding the json stuff so that's helped!

Issue #2582, opened 2020-10-17
[Feature Request] Changing login screen background


agopo:
## Describe what you don't like about Ampache, or what it is missing
I'd like to be able to change the login screen background without having to tamper with the main theme.

## Describe how you would like the feature to work**
Login as admin -> Settings -> User Interface -> User defined URL Loginpage Background (Enter URL or path/to/file.png here)
Logout
Have new background on Loginpage

lachlan-00:
that is a prretty good idea, just needs a database preference and the ability to insert it into the images table probably

theTendo:
I've done some work related to this in commit [dad74f1](https://github.com/theTendo/ampache/commit/dad74f17efc677fce025fc7b391c78dec2ece712):

![commit_dad74f1_1](https://user-images.githubusercontent.com/2078392/102138482-b0db2f80-3e5c-11eb-916f-f1796d0cd311.png)

![commit_dad74f1_2](https://user-images.githubusercontent.com/2078392/102138500-bafd2e00-3e5c-11eb-9ee3-434a0335891e.jpg)

I do have some doubts regarding the preference's id attribute. If the update is applied via `update.class.php` this preference setting is assigned id 172, but apparently this id is already used by "Database preference table", so I don't know if there could be an overlap problem with that.

I haven't done any kind of test apart from the fact that the background is shown correctly, so any testing suggestion is welcomed.

lachlan-00:
You couldn't have picked a nicer background for it!

If you make a pull I can clean it up and put it in tomorrow

theTendo:
PR created :)

lachlan-00:
Cool, I'll do it first up tomorrow.

agopo:
Hey,
I honestly didn't expect this to be implemented so quickly (if at all). Thanks a lot for your work, I'm looking forward to it!
Cheersagopo

    Am Dienstag, 15. Dezember 2020, 09:32:05 MEZ hat lachlan <notifications@github.com> Folgendes geschrieben:  
 
 


Cool, I'll do it first up tomorrow.

lachlan-00:
okay, it looks good, i didn't really change anything just took things out not needed.

this preference will be a system pref as when you go to the login page you don't have a user set.
Things like translations get handled out of regular changes and it doesn't need to be inserted for users to change.

I'll be merging in a few minutes after all the travis checks go through.

lachlan-00:
9546c80a7 now has a custom login background setting!

[...]

PR #2647, opened 2020-12-15
#2582 Customizable login page background

theTendo:
Inserted a new `custom_login_background` preference that allows the user to define a custom login background. This preference is given priority 75 (Catalog Manager), same as the already defined `custom_login_logo`.

Also inserted a **msgid** field in `messages.pot` and `translatable-database-strings.txt` to allow translations of the "Custom URL - Login page backgroud" string.

Screenshots are avaiable in the [issue's 2582 page](https://github.com/ampache/ampache/issues/2582).


lachlan-00:
boom, what a nice little change!

PR #2649, opened 2020-12-16
Make hls view work for video/vlc

nicklan:
The hls view logic didn't used to handle videos, and also used a `Content-Type` header that didn't work for vlc. This fixes both those issues.

You can read more about the header stuff here: https://tools.ietf.org/html/rfc8216#section-4


lachlan-00:
love it, adding the change made hls streaming work in dsub.

nicklan:
hah, yep, that's exactly why I patched it :)

PR #2650, opened 2020-12-16
React client revised

mitchray:
firefox_O6sOI1kd36

I'm chipping away at this until it gets reviewed, recommend squashing the pull request if its all OK to keep it tidy


lachlan-00:
@AshotN can review when he's free again and it's his branch for merging.

Really glad to see more effort from new people though 2020 has been a really good year for contributors. :)

usox:
Mhmm just thinking about this...
Would it make sense to extract the complete react-related code into an external repository? This way, we could separate the development and pull specific builds of the frontend in our core releases. We could then ship the builds as some kind of test-version/alpha with every update to give the interested users something to play with...

mitchray:
I've been wondering something similar, how to get onto the new v5 API. @AshotN may already have a plan for this.

Would having a separate client repo with Ampache as a composer dev dependency work? Then the client can be built for a specific API version and untie the core from the client as you suggested. I'm not sure what kind of work would be needed to make it composer compatible

usox:
@mitchray I'd say that composer won't be _the_ problem.

But we'll need a complete "build" (build artifacts after a npm/yarn run or whatever is used in here) of the version with a certain tag in the repo - I don't know how we can accomplish that in a good manner. Maybe, this could be achieved using github packages...

The other solution would be to pull the source with a specific tag and built it when updating/releasing ampache. But I'd prefefer pre-built artifacts...

mitchray:
The React client (essentially the 'client' folder) would be extracted to its own repo without the rest of Ampache code right? But it needs Ampache to function so is having Ampache brought in via composer a solution? 

That's what I meant with composer, just wanted to clarify. I don't have experience with this scenario but have often wondered, research I've done has led me to submodules/subtrees which I hear is complicated and prone to errors...

usox:
Mhhkay. I think we need to distinguish between development and release packaging.

For development purposes, having separate directories containing the ampache core and the react client should be sufficient imho. This could be configured within the local/dockerized webserver and should be easy to setup and maintain.

For packaging, we need to glue frontend and core together - ideally by having ready-to-use builds of the react client which could be obtained by e.g. composer, so building an ampache release doesn't require yarn/npm to be installed.

But this a re just some ideas of mine - maybe someone has a top notch solution for this... :smile:

lachlan-00:
For now this separate branch is okay.

It will definitely go into a new repo later (I'll fork it at some point) and script up an installer

AshotN:
Sorry for my inactivity

As far as deployment for users goes, I figured a script to download the compiled HTML/JS/CSS files would be sufficient, since it's all client side code it should be OK.

I'm gonna try running through the code changes now.

AshotN:
Looks pretty great. Let me know when you are done making your final touches so I can merge and add some of the missing functionality in the UI. 

One thing I am going to do is remove the feature that uses the back button to close dialogs, if I can't figure out how to fix it, it's messing up the browser history. Causes issues.

@lachlan-00 I've fallen a bit behind, is the new API stable? Can I begin implementing it?

client/components/QueueBar/index.styl
```diff
@@ -1,47 +1,72 @@
+@import "~stylus/color.styl"
 @import "~stylus/sizes.styl"
 
+queuebarWidth = 280px
```
AshotN, line 4: Should be part of the `sizes.styl` file.

client/components/SongBlock/index.module.styl
```diff
@@ -22,11 +19,5 @@
         clear both
         float left
         color inherit
-        &:hover, &:focus
-          color linkHoverColor
   &.playing
     color currentlyPlayingColor
-  &:focus
```
AshotN, line 29: These focus's a lot of the time are for accessibility reasons, for someone to be able to tab through the UI

client/stylus/main.styl
```diff
@@ -1,28 +1,30 @@
+@css {
+  @import url('https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&family=Rajdhani:wght@400;700&display=swap');
```
AshotN, line 2: There is a discussion to be had for if we intend to bundle font's in the project.

client/components/MusicControl.tsx
```diff
@@ -1,107 +1,209 @@
...
+                </div>
+            </div>
+
+            <div className='seekbar'>
+                <Slider 
+                    min={0}
+                    max={musicContext.currentPlayingSong?.time ?? 0}
+                    value={isSeeking ? seekPosition : musicContext.songPosition}
+                    onChange={(event, value: number) => {
+                        // setIsSeeking(true);
```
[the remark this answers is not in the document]
mitchray: Please do, once I changed the react-range-slider to MaterialUI Slider the seeking appeared to be working without them so I just commented them out for the time being

client/components/SongBlock/index.module.styl
mitchray: Absolutely, I still need to restore a lot of these

client/stylus/main.styl
mitchray: Do you mean their usage in general or would this cause React to literally bundle the files as part of the production build?

mitchray:
If you're happy to proceed I'm fine with how it currently is, its still a WIP of course. Let's get it hooked into the new API then I'll jump back in

lachlan-00:
> @lachlan-00 I've fallen a bit behind, is the new API stable? Can I begin implementing it?

Short of additional methods or minor fixes I'm not going to change anything else. The structure and data is all how I wanted

client/stylus/main.styl
AshotN: Both. If we request the fonts from a CDN, people running Ampache on a closed network won’t have the fonts.  There are also privacy concerns regarding Google fonts. 

The only benefit of using Google fonts is speed when loading out of the local network.

AshotN:
> > @lachlan-00 I've fallen a bit behind, is the new API stable? Can I begin implementing it?
> 
> Short of additional methods or minor fixes I'm not going to change anything else. The structure and data is all how I wanted

Cool. 

Now I just got figure out how to squash the commits and I can merge it. Once I’m on a computer I’ll try!

Thank you @mitchray for the beautiful UI

PR #2653, opened 2020-12-19
Fix default value of additional_genre_delimiters

thinca:
`[...]` is a character class.  Repeating `|` is not needed.


lachlan-00:
does the \\\\ just need to be \\ as well? seems like a dupe of the same character

thinca:
We need escaping `\` twice, for string literal, and for regexp.

lachlan-00:
Sounds good.thanks for putting this up. I've always sat on this wondering. Will test out after Christmas week and merge.

PR #2654, opened 2020-12-19
Order by name and disk when searching album

thinca:
I have albums that consists of multiple disks.
I want to display them in disk number order.


lachlan-00:
just needed to test on full_group_by in case it wasn't in the select. all good

PR #2670, opened 2020-12-27
Make http_host auto-estimation more robust

MichaIng:
Prefer `HTTP_HOST` over `SERVER_NAME`, if defined. SERVER_NAME is not reliable as it depends on the webserver config and might be empty, set to the local server hostname, a wildcard/regex or another invalid string for internal identification only. HTTP_HOST matches the client request and should hence work the same way it did for the current client request, when using for redirects. If in unexpected circumstances both variables are not set or empty, "localhost" is used as fallback.
_________
Other uses of SERVER_NAME:

I see `local_web_path` below using SERVER_NAME as well, but since it shall be a local URL explicitly, SERVER_NAME is probably the better guess based on how most webservers set it by default.

What I do not understand is the CLI block below:
```
if (!defined('CLI')) {
    $_SERVER['SERVER_NAME'] = $_SERVER['SERVER_NAME'] ?: '';
}
```
It seems like to have zero effect, assigning the same value if it is set and an empty string if not?

Probably it makes sense to use `$results['http_host']` as well for the OAuth module here: https://github.com/ampache/ampache/blob/develop/modules/oauth/OAuth.php#L374

### DIFF
--- a/lib/init.php
+++ b/lib/init.php
@@ -83,7 +83,13 @@ if (isset($ow_config) && is_array($ow_config)) {
 
 $results['raw_web_path'] = $results['web_path'];
 if (empty($results['http_host'])) {
-    $results['http_host'] = $_SERVER['SERVER_NAME'];
+    if (!empty($_SERVER['HTTP_HOST'])) {
+        $results['http_host'] = $_SERVER['HTTP_HOST'];
+    } elseif (!empty($_SERVER['SERVER_NAME'])) {
+        $results['http_host'] = $_SERVER['SERVER_NAME'];
+    } else {
+        $results['http_host'] = 'localhost';
+    }
 }
 if (empty($results['local_web_path'])) {
     $results['local_web_path'] = $http_type . $_SERVER['SERVER_NAME'] . ':' . $_SERVER['SERVER_PORT'] . $results['raw_web_path'];


usox:
Hi

> Prefer `HTTP_HOST` over `SERVER_NAME`, if defined. SERVER_NAME is not reliable as it depends on the webserver config and might be empty, set to the local server hostname, a wildcard/regex or another invalid string for internal identification only. HTTP_HOST matches the client request and should hence work the same way it did for the current client request, when using for redirects. If in unexpected circumstances both variables are not set or empty, "localhost" is used as fallback.

Afair HTTP_HOST contains the port, if the server runs a non standard-port (80/443) so it couldn't directly be used as a substitute for SERVER_NAME. Despite of that, HTTP_HOST is client controlled, so it may be possible to inject xss code which may be interpreted at some point in the software.

> 
> What I do not understand is the CLI block below:
> 
> ```
> if (!defined('CLI')) {
>     $_SERVER['SERVER_NAME'] = $_SERVER['SERVER_NAME'] ?: '';
> }
> ```

Educated guess: This is a very simple way to ensure the existence of the SERVER_NAME key. As this is used in just 2 cases (captcha and oauth) which are both non-cli, I'd say this line could be dropped...

MichaIng:
> Afair HTTP_HOST contains the port, if the server runs a non standard-port (80/443) so it couldn't directly be used as a substitute for SERVER_NAME

Ah you're right. In this case it would need to be taken as it is, without ever adding a port, i.e. if it contains the port, it likely is required, if not then not.

> Despite of that, HTTP_HOST is client controlled, so it may be possible to inject xss code which may be interpreted at some point in the software.

Yes, indeed, I was reading a bid more into this and when forging POST/CLI requests explicit Ampache config > SERVER_NAME > localhost seems to be safer.

> Educated guess: This is a very simple way to ensure the existence of the SERVER_NAME key.

Makes sense. Yeah probably it would be best to derive a safe canonical hostname only once per request and not read that variable anymore from then on.
_____________
Okay, so I agree that the commit like it is decreases safety in theory. Probably the issue we were facing can be addressed from the other side, e.g. by not forging the whole URL with `$results['http_host']` for client redirects but using it for internal POST and CLI requests only, where it's required. My aim is to script an at least accessible Ampache instance without the need to guess the canonical server name of the target server. This means the Ampache `http_host` setting is not set and the webserver might come with a default value for its server name directive:
- On Nginx the default is an empty string, in which case Ampache falls back to "ampache" to redirect to install/update/login.php scripts on browser access. Not sure where this is done, but IMO it's probably a bad fallback, even compared to "localhost" or "127.0.0.1" 🙂. Most Nginx packages ship a configuration where server_name is set to "_", which is naturally an invalid hostname. Often used is "$hostname" in which case the systems (/etc/)hostname is used, which usually works for local requests but already for local network requests it often does not and very rarely for remote requests. And the Nginx docs as well suggest wildcards, which then only makes sense for Nginx itself to identify the correct vhost but not as canonical hostname.

I'm currently setting up Apache2 and Lighttpd instances, to check how those behave by default.

What do you think about using only location/path redirects when redirecting the client? This seems to be how most other web applications behave, so that whatever hostname/IP was used to access the server is preserved. For internal POST and CLI requests SERVER_NAME should work in most cases, else "localhost" would be a reasonable fallback, but that's what `local_web_path` is used for anyway.

usox:
> What do you think about using only location/path redirects when redirecting the client? This seems to be how most other web applications behave, so that whatever hostname/IP was used to access the server is preserved. For internal POST and CLI requests SERVER_NAME should work in most cases, else "localhost" would be a reasonable fallback, but that's what `local_web_path` is used for anyway.

Well, I don't think there is a need do provide full uris containing protocol, domain/ip and port when redirecting or just displaying some links within the gui. So personally, I don't see a problem in replacing those uris (`http://some-host:42666/some-path/meh` to `/some-path/meh`) - if I understand you correctly.
But honestly: I've no idea of all the pitfalls waiting for us to be explored when changing this... :laughing:

MichaIng:
> Well, I don't think there is a need do provide full uris containing protocol, domain/ip and port when redirecting or just displaying some links within the gui. So personally, I don't see a problem in replacing those uris (http://some-host:42666/some-path/meh to /some-path/meh) - if I understand you correctly.

Yes that is what I was thinking.

> But honestly: I've no idea of all the pitfalls waiting for us to be explored when changing this... 😆

Honestly, me too, i.e. when it comes to proxies and such.
____________
I had a look into how as an example Nextcloud handles it:

_TL;DR: It allows to set an overwrite host, else (without proxy header + trusted proxy array explicitly defined) prefers HTTP_HOST over SERVER_NAME but matches it against a "trusted_domain" array, which is mandatory. The hence forget absolute URL, including scheme and derived hostname, is then used as well for redirects._

1. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/base.php#L1034
2. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/private/legacy/OC_Util.php#L1152
3. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/private/legacy/OC_Util.php#L1104
4. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/private/URLGenerator.php#L245
5. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/private/URLGenerator.php#L271
6. https://github.com/nextcloud/server/blob/86e6f0716203c9fa1032b6fa6484b5b935d61a6e/lib/private/AppFramework/Http/Request.php#L892

Easier to see what is done:
```
# curl -sSfIL my.domain.org/nextcloud | grep Location:
Location: https://my.domain.org/nextcloud
Location: /nextcloud/
Location: https://my.domain.org/nextcloud/login
```
1. Forced HTTPS redirect by webserver
2. Apache2 mod_dir redirects with an automatically appended trailing slash, with path only
3. Nextcloud redirects to login page with full URL
_____________________
So it seems to be not uncommon to not rely on unverified HTTP_HOST and forge absolute URLs not only for POST/CLI requests but for redirects as well. I hence close this PR an will apply reasonable defaults for the webservers server name and/or Ampaches "http_host" setting instead, most likely the local network IP with documented hint that it needs to be adjusted to allow proper remote access.

PR #2672, opened 2020-12-28
Album display tweaks

mitchray:
Move album outline to :before to fix visual glitch from backdrop-blur being on :after; height 100% for album link needed in case other albums are taller; restoring backdrop-blur @supports as latest Firefox is still behind flag and it affects readability


mitchray:
Anything blocking this from being added?

AshotN:
Does #2706 block this?

mitchray:
Absorbed by #2706

PR #2673, opened 2020-12-30
Custom scrollbar colors

mitchray:
Custom scrollbar styling, before & after

scrollbars

### DIFF
--- a/client/stylus/main.styl
+++ b/client/stylus/main.styl
@@ -26,6 +26,27 @@ a
   &.disabled
    pointer-events none
 
+/* W3C spec */
+* {
+  scrollbar-width: thin;
+  scrollbar-color: colorAmber900 colorGray900;
+}
+
+/* Chrome, Edge, and Safari */
+*::-webkit-scrollbar {
+  width: 12px;
+}
+
+*::-webkit-scrollbar-track {
+  background: colorGray900;
+}
+
+*::-webkit-scrollbar-thumb {
+  background-color: colorAmber900;
+  border-radius: 20px;
+  border: 3px solid colorGray900;
+}
+
 :global
   //Pages
   @import "./pages/Playlists"
--- a/client/stylus/main.styl
+++ b/client/stylus/main.styl
@@ -33,15 +33,15 @@ a
 }
 
 /* Chrome, Edge, and Safari */
-*::-webkit-scrollbar {
+::-webkit-scrollbar {
   width: 12px;
 }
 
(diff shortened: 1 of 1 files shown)


client/stylus/main.styl
```diff
@@ -26,6 +26,27 @@ a
   &.disabled
    pointer-events none
 
+/* W3C spec */
+* {
```
AshotN, line 30: I do not believe you need `*` scope for standard.

client/stylus/main.styl
```diff
@@ -26,6 +26,27 @@ a
...
+  scrollbar-width: thin;
+  scrollbar-color: colorAmber900 colorGray900;
+}
+
+/* Chrome, Edge, and Safari */
+*::-webkit-scrollbar {
```
AshotN, line 36: You do not need any scope for WebKit style.

AshotN (changes requested):
Change the scoping for standard compliant to `body`, remove all scoping for webkit style.

mitchray:
On my Windows Firefox 84 the scrollbar-width isn't affecting anything unless its globally `*` , even if I apply it to the elements which have overflow. Could just be something weird on my end but I've left it for now.

AshotN:
> On my Windows Firefox 84 the scrollbar-width isn't affecting anything unless its globally `*` , even if I apply it to the elements which have overflow. Could just be something weird on my end but I've left it for now.

I think I noticed that as well. I’m pretty against scoping with `*`. So we should look into this before this gets merged. 

I do like the custom scroll bars.

mitchray:
If its a performance concern using the ```*``` then I'd say fear not, modern browsers handle it very well

https://www.telerik.com/blogs/css-tip-star-selector-not-that-bad
https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/#section4

Alternatively I can look into 3rd party scrollbar replacements

AshotN (approved):

AshotN:
Maybe I have an irrational fear of the `*` selector. We can always revisit this once it becomes more standardized.

kuzi-moto:
I don't believe it's an irrational fear. I believe it is against general best practices not to use it. The biggest one being performance, since it gets every single element.

AshotN:
> I don't believe it's an irrational fear. I believe it is against general best practices not to use it. The biggest one being performance, since it gets every single element.

The articles he listed seemed to argue otherwise, I would like to dive deeper into it.

Issue #2666, opened 2020-12-26
Minimum targeted browsers?


AshotN:
Do we have any specified minimum required browser versions that we intend to support. A lot of web technologies don't support old browsers, and while a lot of poly-fills exist, it increases the bundle size for everybody and sometimes is a pain.

usox:
Version 5 could be used to make a cut...

AshotN:
That was my thought as well.

lachlan-00:
If we can keep a major release to every 12ish months I would think something like 2/3years back?

I really only use Firefox so I'm definitely not an expert of browser support.

AshotN:
As long as we don't intend to support IE for example, that'll help speed things up.

But it be nice if we decided something like >1% of usage, or certain browser version.

wagnered:
I found a [reference](https://en.m.wikipedia.org/wiki/HTML5test) to a   browser HTML5 test suite and in the article, were references to other tests sites for tests such as Cascading Style Sheets, ECMAScript, Scalable Vector Graphics, or the Document Object Model.

From the results ampache.org could have a recommended browser.

As microsoft moved to the. Edge browser, which now uses  the blink engine,  we can pretty much ignore the IE series.  My rational is that  a user would want the best (read 'latest') browser to experience Ampache.

AshotN:
The reason I want us to have a minimum supported browser list is so we test with those browsers. Even modern Chrome and Firefox sometimes render things differently, once you start adding in older version of all the browsers. You can get some ridiculous inconsistencies.

usox:
Our policy at work is to support browsers 3 years back. Most of the people use modern browsers - companies with strict software policies and long evaluation cycles are the major issue here. But I guess, these are not our main targetgroup :)

lachlan-00:
Trident/ie and non-blink edge are out of what we should care about for sure

AshotN:
I’m good with a 3 year cycle. If we can decide on what specific versions that entails for Chrome/blink and Firefox/gecko(not many browser engines left these days sadly). I can close the issue and figure out how to test that.

mitchray:
Closing as it looks like a consensus was reached 👍

PR #2675, opened 2020-12-31
Offline webfonts

mitchray:
Offline webfonts per https://github.com/ampache/ampache/discussions/2669 and https://github.com/ampache/ampache/pull/2650#discussion_r549054023, modern browser support only per https://github.com/ampache/ampache/issues/2666


AshotN (approved):
Can’t test right now. But I assume you did :). Thank you!

PR #2683, opened 2021-01-07
Albums page

theTendo:
Implented the albums page following the same structure used for artists.

Before:

albums_page_before

After:

albums_page_after


AshotN:
Looks about right. I'll merge this and check it out later. Thank you!

PR #2686, opened 2021-01-10
Play mode selector skeleton [react-client]

mitchray:
Skeleton component for switching between play modes, where each mode has its own queue/interface displayed underneath (not implemented)

I've added short help text below each mode to help decipher how they differ from each other

firefox_CDeyYxBZGH


client/components/PlayModeSelector/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+            enabled: false
+          },
+        stream: {
+            name: 'Stream',
+            description: 'Generate a playlist file for use in another player',
+            enabled: false
+        },
+    };
+
+    const [currentPlayMode, setCurrentPlayMode] = useState('web');
```
AshotN, line 30: Avoid using strings, use enums

AshotN:
The enum/string issue is still blocking this from being merged.

mitchray:
Withdrawing

PR #2687, opened 2021-01-10
R128 gain adjustment [Ampache Reborn]

mitchray:
This adds [EBU R 128 gain adjustment](https://tech.ebu.ch/publications/r128s2) similar to replaygain, and tweaks how the values are stored in database.

- [x] Add R128 fields to tables (Ampache.sql)
- [x] Missing tags should be saved as null in database instead of 0 (as zero is valid) 
- [x] [getID3](https://github.com/JamesHeinrich/getID3) doesn't return [values of 0](https://github.com/ampache/ampache/issues/2687#issuecomment-760159349_), [PR accepted](https://github.com/JamesHeinrich/getID3/pull/284)
- [x] Note for pull, needs update.class.php update


mitchray:
2nd attempt at PR as I borked the first trying to update it from develop

mitchray:
Not ready for review sorry, pressed wrong button, not having a good day

lachlan-00:
one thing i see is that vainfo will read the tags but only create array values if the tag exists.

so in the ```foreach ($id3v2['TXXX'] as $txxx)``` it will create a ```$parsed['replaygain_album_peak']```

Instead of an empty array it might be goo to include all the mandatory/expected fields in the top and overwrite them if parsed?
```
    private function _cleanup_id3v2($tags)
    {
        $parsed = array();
```

so then parsed is a bigger array with all our defaults in it

lachlan-00:
i also deleted data where 
```
replaygain_track_gain = 0 AND replaygain_track_peak = 0 AND replaygain_album_gain = 0 AND replaygain_album_peak = 0
```
and then did an update from tags and the rows are null. and not 0'd out.

It night be good for clean actions to null out these columns?

lib/class/vainfo.class.php
```diff
@@ -449,10 +450,12 @@ public static function clean_tag_info($results, $keys, $filename = null)
 
             $info['lyrics']    = $info['lyrics']
                     ?: strip_tags(nl2br((string) $tags['lyrics']), "<br>");
-            $info['replaygain_track_gain'] = $info['replaygain_track_gain'] ?: floatval($tags['replaygain_track_gain']);
-            $info['replaygain_track_peak'] = $info['replaygain_track_peak'] ?: floatval($tags['replaygain_track_peak']);
-            $info['replaygain_album_gain'] = $info['replaygain_album_gain'] ?: floatval($tags['replaygain_album_gain']);
-            $info['replaygain_album_peak'] = $info['replaygain_album_peak'] ?: floatval($tags['replaygain_album_peak']);
+            $info['replaygain_track_gain'] = $info['replaygain_track_gain'] ?: $tags['replaygain_track_gain'];
```
lachlan-00, line 453: it's possibly in the ternary's here.

null and 0 will count as false here and the ?: means if it's not true it will be null so it's probably skipping 0's here

mitchray: To be honest I don't understand that clean_tag_info function...

If its initialising $info as an empty array, how can the ternary evaluate to anything other than false? Argh! lol

With that said I've already tried wrapping them in ```!is_null```, ```!empty```, ```isset``` and they all seem to result in the values coming back null (or 1...)

Here is my log I kept as I delved into the depths printing out before and after debug lines trying to pinpoint where the 0 is turning to null...

```
local.catalog.php insert_local_song()

seems to be happening in $vainfo->get_info(); in local.catalog.php -> insert_local_song()

back in vainfo... _get_tags

_parse_general the culprit?

clean_tag_info...?

definitely in $vainfo->get_info(); somewhere!

DEFINITELY 100% in _get_tags
```

lachlan-00: which array has the 0? infor or tags?
$info['replaygain_track_gain']
$tags['replaygain_track_gain']

mitchray: Neither does, that's the problem..

Let me set up a demonstration somehow

mitchray:
I managed to get xdebug installed

https://github.com/ampache/ampache/blob/d46dc2428d1e3908ea043e1d785c1d3d4ef04af1/lib/class/vainfo.class.php#L110

Test file has ```r128_album_gain = 99``` and ```r128_track_gain = 0``` but the getid3 library isn't returning r128_track_gain in the array

```
$this->_raw["tags"]["vorbiscomment"]
  'r128_album_gain' => 
  array (
    0 => '99',
  )
```

lachlan-00:
that's good, i can probably dig from there.

lachlan-00:
it could be that the version of getid3 we are using doesn't pull the 0? https://github.com/JamesHeinrich/getID3

wagnered:
@mitchray  

You wrote:

> Missing tags should be saved as null in database instead of 0 (as zero is valid)

According to [MySQL specs](https://dev.mysql.com/doc/refman/5.6/en/data-type-defaults.html#data-type-defaults-implicit). If a default isn't defined when the column is created the default value for an integer column implicitly becomes 0.  I'm fairly sure that's your problem.

mitchray:
@wagnered 

Even before this branch the tables were defined with DEFAULT NULL but were still being saved as 0.0000 floats. This was due to the values being cleansed with floatval(), which for both ```null``` and ```0``` resulted in ```0.0000```.

After removing the floatval() on the replaygain parsing the values are either saved or left as NULL, checkbox ticked, all good there.

So database side of things, nothing more to see here.

EXCEPT the issue now is that files with a replaygain tags of ```0``` are not having those tags returned to us from the **initial file scan**, an issue I believe was always present but masked by that floatval() parsing.

wagnered:
I would like to suggest that lib/class/update.class.php be used to modify Ampache tables instead of Ampache.sql. By using the update class, Ampache can track the updates and can detect what updates need to be applied at install or at a later time.

lachlan-00:
> I would like to suggest that lib/class/update.class.php be used to modify Ampache tables instead of Ampache.sql. By using the update class, Ampache can track the updates and can detect what updates need to be applied at install or at a later time.

yeah, i will update the sql file as part of the release process

mitchray:
@lachlan-00 You are correct sir, I found the exact point where its happening; its evals to false with "0" and skips the array push

https://github.com/JamesHeinrich/getID3/blob/ec1762d8f8fd44a29e8d894a7000fbfdfb422251/getid3/getid3.php#L1498

Mystery solved! So what should I do?

lachlan-00:
> @lachlan-00 You are correct sir, I found the exact point where its happening; it evals to false with "0" and skips the array push
> 
> https://github.com/JamesHeinrich/getID3/blob/ec1762d8f8fd44a29e8d894a7000fbfdfb422251/getid3/getid3.php#L1498
> 
> Mystery solved! So what should I do? Do we have another method of ascertaining this info or should I raise an issue over at the getID3 repo?
> 

Bug report it and then pair them up?
If r128_album_gain and !isset r128_track_gain

Something like that (on phone)

lachlan-00:
If you haven't noticed in the code. I **love** ternary operations

mitchray:
The fix in getID3 has been accepted so I believe this is now ready after a composer update 

I think we can leave out nulling those 4 replaygain fields as it effectively makes no difference, either way no gain will be applied and a catalog/tag update will bring in any new values

lachlan-00:
We're using dev master so I'll include it in this on Monday

lachlan-00:
i'm at work today so on windows here. i'll update the composer deps and merge this one tomorrow when i WFH

lachlan-00:
forgot the json_data class, almost done

lachlan-00:
k, database update and code tested, it's done :)

mitchray:
Awesome! Thanks for your help on this

PR #2691, opened 2021-01-13
Updated _clean_vorbiscomments($tags) for changes to Picard

wagnered:
I believe the required changes can be contributed to both changes to Picard'changes in tag mapping and getID3' changes in gathering the tags.

### DIFF
--- a/public/templates/header.inc.php
+++ b/public/templates/header.inc.php
@@ -74,9 +74,7 @@ $ajaxUriRetriever = $dic->get(AjaxUriRetrieverInterface::class);
 
         <?php require_once Ui::find_template('stylesheets.inc.php'); ?>
 
-        <link rel="stylesheet" href="<?php echo Ui::find_template('jquery-editdialog.css', true); ?>" type="text/css" media="screen" />
         <link rel="stylesheet" href="<?php echo $web_path; ?>/lib/modules/jquery-ui-ampache/jquery-ui.min.css" type="text/css" media="screen" />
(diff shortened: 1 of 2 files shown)


lachlan-00:
should we keep the old and new cases? or is this how it outputs the txxx now?

wagnered:
TXXX is id3v2.3, this is vorbiscomments which is typically used in ogg(flac and opus) containers.

I'm thinking it  is dependent on the getID3 library.  So I'm not sure if the `_clean_vorbiscomments($tags)` function really did  work.

As for TXXX, the  MUSICBRAINZ_TRACKID is actually the music brains **recording** ID.  Is it used anywhere in Ampache? I question why that is gathered, instead of the Musicbrainz **release** track Id.  I think the release track Id might be more consistently used across the various taggers that incorporate the Musicbrainz identifiers. 

Also, Ampache, tests for the UFID tag to gather the MUSICBRAINZ_TRACKID frame.  If Picard Is the only tagger that includes the UFID frame, that might be problematic.

lachlan-00:
i'll clean some file tags and redo them in picard to test

lachlan-00:
which version of picard does this? i've got 2.5.6 which seems to be the latest and the old strings without space are coming through.

lachlan-00:
for release track id i only see the track id. is it recording id?
image

wagnered:
I have found two needed changes to the header.inc.php that didn't happen when source-changes migrated to the new develop branch.

I am going to close this PR and do the fix from this branch.

Issue #2692, opened 2021-01-13
Subsonic API "f" parameter should default to xml if unspecified


tristantarrant:
### Describe the bug
The subsonic endpoint does not respect the API for the `f` parameter by not supplying a default value for it. The DSub Android client, for example, does not explicitly set `f`, expecting the default xml format to be used.

f | No | xml | Request data to be returned in this format. Supported values are "xml", "json" (since 1.4.0) and "jsonp" (since 1.6.0). If                 using jsonp, specify name of javascript callback function using a callback                 parameter.
-- | -- | -- | --

### To reproduce

`curl http://localhost/ampache/rest/ping.view`

### Expected behavior
I should get an error response formatted as XML:
```
<?xml version="1.0" encoding="UTF-8"?>
<subsonic-response xmlns="http://subsonic.org/restapi" status="failed" version="1.13.0">
  <error code="10" message="Missing Subsonic base parameters"/>
</subsonic-response>
```
Instead I get a blank response

## Environment

* Ampache version: 4.2.5
* Web server + version: Apache
* Server operating system: Ubuntu 20.10
* Client operating system: Android

### Client type

* SubSonic

lachlan-00:
merged.

PR #2693, opened 2021-01-13
The default format for the Subsonic API should be XML

tristantarrant:
https://github.com/ampache/ampache/issues/2692


lachlan-00:
this is a great find. it explains why i've been getting errors for only xml where returns weren't coming in.

Another new contributor!!!

PR #2720, opened 2021-01-31
Generate thumbnails at 2x for high DPI output

mitchray:
Output all thumbnails at @2x to improve fidelity, _especially_ on hi-dpi screens. Downscaling also disguises compression artifacts on @1x screens so everyone wins.

I'm making the assumption that the slight increase in storage is an acceptable tradeoff.


usox:
I took just a quick look at the places which use `Art::get_thumb_size`. There is one other usage (\Ampache\Module\Application\Image\ShowAction) which passes the size to Art::get_thumb which generates and saves a new thumbnail. This thumbnails are used in other places (subsonic api for instance) - is it possible, this change may cause problems at those places?

mitchray:
I did wonder that too but Ultrasonic and Dsub appear to be handling it fine. 

I've been using this on my live server for the past couple of days and been all good so far.

usox:
My main concern is that we can no longer distinguish between the thumbnails which have been created before and after this change. But I don't know it that's a point, so...

@lachlan-00 Any final thoughts?

lachlan-00:
Need to check this in more detail tomorrow.

xekon:
Let me know if you need another tester :)

lachlan-00:
okay so the tldr is this is generating the thumbs at a larger size in the DB to have a better looking normal size on output?

That would probably require the removal of all existing thumbs except the size 'original' (if someone wanted to regenerate these)

when i look at my image table i have 50000 original images out of about 143k. so if we wanted to regenerate thumbs or if someone wanted to update their cache it would need the ability to clear !original images

clean_art_table.inc/cli should be extended to clear the generated thumbs too;
possibly a rebuild option as well but the rebuild i'm not so sure about cause i have wild resolutions in my table
image

mitchray:
I guess I'm not understanding why old ones need to be flushed...

In my testing the old thumbnails are being regenerated as they're called, I'm not clearing any caches or manually gathering artwork. Are you experiencing different?

I believe the sizes reported in the database won't change as this now refers to the _display_ size and not the true dimensions

[From the defined sizes](https://github.com/ampache/ampache/blob/f2329b6b7b7da883dfb1a8402850ad1749819776/src/Model/Art.php#L1187-L1236) only the highlighted ones are thumbnails and the rest are original artwork sizes

106960156-62288280-6787-11eb-8c8b-c0e4254820d4

lachlan-00:
i've been running this one for a few days, i'm not bothered as much now lets put it in.

PR #2730, opened 2021-02-04
Added Ajax::button_with_text to have img and text in single link

mitchray:
I noticed quite a few action items had separate links for the img and text, or links were missing from either one.

Now all actions (#information_actions) have both img and text inside a single ```<a>```


lachlan-00:
happy with this one, the separation of links and images is probably a leftover of before there were both in the templates. (really old)

PR #2731, opened 2021-02-05
Maintain column widths during table sort

mitchray:
Before
firefox_bQ8OA8zLk7

After
firefox_RZna7u8GKz


lachlan-00:
took a bit to find an album that would do it but it looks better now.

PR #2732, opened 2021-02-05
Prevent page jump on View click

mitchray:
Also add href to the webplayer actions that didn't have them, just for consistency


lachlan-00:
i like that a lot, the jumping in the ui can seem very random

Issue #2799, opened 2021-03-05
[Feature Request] Ability to use albumartist ID3 instead of artist


SaintDubious:
## Describe what you don't like about Ampache, or what it is missing

I've always been frustrated by the "artist" tag on an MP3 file, specifically when there's more then one artist on a song. When I try to list my collection by artist, I end up with a handful of entries for the same artist, it adds clutter. It's even worse when you have a couple of movie soundtracks and then suddenly the artist list contains a whole bunch of artists you've never heard of.

I'd like for there to be a feature that allows me to use the "albumartist" tag instead of the "artist" tag. This way I can keep the correct information on the file itself, but not deal with all the clutter when looking at my music collection

## Describe how you would like the feature to work**

I'd like for there to be a new boolean setting that allows me to use albumartist instead of artist when building a new catalogue. 

I actually already wrote all of this, but when I went to submit the PR I saw that I had written against the master branch, which is a totally different universe from the develop branch. Then I checked the CONTRIBUTING.md and saw that maybe I should have written this first. I'm happy to submit the PR against the 4.x build if that's possible. Please Advise.

SaintDubious:
I noticed that other folks offer PR against master, so I went ahead and created the PR.

lachlan-00:
i'm not sure if i'll be putting more change into 4.4 branches but there are a few removals in the database i need to do so master is fine.

SePHTaN:
> and then suddenly the artist list contains a whole bunch of artists you've never heard of.
That's one thing i don't like also.
But my opinion is that all the data belonging to songs, albums and artists should go to the database without "malforming" them.
What you want to have is also achievable by not showing songartists in the view but album-artists.

There is always a way to show what you want from the data you have, but no way to show what you don't have.
It makes me a little bit scared seeing only the individual wishes of last months to tweak the contents of the database so it reflects what one wants to see, instead of making the database as extensive as possible and leave the rest to tweakable viewing abilities of ampache.

Why i am scared:
I just ran an update to my database cause i exchanged some of my albums with reencoded versions with higher bitrate.
Now i have several artists put together ("Die Toten Hosen" and "Die Roten Rosen" for example) that were correct separated before. All Albums of "The Toten Hosen" are now from "The Rote Rosen" except the one album i updated.
Also "The Toten Hosen" are missing musicbrainz-ids in the database i suspect having to do with an Feature Request just some weeks ago.
[...]

PR #2800, opened 2021-03-05
Prefer albumartist over artist

SaintDubious:
See issue: https://github.com/ampache/ampache/issues/2799

When building a new local catalog add the ability to use the albumartist tag instead of the artist tag.

### DIFF
--- a/lib/class/preference.class.php
+++ b/lib/class/preference.class.php
@@ -134,7 +134,7 @@ class Preference extends database_object
         'webplayer_confirmclose',
         'webplayer_flash',
         'webplayer_html5',
-        'webplayer_pausetabs'        
+        'webplayer_pausetabs'
     );
 
     /**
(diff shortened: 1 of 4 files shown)


lachlan-00:
ahh, interesting. so if you tag songs as `X feat. y` but the album is `x` then all the songs would join under that artist? that's actually not a bad way of doing it.

lachlan-00:
I think you've done a really good job in picking up all the things needed.

Because this is only on insert i don't think it should affect too much but we need to check some things

* does `update from tags` replace the album artist or recreate as a new id?
* if single disk is okay does verify on the catalog work okay too?
* this should also replace mbid (albumartist_mbid) because they might not be the same

SaintDubious:
@SePHTaN raised some valid concerns with this approach in the Issue thread. It might make more sense to close this PR and pursue a more thorough solution for the "develop" branch. I'm happy to apply this as a simple hack to my own server just so I can get the behavior I want for my Android Subsonic API Client.

I'll leave this open for now. If people argue that they want it then I'll pursue answers to the questions @lachlan-00 raised.

Also, @lachlan-00 , thanks for the quick reply and thoughtful comments. I appreciate the fast consideration.

lachlan-00:
closing as not ideal and will work in the open issue instead

Issue #2828, opened 2021-03-17
Incorrect decimal value: '' for column replaygain_track_gain in song_data


jleipsig:
## Description
When I add an album, the ampache logfile mentions Incorrect decimal value: '' for column replaygain_track_gain
This only happens when enable_custom_metadata = "true" in my config

### Describe the bug
In `Module\Util\VaInfo.php::clean_tag_info` `$info['replaygain_track_gain']` is set to null. When `enable_custom_metadata` is enabled, this null value is set to an empty string, because `!isset($info[$tag])` gives true for null values.

## Environment

* Ampache version: 5.0
* Web server + version: Debian 10.8

## Logs
(Ampache\Module\System\Dba) -> Error_query SQL: INSERT INTO `song_data` (`song_id`, `comment`, `lyrics`, `label`, `language`, `replaygain_track_gain`, `replaygain_track_peak`, `replaygain_album_gain`, `replaygain_album_peak`, `r128_track_gain`, `r128_album_gain`) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
(Ampache\Module\System\Dba) -> Error_query MSG: ["22007",1366,"Incorrect decimal value: '' for column `ampachedev`.`song_data`.`replaygain_track_gain` at row 1"]

usox:
So, fix is merged, therefor we close the issue :)

PR #2829, opened 2021-03-17
don't overwrite null value with empty string in metadata

jleipsig:
Don't overwrite null values with an empty string if custom metadata is enabled.  This fixes a problem with replaygain and r128 gain metadata.
Also removed a redundant check `is_array($value)` in ternary.


jleipsig:
fix for bug described in #2828

lachlan-00:
that's a good one, one of those "i will fix this one day" issues

Issue #2830, opened 2021-03-17
Call to undefined method Ampache\\Repository\\Model\\Metadata\\Model\\MetadataField::findById()


jleipsig:
## Description
When adding an album and `enable_custom_metadata = "true"`,  the metadata is not written to the metadata table and the apache log shows an error

### Describe the bug
`Ampache\Repository\Model\Metadata\Model\Metadata::$fieldClassRelations` returns Model\MetadataField instead of Repository\MetadataField

## Environment

* Ampache version: 5.0
* Web server + version:  Apache/2.4.38
* Server operating system: Debian 10.8

### Client type

* Web

## Logs
'PHP message: PHP Fatal error:  Uncaught Error: Call to undefined method Ampache\\Repository\\Model\\Metadata\\Model\\MetadataField::findById() in home/dev/www/ampache/src/Repository/Model/DatabaseObject.php:125\nStack trace:\n#0 home/dev/www/ampache/src/Repository/Model/DatabaseObject.php(47): Ampache\\Repository\\Model\\DatabaseObject->initializeChildObjects()\n#1 home/dev/www/ampache/src/Repository/Model/Metadata/Metadata.php(89): Ampache\\Repository\\Model\\DatabaseObject->__construct()\n#2 home/dev/www/ampache/src/Repository/Model/Catalog.php(2075): Ampache\\Repository\\Model\\Song->addMetadata()\n#3 home/dev/www/ampache/src/Module/Catalog/Catalog_local.php(873): Ampache\\Repository\\Model\\Catalog::add_metadata()\n#4 home/dev/www/ampache/src/Module/Catalog/Catalog_local.php(426): Ampache\\Module\\Catalog\\Catalog_local->insert_local_song()\n#5 home/dev/www/ampache/src/Module/Catalog/Catalog_local.php(294): Ampache\\Module\\Catalog\\Catalog_local->add_file()\n#6 home/dev/www/ampache/src/Module/Catalog/Catalo...',

PR #2831, opened 2021-03-17
Give correct MetadataField in Ampache\Repository\Model\Metadata\Model\Metadata

jleipsig:
Fixes issue #2830

Also a correction to write only the classname (e.g. 'Song') into table field `metadata.type`, and not the classname with the full namespace.

This also solves a infinite loop when updating the catalog with custom metadata enabled

### DIFF
--- a/src/Repository/Model/Metadata/Metadata.php
+++ b/src/Repository/Model/Metadata/Metadata.php
@@ -24,6 +24,7 @@ namespace Ampache\Repository\Model\Metadata;
 use Ampache\Config\AmpConfig;
 use Ampache\Repository\Model\Metadata\Model\MetadataField;
 use ReflectionException;
+use Ampache\Module\Util\ObjectTypeToClassNameMapper;
 
 trait Metadata
 {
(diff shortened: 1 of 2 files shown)


lachlan-00:
take the lead on these 2 @usox i'll be back tues

usox:
Hi @jleipsig 

Thanks for repairing this issue. Could you please use the static method `ObjectTypeToClassNameMapper::reverseMap` instead of using reflection to get the desired media object type?
The reason for this is that we'd like to avoid using the php class names as media object types. This will allow us to change the php class name in future without causing problems like the one you fixed in here.
Once we've identified/refactored every location, we can use other techniques to retrieve the object type.

jleipsig:
Hi @usox, thanks for your explanation. I changed Reflection to `ObjectTypeToClassNameMapper::reversedMap`. Because the first uppercase in existing data in `metadata.type`, `ucfirst()` is added.

usox:
> Hi @usox, thanks for your explanation. I changed Reflection to `ObjectTypeToClassNameMapper::reversedMap`. Because the first uppercase in existing data in `metadata.type`, `ucfirst()` is added.

Makes sense. Maybe, we will migrate the db values some day to drop this conversion. But sadly I'm not able to see any changes in the code - maybe a mistake during the force push?

jleipsig:
Sorry, did something wrong there. The changes should be visible now.

usox:
@jleipsig Yes, indeed - excellent :)

usox (approved):

usox:
Thanks!

PR #2836, opened 2021-03-20
Two CSS files for jquery dialogs weren't being found.

wagnered:
`Ui:-:find_template` couldn't find them, so directly used `$web_path` and moved them to "stylesheets.inc,php".

### DIFF
--- a/public/templates/header.inc.php
+++ b/public/templates/header.inc.php
@@ -74,9 +74,7 @@ $ajaxUriRetriever = $dic->get(AjaxUriRetrieverInterface::class);
 
         <?php require_once Ui::find_template('stylesheets.inc.php'); ?>
 
-        <link rel="stylesheet" href="<?php echo Ui::find_template('jquery-editdialog.css', true); ?>" type="text/css" media="screen" />
         <link rel="stylesheet" href="<?php echo $web_path; ?>/lib/modules/jquery-ui-ampache/jquery-ui.min.css" type="text/css" media="screen" />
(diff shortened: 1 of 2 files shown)


usox:
Hi @wagnered 
Thanks, I figured this out a week ago but was a bit unsure about the best way to get this fixed. Adding them to the stylesheet file looks like a good idea - does not make sense to call find_template for those css files.

SePHTaN:
> `Ui:-:find_template` couldn't find them, so directly used `$web_path` and moved them to "stylesheets.inc,php".

In my opinion, CSS files should not be hard-coded if ampache should still be able to be styled with themes.
According to `/src/Module/Util/Ui.php`:
```
    public static function find_template($template, bool $extern = false)
    {
        $path      = AmpConfig::get('theme_path') . '/templates/' . $template;
        $realpath  = __DIR__ . '/../../../public/' . $path;
        $extension = strtolower(pathinfo($path, PATHINFO_EXTENSION));
        if (($extension != 'php' || AmpConfig::get('allow_php_themes')) && file_exists($realpath) && is_file($realpath)) {
            return $path;
        } else {
            if ($extern === true) {
                return '/templates/' . $template;
            }

            return __DIR__ . '/../../../public/templates/' . $template;
        }
    }
```
Ampache is searching Cascading Style Sheets in the themes dir, which is `public/themes/reborn`.
Though those two css files reside in `public/templates`, which leads me to the conclusion "theming ampache " needs more dokumentation regarding style sheets required, and maybe identical, for all themes.

usox:
@SePHTaN wow, this was simultaneously. Well I'm ok with this certain change because these css-files are crucial.

PR #2864, opened 2021-04-04
Make VaInfo MultiArtist ready

SePHTaN:
Add artists tag to results object

Move AmpConfig to ConfigContainer
Move debug_event to LoggerInterface


SePHTaN:
Looked at CodeClimate.
Besides any other issues i have overseen, i am just doing some CamelCase renaming of variables.
As reducing some if...else constructs with earlier returns.
Giving VaInfo a better score at least with Style and Clarity

usox:
In cases someone wonders: @SePHTaN asked me to take a look on the code before opening the PR. We discussed some parts in our native language which now unexpectedly show up in the PR.

lachlan-00:
I have a bit of work on but am getting back onto things.

Code climate isn't a mandatory check but I find it's good to keep awareness up.

Thank you for working together on what you've been doing and I'm aware I need to speed up a bit more

SePHTaN:
@lachlan
what i contributed should be all safe for develop, beside the fix for the loading of metadata plugins.
these are the last changes i make directly in develop.
i created the feature branch multi-artist in my fork for the rest

lachlan-00:
changing the additional_genre_delimiters text value will need to be looked at.

Changing the template won't change the live value and will probably break on any config file updates as additional_genre_delimiters won't migrate to additional_delimiters when the version is bumped.

It would actually be better to make another config value for it and keep additional_genre_delimiters.

SePHTaN:
i understand your concerns
as the delimiters are the same for all tags i just leave the old config name.
a new/additional config parameter isn't needed, it though lacks a bit clarity from config-option name
but introducing a new config-option for exactly the same is nothing i would like to do.
i would prefer adding a bit more explanation in the config option comment declaring where and for what it is used

lachlan-00:
i just realised that the plugin issue i fixed in 752523647 will cause the conflict here as well. should be okay to update from develop?

I noticed that the details were never going to fire due to case
image

lachlan-00:
> i understand your concerns
> as the delimiters are the same for all tags i just leave the old config name.
> a new/additional config parameter isn't needed, it though lacks a bit clarity from config-option name
> but introducing a new config-option for exactly the same is nothing i would like to do.
> i would prefer adding a bit more explanation in the config option comment declaring where and for what it is used

keeping it is much better, renaming it should be looked at separately to function and i can work on that separately. we'll need to probably do that in the future anywat

lachlan-00:
put the filesize and merged from develop.

need to revert back to genre_delimiter as well

SePHTaN:
good to see that most of my pull_request made ot through with minor necessary changes.

astonishing enough that no user realised that some plugins were broken since around 30.Aug.2020.
If you like, i began some days ago a project on my fork, were i write down everything that comes to my mind regarding changes which i consider recommendable to do along with achieving multi-artist.
I would be happy if i receive feedback or if you have suggestions of your own

lachlan-00:
going to finish this off today if possible.

wagnered:
I'm not used to seeing so many commits apparently unrelated to the topic (feature) of the PR.  ziphandler?  system core changes?  Maybe I don't see the whole picture.  I would think best practices for a PR would also follow the [Single Responsibility Principle](http://en.wikipedia.org/wiki/Single_responsibility_principle) as the 5.* develop branch appears to follow. 

See [one developer's rational for better pull requests:](https://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/)

And if all the commits are somehow directly related, my apology.

lachlan-00:
> I'm not used to seeing so many commits apparently unrelated to the topic (feature) of the PR. ziphandler? system core changes? Maybe I don't see the whole picture. I would think best practices for a PR would also follow the [Single Responsibility Principle](http://en.wikipedia.org/wiki/Single_responsibility_principle) as the 5.* develop branch appears to follow.
> 
> See [one developer's rational for better pull requests:](https://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/)
> 
> And if all the commits are somehow directly related, my apology.

I think those are leftovers from when i squashed the other pull for @SePHTaN 's last pull.

lachlan-00 (approved):

lachlan-00:
As @wagnered noticed i use squash on pulls you might need to do a reset on your branch to get rid of those extra commits.

TODO after this

* safely rename additional_genre_delimiters (probably a config bump/helper function for renamed values)
* Add theaudioDB and other plugins to `bin/cli run:updateCatalog -u` they all have a lot of info that's pretty useful

SePHTaN:
> I'm not used to seeing so many commits apparently unrelated to the topic (feature) of the PR. ziphandler? system core changes? 

Sorry for this, it definitely wasn't my intention that every pull request got appended to the existing one.
As my apology I can only say that this is my first foray into the world of teambased opensource development and that I still have a lot to learn.

usox:
Don't worry, @SePHTaN  :smile:

wagnered:
This is not directed at @SePHTaN.

Before this merge, the tag writing failed.  I just decided to update from upstream and  noticed this added problem.  I found it by stepping through the code.  Perhaps there should be a policy of actually exercising the code before merging the PR.

In Vainfo.php:302, `!in_array($extension, $extensionMap)`  should have been `!array_key_exists($extension, $extensionMap)`.

@SePHTaN  I would like to suggest that you find some way of stepping through the code to verify it does what you expect it to do before submitting the PR.  I'm doing my development on an android tablet using vim and the vim-debug plugin.  I'm sure that on a full desktop, a number of  excellent IDE's exist. 

I will include this change when I am quite sure the PR is ready to be submitted.

SePHTaN:
@wagnered
sry for this one, 
you are actually right, and i hope it satisfies you when i tell you that i do verify wether the code i write is functional.
For this i have an ampache testserver where i test to see if it does what i expect, except for write_id3.
I must admit, due to my personal usecase, i didn't tested writing of tags. So this one slipped me through.

On the other side, this issue was easy to find cause functionality was/is totally broken for writing of tags.
From my tests in the last days for tag retrieval that VaInfo does or just doesn't do:
ogg files don't have replaygain_data, mb_trackid, mb_albumid, mb_albumid_group, mb_artistid, mb_albumartistid, artists isn't split
flac files are the same like ogg
wav files don't have albumartist, band, mb_trackid, mb_albumartistid isn't split
m4a files don't have replaygain_data
not fully examined the results yet as i had to build testfiles with multiple artists, artist_aliases for all audio formats, and i did the last retrieval tests late last night

so especially the missing retrieval of mb_trackid for 3 major fileformats, altough it is in the tags, is an severe issue because Musicbrainz plugin depends on mb_trackid as ampache itself.

wagnered:
That's good to hear and I apologize for making assumptions.

> so especially the missing retrieval of mb_trackid for 3 major fileformats, altough it is in the tags, 

I noticed that, but because nobody complained, I left it alone. 🙂

So is Vainfo now multi-artist ready?  and what does that  mean anyway? Does a discussion/issue exist in which I can get a better understanding of what you want this feature to do. For example, how will it affect writing song metadata and images to the file? Should there be an image for every artist?

Thanks for your endeavors.

SePHTaN:
Generally yes, most part of adding Artists in Artist.php is done.
Currently working on Catalog.php and Song.php, i haven't fully decided how to loop over the artist names and assigned artistid's.
So every song will get added for each artist ( 6 artist -> 6 song entries and six song-id's) in song_data i have to add joinphrase and joinorder to ensure to be able to reassemble artistnames in correct order.

During first test i stumpled over the missing mb_trackid and stopped work on multi-artist to investigate why mb_trackid is missing.
mb_trackid is needed to group those songs together to one song in sql query for browsing albums/artist and maybe ( i haven't thought of much yet) for searches and api things.

SePHTaN:
Images for artists is an good question. 
I want to add several tuning options for multi-artist:
enable_multiartist : allowing to stay with the old way ampache works ($dotrim option in SplitSlashedlist() in vainfo
enable_aliases       : make ampache aware of artist with an alias/pseudonym
group_aliases       : wether to allow pseudonyms as artist byitself or to group with the primary artist when browsing
Another option i am thinking about is to make it configurable wether only albumartists are viewed when browsing artists,
or allowing to see all artists as individual artists.
When all artists can be browsed then definitely an image for the artist should be available also an summary/bio/begin/end for the artist. 
Those data could also be viewed a bit more structured in an mixed table/grid layout for textual data like summary/bio ...as an fixed grid for images allowing an standardized schema for 2, 4, 6 images to view and automaticly resize.
For writing images to file tags the open question for me is how to determin which picture belongs to which artist.
There has to be an textual hint together with the image, not sure wether this is possible for all fileformats.

SePHTaN:
for multi-artist feature i have created the branch multi-artist in my fork of ampache.
Currently there is nothing to be seen, cause i am working only local.
But when i think multi-artist is in a presentable state i will commit all changes to that branch.
Everybody will then be able to clone and test it, and i would be happy if many do so.
I will be even happy for every finding of issues, better to find them before merging into what ever branch.

wagnered:
With all the one-to-many and many-to-one relationships, I'm guessing you need to create foreign keys for the artist/song/aliases/images relationships.  Otherwise searches will become unmanagable as they very nearly are now.

BTW: both the album and song tables have an 'mbid' column.  I'll take a look at the mb_trackid problem.

The id3.org provides an informal standard for storing pictures and other metadata in mp3 files.  When writing pictures to other file types, getID3 library uses the id3v2.3/4 standard. A file may contain any number of images, but only one with the same description.

SePHTaN:
i have looked into the mb_trackid problem, and it opened another question.
I did build my testfiles mainly from two albums and transcoded them from mp3 to wav, flac and ogg.
The tool i used is audacity, though i have looked into the tags with getid3 demo.basic.php i haven't realized that audacity wrongly named the tags, and espacially mb_trackid was swapped with mb_recordingid.
So i retagged the files with picard and deleted the catalog and added a new one just some minutes ago.
Haven't looked to deep into it yet but looked better.
for ogg totaldisc is wrong, replaygain values are missing, artists and mb_artistid only first element is in results array
thats what i have seen by a quick look

wagnered:
The problem with the tagger testing is that we have no control over which tagger is used.  We  can only specify which standard ampache conforms to and hold steady to it.   Musicbrainz specifies how it writes the id's to the various file types. 
I just looked at the manpages for opusenc. when writing pictures, the default description can be left empty. That isn't conforming to any standard.  From my experience, Easytag conforms best to the id3v2.3/4 specifications.

> how to treat wrongly named tags in audiofiles ? 

 Could you give an example: give details? file type, correct tag name?

Appears that you and others want to turn Ampache into a "Content Management System" (CMS).

SePHTaN:
To be able to test the different features i want to add, i constructed me a small audio catalog.
The requirements I made on this catalog were simple. A set of songs with artist aliases / pseudonyms, a set of songs with several artists per song, and a set with several artists per song as a sampler / compilation (Album Artist = Various Artists).
Each set contains 5 Songs, one song for every codec i want to be able to test (mp3, flac, ogg, wav, mp4).

The issues i have/had came just from the fact that i more or less assumed that audacity writes standard conform tags.
But that wasn't the fact what i since yesterday know. Even retagging all songs didn't helped fully because i didn't allowed Picard to delete the tags before rewriting them ending in lots of tags to be doubled in the files.
So i rebuild that catalog today completly from scratch, ensuring before rewriting the tags that there is nothing left as the pure audio.
After 3 crazy working days because customers literally stormed our warehouse because of the impending lockdown, I hope to finally get further tomorrow. Because I have free time.

Issue #2833, opened 2021-03-18
[BUG] Podcasts catalog update, duplicate entries.


xekon:
### Describe the bug
When the Podcasts catalog is updated, it adds a duplicate entry, I have seen this happen several times now, wanted to make sure it was not a weird one time occurrence.

When it happens there is no way to delete a duplicate in the GUI, so I have been doing it manually via SQL:

```
mysql
use ampache;
select * FROM podcast_episode ORDER BY id DESC LIMIT 1;
DELETE FROM podcast_episode ORDER BY id DESC LIMIT 1;
quit
```

### To reproduce

I have two subscribed Podcasts:
https://feeds.megaphone.fm/darknetdiaries
https://2.5admins.com/feed/podcast

I update the podcasts using a cron job, just like I added to the wiki page here: "Automatic Podcast Catalog Updates" https://github.com/ampache/ampache/wiki/Podcasts

to add it here as well, here is the file:

`cat /etc/cron.d/ampache`

```
#.---------------- minute (0 - 59)
#| .------------- hour (0 - 23)
#| |  .---------- day of month (1 - 31)
#| |  | .------- month (1 - 12) OR jan,feb,mar,apr ...
#| |  | | .---- day of week (0 - 6) (Sunday=0 or 7) OR sun,mon,tue,wed,thu,fri,sat
#| |  | | |
#* *  * * *  user      command
0 1   * * *  www-data  php /var/www/ampache/bin/catalog_update.inc -n podcasts -a -g -i >/dev/null 2>&1
```

### Expected behavior

I expect it to only create one entry for the new episode, but every time it creates two entries for a new episode.

### Screenshots

![2021-03-18_16-04](https://user-images.githubusercontent.com/106644/111708785-b6516880-8803-11eb-8e16-40c4b482ebc4.png)

## Environment

* Ampache version: Ampache 4.4.0-release
* Web server + version: nginx latest
* Server operating system: Ubuntu 20.04.4
* Client operating system: Ubuntu 20.04.4 (Firefox latest)

### Client type

* Web

## Settings

`diff -u /var/www/ampache/config/ampache.cfg.php.dist /var/www/ampache/config/ampache.cfg.php | grep -E "^\+" | grep -v "host" | grep -v "secret" | cut -c 2-`

```
config_version = 46
http_port = "443"
web_path = ""
database_port = ""
database_name = "ampache"
database_username = "www-data"
database_password = ""
database_charset = "utf8mb4"
[...]

PR #2883, opened 2021-04-14
Systemd service and timer to update podcasts

stebe:
I was following issue #2833 and noticed that there are no examples of systemd service and timers.  Here they are.


usox:
Thanks @stebe. This a nice workaround to avoid problems like #2833

docs/examples/ampache_podcasts.service
```diff
@@ -0,0 +1,17 @@
...
+Documentation=https://github.com/ampache/ampache/wiki
+
+[Service] 
+PrivateTmp=true 
+KillMode=mixed
+Type=oneshot
+User=www-data
+Group=www-data
+ExecStart= php bin/cli run:updateCatalog podcasts -a -g -i
+WorkingDirectory=/var/www/ampache
```
usox, line 13: Maybe this should match the WorkingDir in ampache_cron.service

stebe: You are correct. I took it from my own setup

usox (approved):

lachlan-00:
love docs and examples!

PR #2890, opened 2021-04-19
fix plugin metadata not in results

SePHTaN:
another fix to tag sources case, causing plugin metadata not in results
convert all tagsources to lowercase, also occurrences of 'getID3'
only loop over plugins in _get_plugin_tags
return artists as array, don't prepare metadata outside VaInfo
HINT: replaygain metadata partly broken

### DIFF
--- a/src/Module/Util/VaInfo.php
+++ b/src/Module/Util/VaInfo.php
@@ -1643,7 +1643,7 @@ final class VaInfo implements VaInfoInterface
             if ($items === false) {
                 throw new Exception('Pattern given in additional_genre_delimiters is not functional. Please ensure is it a valid regex (delimiter ~)');
             }
-        $data = $items;
+            $data = $items;
         }
         if ((isset($data) && isset($data[0])) && $doTrim) {
             return $data[0];
--- a/src/Module/Util/VaInfo.php
+++ b/src/Module/Util/VaInfo.php
@@ -402,7 +402,7 @@ final class VaInfo implements VaInfoInterface
             'getid3_tag_order' => static::getConfigContainer()->get(ConfigurationKeyEnum::GETID3_TAG_ORDER)
         ];
 
-        $order = array_map('strtolower', $tagorderMap[$configKey] ?? '');
+        $order = array_map('strtolower', $tagorderMap[$configKey] ?? []);
 
         // Iterate through the defined key order adding them to an ordered array.
         $returned_keys = array();
@@ -761,7 +761,7 @@ final class VaInfo implements VaInfoInterface
         ];
 
         // convert to lower case to be sure it matches plugin names in Ampache\Plugin\PluginEnum
-        return array_map('strtolower', $tagorderMap[$this->get_metadata_order_key()] ?? '');
+        return array_map('strtolower', $tagorderMap[$this->get_metadata_order_key()] ?? []);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Module/Util/VaInfo.php
```diff
@@ -401,7 +402,7 @@ public static function get_tag_type($results, $configKey = 'metadata_order')
             'getid3_tag_order' => static::getConfigContainer()->get(ConfigurationKeyEnum::GETID3_TAG_ORDER)
         ];
 
-        $order = (array) ($tagorderMap[$configKey] ?? '');
+        $order = array_map('strtolower', $tagorderMap[$configKey] ?? '');
```
lachlan-00, line 405: for array_map should the empty string be an empty array instead?

src/Module/Util/VaInfo.php
```diff
@@ -401,7 +402,7 @@ public static function get_tag_type($results, $configKey = 'metadata_order')
             'getid3_tag_order' => static::getConfigContainer()->get(ConfigurationKeyEnum::GETID3_TAG_ORDER)
```
lachlan-00, line 402: does static::getConfigContainer()->get return arrays or strings?

lachlan-00:
just these issues i can see in scrutinizer
https://scrutinizer-ci.com/g/ampache/ampache/inspections/3de017b3-e20c-492f-a457-5d14bb28b313/issues/files/src/Module/Util/VaInfo.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

src/Module/Util/VaInfo.php
usox: @lachlan-00 Should be the same as AmpConfig::get.

src/Module/Util/VaInfo.php
```diff
@@ -757,7 +760,8 @@ private function get_metadata_order()
...
         ];
 
-        return (array) ($tagorderMap[$this->get_metadata_order_key()] ?? '');
+        // convert to lower case to be sure it matches plugin names in Ampache\Plugin\PluginEnum
+        return array_map('strtolower', $tagorderMap[$this->get_metadata_order_key()] ?? '');
```
usox, line 764: Any chance for multibyte characters? mb_strtolower would be better then

src/Module/Util/VaInfo.php
SePHTaN: interesting finding even for me who produced it
$configKey can only be one of the three : metadata_order, metadata_order_video, getid_tag_order
these are hardcoded so if nobody produces a typo there is no chance $tagorderMap[$configKey] is undefined
$order = array_map('strtolower', $tagorderMap[$configKey]); should be sufficient
the other way 
$order = array_map('strtolower', $tagorderMap[$configKey] ?? []);
would presume that something might go wrong and than additional checks would be advidsed

src/Module/Util/VaInfo.php
SePHTaN: both return arrays:
example for metadata_oder
$order : array (
  0 => 'getid3',
  1 => 'musicbrainz',
  2 => 'theaudiodb',
  3 => 'filename',
)

src/Module/Util/VaInfo.php
usox: I'd say, `array_map('strtolower', $tagorderMap[$configKey] ?? [])` should be sufficient.

src/Module/Util/VaInfo.php
lachlan-00: thank you for confirming

src/Module/Util/VaInfo.php
lachlan-00: there isn't now and i think unlikely will so not a big deal

lachlan-00 (approved):

PR #2899, opened 2021-04-30
Silent notifications + bonus typo fix

### LINKED ISSUE #1626: HTML5 Desktop Nofication Sounds
# Environment

 * Ampache version: Lastest develop branch

 * Client type: Web
 * Browser: Firefox
 * Client OS: Mac

# Description

The HTML5 desktop notifications make a sound, which is kind of annoying. It seems like Chrome doesn't make a sound by default, but neither Firefox nor Chrome have configurable notification sound settings. The notifications API, however, offers an option for silent notifications: [`Notification.silent`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/silent).

I propose that either:

 1. notifications be completely silent, or
 2. a configuration option is added in Ampache to make them either noisy or silent.

It looks like this would be a simple change to the function `NotifyBrowser` in `templates/show_html5_player_headers.inc.php`. I would be happy to submit a patch for this.

mitchray:
Fixes #1626

### DIFF
--- a/src/Config/functions.php
+++ b/src/Config/functions.php
@@ -1233,7 +1233,7 @@ function toggle_visible($element)
  */
 function display_notification($message, $timeout = 5000)
 {
-    echo "<script";
+    echo "<script>";
     echo "displayNotification('" . addslashes(json_encode($message, JSON_UNESCAPED_UNICODE)) . "', " . $timeout . ");";
     echo "</script>\n";
 }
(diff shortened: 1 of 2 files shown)


src/Config/functions.php
```diff
@@ -1233,7 +1233,7 @@ function toggle_visible($element)
  */
 function display_notification($message, $timeout = 5000)
 {
-    echo "<script";
+    echo "<script>";
```
usox, line 1236: :+1:

PR #2903, opened 2021-05-01
Search bar improvements

### LINKED ISSUE #371: Different Behaviors when clicking on search result or pressing enter with keyboard
When using the rather excellent search feature of Ampache there are two methods of selection, each of which performing different results. 

menu

If I were to click on the result it would bring me to either the specific album or artists that was highlighted. However, if I were to press [Enter] while it is highlighted it then awkwardly places the exact string in the search bar. Pressing [Enter] again then merely searches for the exact string.

I would expect the [Enter] behavior to bring me directly to the result rather than perform a search (like clicking the mouse does).

mitchray:
- Fixes ```Uncaught TypeError: n is undefined``` when result rows are hovered on, due to the autocomplete widget expecting .data()
- Fixes #371, pressing [Enter] now navigates to the selected item
- Clicking into search field with existing value will reopen results (having to add/remove characters just to trigger previous results is annoying)
- Preserve the search string instead of replacing it with the selected item
- Fixes the [Tab] key bind

### DIFF
--- a/public/templates/header.inc.php
+++ b/public/templates/header.inc.php
@@ -228,9 +228,14 @@ $ajaxUriRetriever = $dic->get(AjaxUriRetrieverInterface::class);
                             return false;
                         },
                         select: function( event, ui ) {
-                            if (ui.item != null) {
+                            if (ui.item !== null) {
                                 $(this).val(ui.item.value);
                             }
+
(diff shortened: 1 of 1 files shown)


lachlan-00:
that's some nice fixing

PR #2904, opened 2021-05-01
Add missing Play Next options where there is already Play Last

mitchray:


usox:
Note to me: Update new template files for podcasts in edge/podcast branch

lachlan-00:
@mitchray it took me a while to realise but 'play next' on albums is correctly putting the objects next but it's not adding 1 to each track so it ends up putting the album in reverse.

not sure where the position of next is set but we could reverse the list or work out where that is

PR #2905, opened 2021-05-01
Replace hardcoded odd/even classes with CSS selector; remove now unus…

mitchray:
Replace hardcoded odd/even classes with CSS selector; remove now unused ```flip_class()``` method


usox:
This looks really good!
Maybe, the new template files need some changes, too? https://github.com/ampache/ampache/tree/develop/resources/templates

usox (approved):
A huge one, but I'm quite happy with it.

PR #2912, opened 2021-05-06
replaygain not read in ogg files.

pcjco:
It seems that the replaygain tags put in OGG files by foobar2000 for example are processed by getId3 but not by vainfo.php. As a result, no replay gain is seen during playback of OGG files. This patch should process these tags from OGG files:
- replaygain_track_gain
- replaygain_track_peak
- replaygain_album_gain
- replaygain_album_peak


lachlan-00:
tested it out and the tracks with these tags are imported now

lachlan-00:
another new contributor!

PR #2915, opened 2021-05-07
Specify SameSite on all cookies (use php 7.3 array parameter)

### LINKED ISSUE #2910: sidebar cookie SameSite value
Just noting this so I don't forget

```Cookie “sidebar_state” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite```

mitchray:
Fixes #2910

- Use the PHP7.3 array parameter for setting cookies, which lets you specify a ```samesite``` value.
- Amends the JS cookies which were using an old workaround to set samesite.

One problem remains though, I haven't been able to get the ```ampache``` (session_name) cookie to accept a samesite value, and I don't fully understand why that is as according to the PHP docs it should happily accept the samesite value. 

Assuming its to do with it being a session cookie (session-set-cookie-params), the furthest I could get was changing the PHP setting ```session.cookie_samesite = "Strict"```...

### DIFF
--- a/public/templates/header.inc.php
+++ b/public/templates/header.inc.php
@@ -497,7 +497,7 @@ $ajaxUriRetriever = $dic->get(AjaxUriRetrieverInterface::class);
                     $('#sidebar').show(500);
                 });
 
-                Cookies.set('sidebar_state', newstate, { expires: 30, path: '/; samesite=strict'});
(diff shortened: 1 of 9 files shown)


lachlan-00:
I think that 'ampache' cookie does require php config on the server. i didn't actually check for a workaround yet though

PR #2937, opened 2021-06-09
crop images from the centre instead of resizing into a square

mitchray:
First change is to keep aspect ratio of original images during art selection

art-selection

Second is to crop non-square images from the centre (or slightly above centre for portrait images making the assumption that faces would most likely be higher), instead of squashing the image into being a square. This does mean there is a possibility of faces being cut off but I feel that is better than distorted images

before-after

### DIFF
--- a/public/templates/show_arts.inc.php
+++ b/public/templates/show_arts.inc.php
@@ -49,7 +49,7 @@ while ($count <= $rows) {
             echo "<td>&nbsp;</td>\n";
         } else { ?>
             <td>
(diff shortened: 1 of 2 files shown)


lachlan-00:
haven't checked with the new commit (will later) but image search would not insert the image after selecting one with the first one.

mitchray:
It was working fine prior I just thought it best to be thorough, is it failing for any attempt or some specific artists?

lachlan-00:
i'll make sure i clean up my server and try again. it may have been from other work i was doing

lachlan-00:
looking good now, must have just been me. just pulled and merged with develop so no issue now
image

PR #2958, opened 2021-06-24
Faster regex for splitting musicbrainz UUIDs

SePHTaN:
improved computation time only needs 20 steps instead of 74 to split UUID's


wagnered:
@SePHTaN

I don't understand the premise of your PR:  Why does the UUID need to be split? verified for sure, 

I am and have been working on writing/importing both vorbiscomments and id3v2.3 frames.  I have been adding frames not removing any.  I don't think any frames defined by the id3v2.3 standard should be removed.  

At first glance, at least some of your work is in conflict or a duplication of what I have been doing.

Just thought I would let you know.

SePHTaN:
the UUID (all musicbrainz ids are UUIDs) is split when there are multiple in a tag because for example two mb_artistid's are store with delimiter .
It's an alternative lachlan added for the trim_slashed_list in Artist.php to get an array instead only the first mbid.
I only exchanged the regex with one that is much faster.

MCDI aka music_cd_identifier i removed to get rid of:
```
(Ampache\Module\System\Dba) -> Error_query SQL: INSERT INTO metadata (object_id,field,data,type) VALUES(?,?,?,?)
(Ampache\Module\System\Dba) -> Error_query MSG: ["22007",1366,"Incorrect string value: '\\x9F\\x00\\x10\\x03\\x00\\x00...' for column `ampachetest`.`metadata`.`data` at row 1"]
```
at least the error disappeared with the tag.
Honestly i have to admit that a quick test, without removing the tag, didn't gave me this error.
My intention was only to get the logs clean especially when the cause for the error is something ampache doesn't use or need (at the moment).
Taking a closer look tomorrow to be sure not to have overseen anything regarding MCDI tag.
When the error persists i will add an comment here, and hopefully you are able to help to correctly read the MCDI tag.

medium counts (disk, totaldisks, totaltracks) weren't complete or correct.
haven't implemented following specification but reallife tags (from picard) 
those are needed for musicbrainz search to be able to correctly identify releases, recordings when there are no mbids within tags

so if there any things left that interfere with your work let me know, the last thing i want is causing regressions or errors instead of removing them.

wagnered:
> My intention was only to get the logs clean especially ...

I'm not sure that is an appropriate method to change something that simply irritates you.  🙂

Well, I did some research on music cd identifier.  I believe it was added to the id3v2 standard  to support the cddb database.  The CDDB database was bought and  then discontinued by Gracenotes.  Freedb and other databases are being discontinued so the MCDI frame is pretty much useless now.  I agree: remove it.

wagnered:
> the UUID (all musicbrainz ids are UUIDs) is split when there are multiple in a tag because for example two mb_artistid's are store with delimiter .

Tha

wagnered:
> the UUID (all musicbrainz ids are UUIDs) is split when there are multiple in a tag because for example two mb_artistid's are store with delimiter .

That complicates matters when writing all song artists and  mbid's back to the file tag.

As for the mysql error, it looks like the php hex presentation of the data is attempted to be written to the column.    or possibly a null character ('\0'). I can't be sure until I start working on the id3v2 part.  Possibly using hex2bin on the string might work then using `hex($data)` in the mysql insert statement.  Anyway, I guess it is a moot point.

lachlan-00:
i don't actually have and mcdi that i can see in my tags for any files so i don't think it's important either.
https://id3.org/id3v2.3.0#sec4.5

lachlan-00:
no issues with import/verify/update using this merged with develop that i can see

SePHTaN:
> no issues with import/verify/update using this merged with develop that i can see

During my first tries for multi-artist feature and later for expanding musicbrainz plugin is selected several albums for testing that fulfilled some requirements like:
1. being an alias for an artist (Die Roten Rosen alias for Die Toten Hosen), or having multiple artists for songs (MA Feature)
2. being an multiple disc album with different folder structure (musicbrainz plugin)

I edited that tags for those test albums especially i deleted every tag that a normal cd-ripper like EAC doesn't add by itself to have a minimal tag scenario for testing the musicbrainz plugin functionality.
I then stumpled over the mcdi tag causing metadata insert errors and realized that no tagger i tested allowed me to see or remove this tag at least not in the standard configuration.
That's why i decided to put it in the switch: case: block to suppress that it got added to the final $info array.

PR #2960, opened 2021-06-26
Proper handling of album titles with prefixes.

wagnered:
With these changes, editing album name doesn't require creating new album and migrating song data.


lachlan-00:
so the prefix was put into the name field instead of splitting up? that seems like a very obvious oversight of how the table is set up

PR #2976, opened 2021-07-23
Update file tags

wagnered:
Second try.

1. Added capability of writing musicbrainz data.
2. Changed `SongId3TagWriter` class and all references to `SongTagWriter`.
3. Changed `write_id3` config switch to `write_tags`.
4. Removed `write_id3_art` switch.  Because the entire tag, metadata
   and images, must be written each time, both are automatically updated when
   updating file tags.
5. Added getter to Song model for `total_count`.
5. Added song's total play count to writing of POPM frame.


lachlan-00:
will have time to run on my server tomorrow. seems okay without running the change though.

wagnered:
The various musicbrainz ids were rather ambiguous and undocumented as to which 'mb', recording or release, was which. What I'm assuming is their flac and TXXX designations should be verified before merging.

lachlan-00:
sorry for the slight delay i'll be merging this one tomorrow  once i've put it on my local to test.

wagnered:
No hurry.  Just want to have verified that mbids in the various tables are being written to the correct file tags.

lachlan-00:
found a few things, a file without any text frames will cause an error.
```
2021-07-28T10:56:29+10:00 [ampache] (log.lib) -> [Runtime Error] Invalid argument supplied for foreach() in file /var/www/music/src/Module/Song/Tag/SongTagWriter.php(92)
2021-07-28T10:56:29+10:00 [ampache] (log.lib) -> [Runtime Error] Invalid argument supplied for foreach() in file /var/www/music/src/Module/Song/Tag/SongTagWriter.php(101)
```

I also just pushed up some missing write_id3 & write_id3_art => write_tags converts.

wagnered:
Thanks

> found a few things, a file without any text frames will cause an error

Also if an mp3 originally has no metadata, it can't be written to.  Looks like same with FLAC & OGG files.  Forgot to check that situation.  Will have to rethink it.

src/Config/ConfigurationKeyEnum.php
```diff
@@ -106,8 +106,7 @@ final class ConfigurationKeyEnum
     public const LOCALPLAY_LEVEL               = 'localplay_level';
     public const ALLOW_ZIP_DOWNLOAD            = 'allow_zip_download';
     public const ENABLE_CUSTOM_METADATA        = 'enable_custom_metadata';
-    public const WRITE_ID3                     = 'write_id3';
-    public const WRITE_ID3_ART                 = 'write_id3_art';
+    public const WRITE_TAGS                    = 'write_tags';
```
wagnered, line 109: Don't understand what happened here and other locations.  I even documented the changes in the PR comment. This is one location that I knew needed changing.  Maybe I  somehow inadvertently reverted the changes.

lachlan-00: git has the most ridiculous merging system sometimes.  i need to put in the pre-commit hooks so i can stop adding php-cs fix commits

lachlan-00:
getting close to done on a lot of things now so soon this will need to merge in soon.

No errors with the process after your updates just the untagged files don't modify.

Do we keep untagged files out of tagwrite for now as a caveat because "Ampache is not a tag editor" and that we expect people to tag things first?

wagnered:
> No errors with the process after your updates just the untagged files don't modify.

By "untagged" do you mean absolutely no id3v1 or id3v2.3 data

wagnered:
Sorry about that.

I'll assume for now that is the case and find a way to create a test mp3 audio file without any metadata.

It looked like it should have worked, but don't know what I'm overlooking.

lachlan-00:
> Sorry about that.
> 
> I'll assume for now that is the case and find a way to create a test mp3 audio file without any metadata.
> 
> It looked like it should have worked, but don't know what I'm overlooking.

i'll see if i can tag my untagged test as well

wagnered:
I was able to successfully write metadata to an mp3 file without any id3v1 or id3v2.3 metadata.  I forgot to test with musicbrainz metadata. (TXXX frames).

lachlan-00:
that's good, it could just be this specific file then

lachlan-00:
> I was able to successfully write metadata to an mp3 file without any id3v1 or id3v2.3 metadata. I forgot to test with musicbrainz metadata. (TXXX frames).

working on test files, there's an issue with my untagged file when updating, just checking it out.
```
2021-08-03T12:53:58+10:00 [user] (Ampache\Module\Api\Edit\AbstractEditAction) -> Called for action: {edit_object}
2021-08-03T12:53:58+10:00 [user] (Ampache\Module\Api\Edit\AbstractEditAction) -> Ampache\Repository\Model\Album
2021-08-03T12:53:58+10:00 [user] (Ampache\Module\Api\Edit\AbstractEditAction) -> 161180
2021-08-03T12:53:58+10:00 [user] (Ampache\Module\System\Dba) -> Error_query SQL: UPDATE `album` SET `album_artist` = ? WHERE `id` = ?
2021-08-03T12:53:58+10:00 [user] (Ampache\Module\System\Dba) -> Error_query MSG: ["22007",1366,"Incorrect integer value: 'New album 2' for column `music`.`album`.`album_artist` at row 1"]
```

should be all we need to fix now so that's another big tick off the list!

src/Repository/Model/Album.php
```diff
@@ -950,14 +950,14 @@ public function update(array $data)
             }
 
             if (empty($data['album_artist_name']) && !empty($album_artist) && $album_artist != $this->album_artist) {
-                self::update_field('album_artist', $album_artist, $this->id);
```
lachlan-00, line 953: revert this and we're done. it's trying to put a string album name in for artist id

lachlan-00:
fyi i found my issue with the upload, it's being skipped because it's an upload catalog (not == 'local')

i'm going to do a separate commit for that but we're done after you put hte artist back :)

PR #2990, opened 2021-08-12
Fix invalid comment syntax in sql file

biondizzle:
Web installer fails because /**/ is not valid comment syntax for an SQL file. Converted to valid .sql comment syntax


lachlan-00:
a new contributor!

PR #3007, opened 2021-09-01
fix SQL for random selection of albums

cfe-dev:
fixes incorrect SQL (missing WHERE) for some cases.


lachlan-00:
i was trying to think of a more elegant way without 1=1 but it's more convoluted so this looks fine

cfe-dev:
should be fine as long as there's no OR (outside of parentheses) in the WHERE clause of this specific SQL query.

PR #3011, opened 2021-09-04
Update View > Limit once focus moves away, instead of after key up

### LINKED ISSUE #2435: Can't change view limit on Android.
By the time I click the drop down list. place the cursor into the textbox, then move down to the numbers, the screen refreshes and no changes are made.
This could be an ideosyncracy of android.  Does this happen on a desktop?  If not, I would like to suggest a  work around by adding an adustable default page limit.

Also if not, perhaps the react interface might provide a cure.

mitchray:
Fixes #2435


lachlan-00:
i have safely never done this before but that is annoying AF

PR #3037, opened 2021-10-10
Fix logic of SQL query to get random albums

cquike:
This commit fixes the SQL query found in getRandom() of AlbumRepository.

Note that there are several configuration options that affect the result of getRandom(): album_group, catalog_disable, catalog_filter, rating_browse_filter. It would probably be useful that a unit test is added that tests all the several combinations of those parameters. However I am not familiar with php unit testing, so I couldn't add a test myself. I have nevertheless tested manually a few of the possible combinations.


cquike:
Oh, I see now that there is a conflict between this PR and a commit in the develop branch: 9261184
Actually that commit already fixed the problem, but in a way that uses again "where 1=1" which is exactly what commit 27c5bb5 tried to avoid. I am not sure now which is the best course of action here...

lachlan-00:
i use a where join usually for these.

This does unify the style a bit more compared to other areas so i'll update that as well

cquike:
Excellent. Thanks!

PR #3042, opened 2021-10-13
Cache bust some frequently updated assets

mitchray:
This uses the MD5 hash of a file to create a unique cached version, to avoid 'just clear your browser cache' issues

- Reference a cached version of asset if it exists, otherwise create cached version and reference it, or fallback to original file
- Added 'Clear Assets Cache' option to Ampache Debug page
- Change paths in CSS to be web root relative to account for different physical locations of original and cached CSS file

Not sure if I should also clear the asset cache during clean_catalog() just so its periodically cleared?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ public/develop
 public/master
 public/foam
 public/ample
+public/cached_assets/
 storage/
 tmp/
 /vendor/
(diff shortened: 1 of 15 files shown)


lachlan-00:
could do it as part of the update function in AutoUpdate::update_files?
image

that would then clear it after an update

PR #3049, opened 2021-10-14
only glue with comma if more than 1 item

mitchray:
firefox_YTkLDrYzWP

Prevents comma being added when array is empty, or when it has just one of the items


mitchray:
Looks like it has!

Issue #220, opened 2014-05-29
Top tracks


SUTJael:
Would be great to add a top track panel like deezer for each artist (http://www.lastfm.fr/api/show/artist.getTopTracks). We could also think about a global "Top tracks" section based on top tracks per country (http://www.lastfm.fr/charts/tracks/top).

LesKent:
I think such a feature would be great for music discovery. Plex's music UI has this feature and I find it really useful and would love to see this in Ampache.

lachlan-00:
putting this in the ui would be good to do still. i use it in subsonic all the time so will look at it again

PR #3052, opened 2021-10-15
Top Tracks

mitchray:
This is my attempt at fulfilling Top Tracks #220, it loads 20 songs and displays them in a table

firefox_WuAINJU5Y6

I couldn't get the grid layout (or any of the View options for that matter) to apply without a manual refresh so I abandoned that, and cut corners by using JS to populate the indexes instead of editing the song_row.xhtml template...

Hopefully it is at least a helpful jump point for refining, and I won't feel hurt if it doesn't get used at all haha, it gave me better insight into the inner workings


lachlan-00:
just going to add popular threshold as the count option and while i'm at it i'll add the getsimilartracks that subsonic has to get related tracks.

need to check if spotify has the same type of api calls as last.fm so we don't have to rely on one provider

PR #3071, opened 2021-10-27
Misc. fixes on webplayer js/css/img

arnaudbey:
- remove mozfullscreen deprecated msg
- vizualizer toggling regardless of the theme used (playlist etc. was reset as black) 
- misc.


lachlan-00:
> @lachlan-00 there is a lot to do, and it is quite difficult to get a satisfactory result, with the mix of HTML / PHP / JavaScript.
> 
> sometimes there is php in JavaScript just to write a config value, the webpath, etc. Maybe a good solution (good, I don't know, but maybe more elegant than the current one) would be to store some values as data-attribute in < html > for example. This would make them accessible to all scripts without having the PHP interrupt in JavaScript.
> 
> There is also the problem for translations. For that, exposing a json file that contains the translations could be good?
> 
> What do you think about this?

separating the php and JS is a pretty good idea because it is really ridiculous how it's generated currently. @mitchray was looking at that i think too.

keeping it simple would be a good start, little pieces like taking out web path variables and other weird little things would make the code easier.

Translations are really important to maintain and do in an easier way would help. I don't use JS outside this project so i'm happy to do whatever works.

To me i think the best thing to fix are things that are hard to read or understand. I've been doing this for 4 years and it still requires a lot of effort to follow the process in the JS. That's where the long term bugs usually sit too.

mitchray:
https://github.com/mitchray/ampache/commit/749a4bda2590001cfc62609affd2bc6abbeb3410#diff-5146f230c067a00c9d0be5ef28e88229c6bf42ace0e569fd0c13c9a04f6c750bR83

This is my trial at separating the PHP from the JS, by initialising all the PHP variables into a single JS object. 

I left the PHP translation calls mixed in but just having all the JS be a single block and not a bunch of PHP echo strings was an improvement I thought

https://github.com/mitchray/ampache/commit/749a4bda2590001cfc62609affd2bc6abbeb3410#diff-5146f230c067a00c9d0be5ef28e88229c6bf42ace0e569fd0c13c9a04f6c750bL173-R212

RobDeLuca60 (approved):

lachlan-00:
closing as it's been sat on for so long its out of date now

Issue #3091, opened 2021-11-12
Getting Loading issue in 5.1.1


GaneshKandu:
### Describe the bug
After successful install of Ampache 5.1.1 i am getting loading issue.
backend login.php is unable to load assets

> I am installing on server with ssl installed and want to install ampache with https protocol

### To reproduce
Try installing with ssl and https in url 

### Screenshots

![image](https://user-images.githubusercontent.com/12833130/141440110-592ad895-ab8e-4d3d-bb59-439c90a1a202.png)

## Environment

* Ampache version: 5.1.1
* Web server + version: Apache 2.4.51
* Server operating system: Linux
* Client operating system: Windows

### Client type

* SubSonic

## Settings

<!-- Either put a description of your settings, or paste the contents of `conf/ampache.cfg.php` after removing sensitive information (server host, database connection, etc.). If posting full config file, make sure to put triple back-ticks ``` at the top and bottom to make it into a code block.

[ampache.cfg.php.txt](https://github.com/ampache/ampache/files/7526211/ampache.cfg.php.txt)

```

## Logs

<!-- To enable logging, visit https://github.com/ampache/ampache/wiki/Troubleshooting#enable-logging

Please post relevant Ampache logs, and web server access/error logs, making sure to surround in triple back-ticks ``` to make into a code block. If you have large logs, it may make sense to trim them to a shorter time-frame if you know exactly when the error occurred. If appropriate, post any client error logs as well -->

lusum:
Can you try with this package: https://github.com/ampache/ampache/archive/refs/heads/develop.zip ?

Similiar issues are solved by a recent patch. If it is not solved, can you post your ampache configuration and the logs?

lachlan-00:
i've tested both squashed zips with php7.4 and php8 on apache without issue so it's probably a config issue somewhere

GaneshKandu:
Hi @lachlan-00 

I am running PHP 7.4.19 with FCGI mode enabled
FastCGI seems to causing this strange behavior

```php
    public function isSsl(): bool
    {
        return (
            (
                filter_has_var(INPUT_SERVER, 'HTTP_X_FORWARDED_PROTO') &&
                Core::get_server('HTTP_X_FORWARDED_PROTO') == 'https'
            ) || (
                filter_has_var(INPUT_SERVER, 'HTTPS') &&
                Core::get_server('HTTPS') == 'on'
            )
        );
    }
   ```
on line no 243

```php
[...]

PR #3099, opened 2021-11-23
filter_has_var is returning false in FCGI enabled PHP

GaneshKandu:
filter_has_var is returning false even if $_SERVER has HTTPS defined in FCGI enabled PHP versions, that causing this issue.

### DIFF
--- a/public/.maintenance.example
+++ b/public/.maintenance.example
@@ -12,7 +12,7 @@ if (empty($domain)) {
     $domain = $_SERVER['SERVER_NAME'];
 }
 $scheme = "http";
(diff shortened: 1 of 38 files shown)


GaneshKandu:
#3091

lachlan-00:
cool! thanks for digging into this

lachlan-00:
do you also have issues with filter_var or just filter_has_var?

lachlan-00 (approved):

lachlan-00:
changed some more to match in 9e6448de0

GaneshKandu:
> do you also have issues with filter_var or just filter_has_var?
with filter_has_var only

Issue #2388, opened 2020-05-18
ldap_require_group does not work


mneiger:
## Description
The ldap_require_group does not work with openldap because teh code checks for the user dn in the memberuid instead of the username

Setup ldap authentication with openldap
once it working, add searching for member ship in a group

##Steps to correct the behavior:
replace in lib/class/ldap.class.php / line 288

```
                if (! p`reg_grep("/^$username\$/", $group_infos[$member_attribute])) {
                    throw new LDAPException("`$username` is not member of the group `$group_dn`");
                }

```

## Environment

* Ampache version: 4.1.1-release
* Web server + version: nginx 1.14.2
* Server operating system: debian buster
Linux yohanan 5.4.0-0.bpo.4-amd64 #1 SMP Debian 5.4.19-1~bpo10+1 (2020-03-09) x86_64 GNU/Linux
* Client operating system: linux mint 19.3 Firefox

### Client type

* Web Firefox 76

## Settings

ldap_url = "ldap://localhost/"
ldap_search_dn = "ou=People,dc=xxx,dc=org"
ldap_objectclass = "posixAccount"          ; OpenLDAP
ldap_filter = "(uid=%v)"                   ; OpenLDAP
ldap_require_group = "cn=ampache,ou=Groups,dc=xxx,dc=org"
ldap_name_field = "cn"
ldap_member_attribute = "memberuid"

### Non working log

> 2020-05-18 22:20:55 [ampache] (ldap.class) -> binding with username `cn=az test,ou=People,dc=xxx,dc=org` 
> 2020-05-18 22:20:55 [ampache] (ldap.class) -> reading attributes ["memberuid"] in `cn=ampache,ou=Groups,dc=xxx,dc=org` 
> 2020-05-18 22:20:55 [ampache] (ldap.class) -> Exception: `cn=az test,ou=People,dc=xxx,dc=org` is not member of the group `cn=ampache,ou=Groups,dc=xxx,dc=org` 
> 2020-05-18 22:20:55 [ampache] (ldap.class) -> Error during authentication: `cn=az test,ou=People,dc=xxx,dc=org` is not member of the group `cn=ampache,ou=Groups,dc=xxx,dc=org` 
> 2020-05-18 22:20:55 [ampache] (ldap.class) -> Return value of authentication: {"success":false,"error":"`cn=az test,ou=People,dc=xxx,dc=org` is not member of the group `cn=ampache,ou=Groups,dc=xxx,dc=org`"} 
[...]

PR #3107, opened 2021-12-03
fix for ldap_require_group does not work #2388

mneiger:
replaced user_dn with username as property to check for group membership in ldap.

### DIFF
--- a/public/templates/show_album.inc.php
+++ b/public/templates/show_album.inc.php
@@ -75,8 +75,9 @@ if ($directplay_limit > 0) {
     <?php } ?>
     </div>
     <?php
+        $name  = '[' . scrub_out($album->f_album_artist_name) . '] ' . scrub_out($f_name);
         $thumb = Ui::is_grid_view('album') ? 32 : 11;
-        Art::display('album', $album->id, $title, $thumb); ?>
+        Art::display('album', $album->id, $name, $thumb); ?>
 </div>
 <?php if (User::is_registered()) { ?>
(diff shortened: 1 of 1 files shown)


lachlan-00:
is distinguished name used in a specific ldap server? 

you could do a `(?=$user_dn|$username)` that would cover both?

mneiger:
Notwithstanding the fact that I'm not able to check the exact PHP syntax, i agree that checking against both user_dn and username would make sense. My ldap server is openldap;  we should have someone using Active Directory and/or more proficient with LDAP give his 2 cents.
I'd be glad to adjust the pull request if you feel so.

mneiger:
my latest findings:
I first thought it was an OpenLdap vs Active DIrectory directory difference, but it's not.
The member attribute array of a group can either store dn or uid (i.e. username)., For users, it usually stores the username rather than dn, dn is mainly used for nested groups.
Though we should theoretically have a more comprehensive code to deal with both case and nested groups, we would need someone actually knowledgeable about ldap or willing to investigate. I do not have the time and skills myself to go much further.

lachlan-00:
i think we've had enough problems with groups that this will probably fix it.

If there are implementations of ldap that do it differently i'll make it a config option.

lachlan-00:
> replaced user_dn with username as property to check for group membership in ldap.

Welcome as a new contributor to Ampache!

mneiger:
Many thanks, this is actually my first real pull request on github. I'm proud it to be for ampache.

Issue #3116, opened 2021-12-12
[Feature Request] Gitpod support


ghost:
## Describe what you don't like about Ampache, or what it is missing

As a newbie to some language or project, it's nice to be able to just spin up an environment where everything Just Works™️ and one can get to coding or exploring the project without having to go through the hassle of installing a bunch of tools before one can even get started.

I'm missing this for Ampache as I know C++, Java, Javascript, Python and Rust, but have always avoided PHP. This is the first project I'd like to contribute to in PHP, but now I have to go through the process of setting up an environment to do so.

## Describe how you would like the feature to work**

[Gitpod](https://www.gitpod.io/) is a free to use product (I think it's by Redhat) that allows editing a project online in a VS Code environment. Project "just" have to provide a `.gitpod.yml` then users can point GitPod to it and a functional, working environment will be set up, and the dev can get straight to developing and/or exploring.

[Jetbrains Space](https://www.jetbrains.com/space/) is promising to be similar, but it's not there yet.

mitchray:
Closing due to inactivity

PR #3119, opened 2021-12-14
Add docker-compose for quick dev env

LoveIsGrief:
#3116 (gitpod support) made me realize it'd be easier to work with something I know, so here's a PR to help newbies like me to setup a new environment without having to deal with installing LAMP on the host machine.

I added the information to `CONTRIBUTING.md` because I don't have access to the wiki.

It works without a hitch in my linux env + PHPStorm + XDebug.

Devs can now `docker-compose up -d` and access http://localhost:8080 with user:passwd `ampache:ampache`. They can shove some media files into `docker/volumes/media` and create a catalog there which works.

With a debugger of their choice, they can listen on host port `9003`, create a `docker-compose.override.yml` (see `CONTRIBUTING.md`) and `docker-compose up -d ampache` to get callbacks from XDebug.


lachlan-00:
cool, i'm going to be away for a week from tomorrow but when i come back i'll set myself up this way and test it all out.

I'll be back about the 24/25th and will merge this when i've tested how it works and can update the wiki too

LoveIsGrief:
I hope it'll work and look forward to the feedback.

Happy holidays! :bell:

LoveIsGrief:
@kuzi-moto we should probably have the discussion about docker in here. I can always drop the commits from #3124 that have the docker stuff in them.

I had a look at the https://github.com/ampache/ampache-docker/blob/master/Dockerfile and keeping it in a separate repo has the problems, imo:

 - Builds aren't triggered through DockerHub automatically when this repo is updated/pushed (probably manually done or with an external service to trigger Dockerhub)
 - It's a monolith image with all services in one --> adding separate services e.g redis or nginx requires rebuilding the entire image
 - Using the repo in here either means a submodule (either in here of over there) which comes with its own set of complications
   * Not everybody knows about submodules, so there will be separate documentation needed that will have to be easy to find
   * Modifying the `Dockerfile` / docker setup will require the following steps before merging it into this repo:
      - Forking ampache/docker
      - ampache/ampache: pointing the submodule to a forked repo
      - ampache/docker: Getting a PR for merged there
      - ampache/ampache: pointing to the `master` tip of `ampache/docker`

It's definitely the latter reason why I put all the docker stuff in here. I didn't it up for Dockerhub since I was/am purely focused on getting a developer environment up and running. I simply couldn't find any documentation on how to do that and have never used PHP before.

lachlan-00:
back,

why is the website argument used for email? that's a bit of a bug.

docker-compose.override-template.yml
```diff
@@ -0,0 +1,25 @@
+version: "3"
```
lachlan-00, line 1: move this into templates folder `docs/templates/docker-compose.override-template.yml`

.gitignore
```diff
@@ -50,3 +50,8 @@ web.config
 *.v11.suo
 *~
 config/ampache.cfg.php.backup
+
```
lachlan-00, line 53: put this just in the file as normal
image

CONTRIBUTING.md
```diff
@@ -7,7 +7,41 @@ submitting bug reports and feature requests or writing code which can be incorpo
 Anyone can take part in our community and it there are no rules or requirements stopping you from joining.
 Hopefully this document will help you make the jump!
 
-Please read [Development section](https://github.com/ampache/ampache/wiki#development).
```
lachlan-00, line 10: duplicate the updated file in `.github/CONTRIBUTING.md`

src/Module/Cli/AdminAddUserCommand.php
```diff
@@ -55,11 +55,26 @@ public function execute(
         $values     = $this->values();
         $interactor = $this->io();
 
+        $email = $values['website'];
```
lachlan-00, line 58: is this a bug you found? i'd keep it simple with checking email and user separately
image

lachlan-00:
update from develop as well and when we get this all done i'll just squash it down into one commit

src/Module/Cli/AdminAddUserCommand.php
LoveIsGrief: Yeah, I was able to add two users with the same email and username. I'm not going to touch migrations in here, but the email and username columns should probably be `UNIQUE`.

CONTRIBUTING.md
LoveIsGrief: Not for this PR, but maybe a symlink would work too. I copied it nonetheless.

LoveIsGrief:
It should be ready now @lachlan-00

lachlan-00:
looks good now, i'll run it up here and if i can handle it then everyone else can :)

docker-compose.override-template.yml
```diff
@@ -1,25 +0,0 @@
-version: "3"
```
lachlan-00: might not have added this in to the pull?

LoveIsGrief:
It's renamed and moved in the docs/templates/docker-compose.override.yml

I figured `override-template` wasn't necessary since it's in the templates folder anyway.

lachlan-00:
the web server should go to the root instead of /public as well. will probably create a few errors if not using the right webroot.

not sure how it's configured in nginx but will try again tomorrow

LoveIsGrief:
Hmm, I'm not sure what you mean. The root is /var/www/html and it only redirects to /public when when matching exactly "/". Even /index.php states that the root ampache folder is /public.
The repo root is mounted at /var/www/html in the nginx and ampache services.

lachlan-00:
when i go to localhost:8080 i get redirected to /public
image

The public folder is where the document root for the server should be

LoveIsGrief:
OK, I see. I misunderstood the message in `/index.php`. Your issue was actually being caused by the `ampache` service not being available (502 means nginx can't reach the upstream host), due to the db not being ready during ampache's setup script run. `docker-compose start ampache` would've resolved it.

I added a `restart: on-failure` for the `ampache` service to make manual restarting unnecessary, updated the document root and removed `web_path = "/public";` from the config to remove `/public/` from URLs.

lachlan-00:
i get it now, my docker lock was set to php7.4 which was stopping the service completing

LoveIsGrief:
Is there anything else I need to do?

lachlan-00:
Still not working well on this pc but i'm going to clean up my other laptop and try to run it from there. i think i screwed around too much with the docker on this thing so i think it's me whos causing the problems rather than what's here now

lachlan-00:
i'm looking at making this a bit simpler, the nginx stack is just overly complicated for me (mostly because i don't use nginx or docker regularly) and i have not been able to get it working at all on my PC.

https://github.com/ampache/ampache/tree/localdevenv

@LoveIsGrief and @kuzi-moto i've added this branch for a bit of a review compared to this pull using apache2 instead

i've tried to keep this simple by running a shell script with the perms and config. from there build and compose up should make it pretty simple.
image

The only thing i'm not sure about is xdebug which i've enabled as a php extension but not something i actually use either

LoveIsGrief:
I'm not sure why it didn't work for you, but just to be sure it wasn't a "it works for me" problem, I tested my branch in clean VMs running Arch Linux, OpenSuse 15, and Ubuntu 20.04. They all work https://github.com/LoveIsGrief/ampache-dev-docker-test . Even tested it on my local NixOS and it works too. What errors were you running into?

`nginx` is in a separate container in order to allow anybody to switch out the webserver for whatever they like: apache, httpd, caddy, etc. `nginx.conf` was also taken from the [wiki](https://github.com/ampache/ampache/wiki/Installation#nginx).

The reason I didn't add any scripts to run before running docker was because imo, somebody new to the project should just be able to run `docker-compose up` or a single command and be up and running. Additionally, no further steps e.g creating a user, configuring the DB access, etc. should be necessary.

lachlan-00:
for me the nginx site isn't able to connect to anything so the ampache-fpm and db are running okay as far as i can tell

I haven't been able to site down even on another laptop for a while so it's one of thing things i haven't really worked out why yet.

LoveIsGrief:
I'm not using ampache anymore and don't think I'll be touching PHP again, so I'm closing this

Issue #2382, opened 2020-05-12
[Question] About cron.php: Does it really work?


wagnered:
Could someone please explain how this works: Maybe I'm missing a critical piece of information.

Tbe following is what I understand:

The script is run using php.  The modern webserver runs either php-cgi or php-fpm which I believe manages a pool of php-cgi processes. Php-fpm runs continuously in it's **own process space** while the webserver is running.

As I understand the Linux system, the cron script run by php is  a separate process from  the php-fpm, or php-cgi process created by the webserver. To get this to work correctly, the cron script needs to be run once by Ampache upon its startup. That way, it will be in the same process space created by the webserver and can fill the cache in that process space.

Even if the cron.php file was executed using php-fpm outside the webserver process, the system would create a separate process and it would be isolated from the webserver's php-fpm process.

Of course this begs the question: is a cron task necessary?  I say "no".

BTW & FYI & FWIW: I just learned that if the webservers are running php-fpm, init.php would need to be run only once and not in every top-level script because of php-fpm's persistence.

lachlan-00:
you're right, fpm is loading processes all the time so parts of this won't work.

It would need to move into memcached to be effective this way

wagnered:
You might consider [Redis](https://redis.io/documentation) which is BSD licensed.

Memcached is meant to be, but doesn't need to be a distributed(cloud) in-memory cache system.  It does one thing and it does it well.

Redis can be an in-memory cache, and database with on-disk persistence, and a publish/subscriber messaging system.  Thought I'd suggest it because of it's apparent versatility.

I didn't take a close look at the data types of each system, but I I remember Ampache caching only basic data types.

lachlan-00:
i'm using redis for nextcloud and it seems simple enough. i'll look at that first

lachlan-00:
it seems to be that a lot of these solutions are good for storing strings related to values but not really for storing arrays of data? 

redis and memcached don't seems to have the same drop in requirements of the live php arrays/indexes

wagnered:
Whew...  I thought I was going senile.  I took another look at cron.inc and the live cache statements were gone.  I **eventually** decided to check the history and discovered why. :smile:

What I haven't discovered is the purpose of compute_cache.inc.  when I looked at its original commit, I didn't see any corresponding changes to Ampache : Specifically, changes to access table "cache_object_count".  Were the changes made in some unrelated commit?  If so, that's a bad habit to get into.

[...]

PR #3124, opened 2021-12-16
Use redis for caching

LoveIsGrief:
# Intro

[Redis] is an in memory data-store with a versatile API and a wealth of clients.
Since ampache is often run in a multi-process environment (PHP FPM e.g), the memory of a single PHP instance dies when then process dies.
Externalising the cache to another process is what redis takes care of for `database_object` subclasses.

**Sorry about the docker stuff!** Once #3119 is merged (or not) I'll know whether to keep it.

# Questions

This is still a draft, but feedback would be great!

## Where else should this be used?

Probably there are other places where the memory cache is used 

## How should the cache be filled?

It seems like only items that are accessed at least once are added to the cache, however certain objects can consist of many other objects that aren't yet in the cache (notably playlists). [\Ampache\Module\Util\Cron::run_cron_cache](https://github.com/ampache/ampache/blob/743782319d3692e9cf1f6084ce450131f8798446/src/Module/Util/Cron.php#L88) was deprecated in a10da55b8659f9d7fd801c267348a305432e52dc and would seem like a great command to run periodically in order to refresh the cache.

----------------------

Related to #2382

[Redis]: https://redis.io

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git/
+.github/
+.tx/
+vendor
+public/lib/components
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@
 .tx/
 vendor
 public/lib/components
+docker/volumes/media
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ config/ampache.cfg.php.backup
 
 # Devs should be able to customize their docker-compose env without unintentionally committing it
 docker-compose.override.yml
+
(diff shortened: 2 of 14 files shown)


kuzi-moto:
Some interesting work here, though there is already a repository for docker. Have you seen it already?

I'm not sure adding docker information to the main Ampache repo is a good idea, or it may be confusing for people new to the project.

I think there is an argument to be made for moving development of Docker to this repo, but that is a discussion we should have with @lachlan-00

PR #3148, opened 2022-01-21
Improve description of rss feed

cquike:
This PR will improve the description of the rss feeds that ampache creates. The new content looks like this:
```
User: my_user
Title: The Song title
Artist: My Artist
Album: The Album Titile
Play date: 21/1/22 19:29

```
Apart from improving a bit the contents so that it is more useful there is an additional benefit: in some RSS readers (like Akregator), if the description of the feed does not change with respect to the stored feed items then it is considered as the same feed item. If a song is played twice within the 10 last feed items then it appears as only one played item. With this change it will appear as two distinct items with a different play timestamp.

### DIFF
--- a/src/Module/Util/AmpacheRss.php
+++ b/src/Module/Util/AmpacheRss.php
@@ -284,11 +284,12 @@ class AmpacheRss
             $is_allowed_recent  = ($user) ? $user->id == $row_id : $has_allowed_recent;
             if ($song->enabled && $is_allowed_recent) {
                 $song->format();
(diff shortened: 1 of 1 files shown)


lachlan-00:
happy to do this should we make the date an ISO 8601 date though? `date("c", time())`

if it's just to differentiate for the user and not used by rss clients then this is fine as is.

cquike:
Yes, it is for user information, the rss clients do not parse it.

lachlan-00:
all good then!

PR #3156, opened 2022-02-07
fix wrong translation

teknoraver:
`Disabilitato` means `disabled`, not `disable`


mitchray:
Unfortunately that file you edited is a compiled file, all translations need to be done through Transifex: https://www.transifex.com/ampache/ampache/

If you are able to make an account there and update the translation it would be most helpful

lachlan-00:
Thanks for creating @teknoraver. Transifex acts as the translation source and is merged back into Ampache from there every month or so.

If you join the team and make your changes there it will be merged back in.

teknoraver:
joined, fixed

lachlan-00:
thanks for helping us out!

PR #3159, opened 2022-02-11
subsonic: Do not send convertArt for album if art does not exist

### LINKED ISSUE #3149: Subsonic: do not return the coverArt id if it doesn't exist
### Describe the bug

When using the subsonic API a request like this /rest/getAlbum.view?id=200000XXX will return a JSON object which includes the coverArt field _even_ if that album doesn't have a cover. That leads to potential errors in the client when they try to access the cover with getCoverArt. Since the coverArt field in the subsonic API is optional I think it would be better not to fill it when there is no cover. 

## Environment

* Ampache version: 5.1.0
* Web server + version: apache
* Server operating system: Raspberry os 11
* Client operating system: Debian 11

### Client type

* SubSonic (strawberry player)

cquike:
This PR will avoid sending the cover art for albums that do not have art. This prevents errors when the client will ask for a non-existing art.
It would fix #3149.

### DIFF
--- a/src/Module/Api/Subsonic_Xml_Data.php
+++ b/src/Module/Api/Subsonic_Xml_Data.php
@@ -691,7 +691,9 @@ class Subsonic_Xml_Data
         $xalbum->addAttribute('isDir', 'true');
         $xalbum->addAttribute('discNumber', (string)$album->disk);
 
-        $xalbum->addAttribute('coverArt', 'al-' . self::getAlbumId($album->id));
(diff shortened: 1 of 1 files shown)


lachlan-00:
might be good to turn it into a lookup like $object->get_has_art() ?

it's probably something i should have done already. i'll merge this and then update the other objects

cquike:
Thanks!
Yes, that makes a lot of sense, to have a function has_art() for every object.

PR #3191, opened 2022-03-02
Translation Update March 2022

Psy-Virus:
Translation Update March 2022


lachlan-00:
@Psy-Virus with 982ff9c28 i've found that some of the xhtml templates do not have php T_() translations so they will have been dropped from transifex.

I've updated gather messages and added a text file with the missing strings. i'll do another quick check to make sure strings are in.

if i update the pot file that should sync with transifex?

Psy-Virus:
Yepp, Transifex pulls the pot in some interval automatically if changed.
But wouldn't it make more sense to put the ```xhtmltxt='../../resources/untranslated-strings.txt'``` into the ```locale/base``` dir to keep all the translation stuff together?

lachlan-00:
i was thinking to keep it with the templates in resources but either way is fine with me so i can put it there.

lachlan-00:
moved and set with a42e4dd40

PR #3239, opened 2022-04-30
Improvements to shuffle on the web player

### LINKED ISSUE #3212: [Feature Request] Web Player Track Shuffle Improvement
When using the web player, I would like the shuffle function to keep the current track playing without interruption, and shuffle the other tracks in the play queue behind the currently playing track.

An example of this is in most Android Subsonic clients like Dsub, which does the same thing when the shuffle button is pressed on the play queue.  The current song keeps playing and all other songs in the queue (even songs that have already been played) are shuffled below it.

troycarpenter:
Change the shuffle behavior so that:
1. Current selected track will become the first track and everything else shuffled below it.
2. New shuffled playlist isn't automatically played.  This means that if a song was playing, it will continue to play.  If nothing was playing, the shuffled playlist does NOT start playing.

Closes #3212


troycarpenter:
This pull request is for feature #3212

troycarpenter:
The jplayer.playlist.min.js was created with an online tool, however, now that I see the diff it is quite different than what was there before in terms of variable names and such.  If necessary, I can find the added code from my commit and make it match the original better.

lachlan-00:
if you've let maintainers edit the pull i can handle that part.

looking forward to testing it out cause the webplayer has really needed this kind of focus

troycarpenter:
How do I be sure the maintainers can edit?

lachlan-00:
now that is a great improvement. just going to test the min and we good.

lachlan-00:
congrats on becoming the latest Ampache contributor!

troycarpenter:
Thanks!  Already working on the next thing...the implementation of the catalog access lists feature.

PR #3258, opened 2022-05-08
Fix db query error in xbmc localplay

lusum:
The correct name of the field is pass, not password


lachlan-00:
must be a copy paste from the other localplays

Issue #2934, opened 2021-06-03
The next song always jump back to the first song in the list in Subsonic clients when Ampache server used in Jukebox/Localplay mode.


gerroon:
### Describe the bug

I am using Ultrasonic with Ampache in Subsonic API. It works fine in general but there is an issue with localplay and jukebox mode (the client controls it)

I can load bunch of songs in jukebox (which is localplay in Ampache) mode in Ultrasonic or Dsub. I can play the songs in localplay mode from the clients, but everytime I pres next, it goes back to the first song which you can see in the Ampache logs below too.

I can however move to the next song alright if I just use the local Mpd controller like Ncmpcpp. So this must be something todo with the Subsonic Api rather than the actual Mpd controller or the Subsonic clients.

It would be super nice if this functionality is supported properly since it makes it easy to control it from a mobile device as well. it seems like it is almost almost there.

### To reproduce

1. Download Dsub or Ultrasonic
2. Add you Ampache as a server
3. add bunch of songs
4. Select jukebox mode in Ultrasonic or casting in Dsub
5. press next to try to load the next song.

### Expected behavior
it should play the playlist properly.

## Environment

* Ampache version: 4.4.2
* Web server + version: Apache2 2.4.46
* Server operating system: Debian Linux
* Client operating system: Android

### Client type

* SubSonic

## Logs
```
2021-06-03T00:21:03-05:00 [USER] (Stats) -> Repeated the same Song too quickly (168/559s), not recording stats for {392999} 
2021-06-03T00:21:08-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.0 
2021-06-03T00:21:08-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.0 
2021-06-03T00:21:13-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.0 
2021-06-03T00:21:13-05:00 [USER] (Localplay) -> Error: Unable to skip to next song, check mpd controller 
2021-06-03T00:21:14-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.0 
2021-06-03T00:21:19-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.0 
2021-06-03T00:21:24-05:00 [USER] (Subsonic_XML_Data) -> API fail in function jukeboxcontrol-1.13.

[...]

PR #3261, opened 2022-05-09
Fix autoplay in xbmc localplay

lusum:
Not perfect patch, but a working one.
Fixes autoplay when playlist is new.
Do not autoplay when another song is playing ( changing playlist ) , cause both have the same id.
It should be better to place getActivePlayers in the status, but i have no idea ho to do it.


lusum:
Partially solves #2934,  index.php seems due to url rewrite of something similar, not checked really ( different behaviour between release and develop )

PR #3330, opened 2022-08-27
Fix album time update when time is NULL

Simounet:
Hi there,
Thanks a lot for this amazing software I'm using for such a long time!
The initial issue was that new album didn't get the time column filled in.
I fixed the album time value not updated when the initial value is `NULL`. The initial value should be `0` instead of `NULL` if we expect an integer, don't your think?
Have a nice day!


lachlan-00:
i didn't realise just how many of my albums had no time! that's a good find

PR #3372, opened 2022-10-22
jQuery purge, sidebar & html5_player_headers

mitchray:
Beginning purge of non-plugin jQuery (i.e. jPlayer)

### sidebar
- Replaced jQuery slideToggle (only instance in Ampache) with CSS max-height animation
- Cookie set/get tested and working
- No more delay on arrow rotation

### webplayer
- Replaced jQuery slideUp/slideDown with [Web Animations API](https://caniuse.com/web-animation) 
  - Preserves visibility checks used elsewhere which rely on 'display'

### DIFF
--- a/public/templates/show_html5_player_headers.inc.php
+++ b/public/templates/show_html5_player_headers.inc.php
@@ -92,8 +92,9 @@ if ($iframed || $isShare) { ?>
 <script>
     function ExitPlayer()
     {
-        $("#webplayer").text('');
-        $("#webplayer").hide();
+        var webPlayer = document.querySelector("#webplayer");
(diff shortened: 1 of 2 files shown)


lachlan-00:
seems to stop working once you change the tab between sidebars and the state of the cookie seems to be ignored when closed?
image

lachlan-00:
i wonder as well if the jquery is so out of date and that's what killed hte message notifications.

mitchray:
Ah yep I know why, will update it

mitchray:
I'll give jQuery this, the way its selector engine works with ajaxed content is quite useful... Have fixed the issues but I'm going to rethink this

Issue #3386, opened 2022-11-21
Thumbs - Not the same if they are on disk or in database


fufroma:
In this function:

https://github.com/ampache/ampache/blob/9c9f5901522cb8b78f086cb214a84e91c951809b/src/Repository/Model/Art.php#L249

```php
        while ($results = Dba::fetch_assoc($db_results)) {
            if ($results['size'] == 'original') {
                if (AmpConfig::get('album_art_store_disk')) {
                    $this->raw = self::read_from_dir($results['size'], $this->type, $this->uid, $this->kind, $results['mime']);
                } else {
                    $this->raw = $results['image'];
                }
                $this->raw_mime = $results['mime'];
            } elseif (AmpConfig::get('resize_images')) {
                if (AmpConfig::get('album_art_store_disk')) {
                    $this->thumb = self::read_from_dir($results['size'], $this->type, $this->uid, $this->kind, $results['mime']);
                } elseif ($results['size'] == '275x275') {
                    $this->thumb = $results['image'];
                }
                $this->raw_mime = $results['mime'];
            }
            $this->id = (int)$results['id'];
        }

```

There are two case:
- if the thumbs are stored in database: return the 275x275 thumb
- if the thumbs are on disk: iterate on each thumb listed in database, each time overwrite $this->thumb with the blob. Return the last one.

There are three issues with this:
- the thumb returned aren't the same size depend of the storage media (either 275x275 or random)
- when stored on disk, we iterate on each file, this is a performance issue.
- when stored on database, we return no thumb if there are no 275x275 thumb existing.

Edit: another issue : where stored on file, if the last thumb item of the database have no corresponding file, $this->thumb is null 

A first dummy patch to address the performance issue:
```diff
@@ -261,6 +261,7 @@ class Art extends database_object
                 }
                 $this->raw_mime = $results['mime'];
             } elseif (AmpConfig::get('resize_images')) {
+               if ( ! empty($this->thumb)) continue;
                 if (AmpConfig::get('album_art_store_disk')) {
                     $this->thumb = self::read_from_dir($results['size'], $this->type, $this->uid, $this->kind, $results['mime']);
                 } elseif ($results['size'] == '275x275') {

```
This way we stop to read files as soon as we have a thumb. 

The two other issue still exist, but we must choose what the function should do. I rather let you decide :)

fufroma:
@lachlan-00 Maybe I could just make a PR for the light dummy patch, and we'll address the global issue later?

[...]

PR #3390, opened 2022-11-21
Fix 304 for images

fufroma:
Hi,

I found several issues in the actual implementation of the ETag header to enable 304 responses.

- Based on file https://github.com/ampache/ampache/blob/2ca159ba032fd1921438b27a4eb4b280f4719d0c/src/Module/Application/Image/ShowAction.php : ETag is compared at line L195 but is modified at line L230 (yes there is an explode at line L192 but, still, it isn't the same image).
- ETag should be betwen `"`
- 304 responses must have ETag header, or the next answer won't be cached.
- Expires headers is recommended with ETag

### DIFF
--- a/src/Module/Application/Image/ShowAction.php
+++ b/src/Module/Application/Image/ShowAction.php
@@ -230,7 +230,7 @@ final class ShowAction implements ApplicationActionInterface
             if (!empty($etag)) {
                 $response = $response->withHeader(
                     'ETag',
-                    '"'.$etag.'"'
+                    '"' . $etag . '"'
                 )->withHeader(
                     'Expires',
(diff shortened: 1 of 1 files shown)


fufroma:
Glad to be of help! And thank you for the quick review!

I'm not a daily user of Ampache, but I work with @Afterster on the big https://play.dogmazic.net/ instance, so I may find some useful things along the road :)

fufroma:
That's very kind of you, thank you :bow: 

Don't worry, all the features and/or question I may have will be a PR or an issue. I already made a few :wink: 
- https://github.com/ampache/ampache/pull/3392
- https://github.com/ampache/ampache/issues/3394
- https://github.com/ampache/ampache/issues/3386

PR #3407, opened 2022-12-07
Update Upnp_Api.php

drder66:
Fix deprecated old php syntax no longer acceptable at line 295, update error message 2 lines later


drder66:
Working to fix uPnP functionality; another change coming for PlayAction.php
Relates to issue #1547

PR #3421, opened 2022-12-19
Subsonic_Api: Allow empty queries

### LINKED ISSUE #3420: Subsonic API: "empty" search3 query not properly supported
### Describe the bug

[Symfonium](https://symfonium.app/) uses the Subsonic search3 API early in its Sync process. The [query parameter is non-option](http://www.subsonic.org/pages/api.jsp#search3) so the developer has chosen to use an empty query (i.e., "") instead. This returns nothing because the current code interprets `"something"` as an equals-title clause, and very few titles are empty strings.

### To reproduce

`curl https://<server.com>/rest/search3.view?query=""<&base parameters>``

### Expected behavior

The `searchResult3` response is not empty.

## Environment

* Ampache version: 5.5.5-release
* Web server + version: N/A
* Server operating system: N/A
* Client operating system: Android N/A

### Client type

* SubSonic

Symfonium 3.0.0a

LoganK:
This allows clients such as Symfonium to sync properly.

Preserve the existing equals logic primarily because I am not familiar with common client patterns.

fixes #3420

### DIFF
--- a/src/Module/Api/Subsonic_Api.php
+++ b/src/Module/Api/Subsonic_Api.php
@@ -1056,7 +1056,11 @@ class Subsonic_Api
         $query    = $original;
         if (substr($original, 0, 1) == '"' && (substr($original, -1) == '"')) {
             $query    = substr($original, 1, -1);
-            $operator = 4; // equals
+            // query is non-optional, but some clients send empty queries to fetch
+            // all items. Fall back on default contains in such cases.
+            if (strlen($query) > 0) {
+                $operator = 4; // equals
+            }
         }
         if (substr($original, 0, 1) == '"' && substr($original, -2, 2) == '"*') {
             $query    = substr($original, 1, -2);


lachlan-00:
good to know, i see a lot of clients use an empty search as a list a lot more than i thought.

Welcome as a new contributor!

Tolriq:
@lachlan-00 Hi, I'm Symfonium author, a modern music player that support many different providers.

Does Ampache have someone in charge of the API and subsonic part? Since most providers have improved the search3 endpoint, I'd like to start a discussion with the major actors on extending the Subsonic API to support the missing parts to allow client apps to match what other media providers already support.

If yes and have some interest it would be nice if he could register at https://support.symfonium.app/t/expanding-subsonic-api/1115 and already post his nickname, there's already a couple of other providers registered there that I'll ping when I write the needs if enough interest.

lachlan-00:
I've put some of my initial thoughts on that thread.
https://support.symfonium.app/t/expanding-subsonic-api/1115/1

Subsonic is falling behind in Artist queries and the auth is a problem with me as it seems like it's kept in plaintext in the DB

> @lachlan-00 Hi, I'm Symfonium author, a modern music player that support many different providers.
> 
> Does Ampache have someone in charge of the API and subsonic part? Since most providers have improved the search3 endpoint, I'd like to start a discussion with the major actors on extending the Subsonic API to support the missing parts to allow client apps to match what other media providers already support.
> 
> If yes and have some interest it would be nice if he could register at https://support.symfonium.app/t/expanding-subsonic-api/1115 and already post his nickname, there's already a couple of other providers registered there that I'll ping when I write the needs if enough interest.

I've put some of my initial thoughts on that thread.
https://support.symfonium.app/t/expanding-subsonic-api/1115/1

Subsonic is falling behind in Artist queries and the auth is a problem with me as it seems like it's kept in plaintext in the DB (or it's kept as an md5?)

As far as interest goes i think it's a solid API that needs to be maintained and i'm happy to extend and maintain Subsonic support beyond the 1.16.1 on subsonic.org in Ampache

Tolriq:
@lachlan-00 don't know if the forum email goes to spam but have you seen https://support.symfonium.app/t/subsonic-servers-participation/1233? If you are out of the project please just confirm.

lachlan-00:
sorry, i'm not around lately with work and home. i'm replying in the affirmative now

PR #3464, opened 2023-02-19
Art response was hardcoded to artist

mitchray:


lachlan-00:
that would explain why it would randomly error in scripts

PR #3469, opened 2023-02-22
Add possibility to export just one playlist

fufroma:
`./bin/cli export:playlist /tmp/ezstream-tempo/ m3u playlists` export **all** playlist. Can be quite long.

This patch add an optional parameters to allow export of just one playlist. Example: 
`./bin/cli export:playlist /tmp/ezstream-tempo/ m3u playlists 42`


lachlan-00:
i'm going to make it slightly different for the string and the logic on adding but not needed to change here

PR #3485, opened 2023-03-16
Various quick fixes / features

fufroma:
Lot of quick fixes, most of them based on what /var/log/ampache/ampache.log warned me.

Sorry to push a mess like that.

Also, a few features.

Features:
- Add possibility to have a custom.js
- Add header to allow browser cache on waveform
- Add filename properties to song object
- Add an easy optimisation when it's the guest user
- Add log event when preventing Administrator to reset their password

Fix:
- Add missing doctype
- Handle $user is null
- Handle if there are no info for the user
- Handle bad object
- Handle if the json can't be decoded
- Force cast on some value before doing arithmetic
- Display filename and not the whole path and filename
- Partial fix of https://github.com/ampache/ampache/issues/3386 - Better, but not good

### DIFF
--- a/src/Application/Api/Ajax/Handler/SongAjaxHandler.php
+++ b/src/Application/Api/Ajax/Handler/SongAjaxHandler.php
@@ -83,10 +83,11 @@ final class SongAjaxHandler implements AjaxHandlerInterface
                     echo "shouts = {};\r\n";
                     foreach ($shouts as $shoutsid) {
                         $shout = new Shoutbox($shoutsid);
-                        $key   = (int) ($shout->data);
+                        $key   = (int)$shout->data;
+                        $time  = (int)$media->time;
                         echo "if (typeof shouts['" . $key . "'] === 'undefined') { shouts['" . $key . "'] = new Array(); }\r\n";
                         echo "shouts['" . $key . "'].push('" . addslashes($shout->get_display(false)) . "');\r\n";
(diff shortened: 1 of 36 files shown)


fufroma:
@lachlan-00 feel free to amend or criticize those patch, do not hesitate if there are thing I've done the wrong way

lachlan-00:
do you know what data column in the user_shout table is for? it doesn't seem to be used on any of the links in the site

I was assuming something like a timestamp but i haven't seen anything about it. might remove it entirely

src/Application/Api/Ajax/Handler/SongAjaxHandler.php
lachlan-00, line 1: not really sure about what $shout->data actually is?

lachlan-00:
k, pretty much done here i'll run it on my server this week and merge. i use the filename path in the song view a lot and not having the folder would freak out my mind so i'll see what performance is like adding the folder in as a property as well

src/Application/Api/Ajax/Handler/SongAjaxHandler.php
fufroma: Neither do I. There was warning in the ampache.log, so I fixed it...

fufroma:
> k, pretty much done here i'll run it on my server this week and merge. i use the filename path in the song view a lot and not having the folder would freak out my mind so i'll see what performance is like adding the folder in as a property as well

I saw filename when it was path + filename, and I though that web user don't need to know where the files are (except for the sysadmin).

Thank you for your patch, I'll take notes for my next one!

lachlan-00 (approved):

PR #3526, opened 2023-05-13
Fix htaccess parameters to pass valid characters

crees:
Since Apache r1908095 [1], it has become far less forgiving of 'invalid characters' in Rewrite patterns, for example patterns with spaces.

Songs often have spaces in the filenames, so it is necessary to pass them.

[1] https://svn.apache.org/viewvc?view=revision&revision=1908095

[2] https://webmasters.stackexchange.com/questions/141837/ah10411-rewritten-query-string-contains-control-characters-or-spaces

Please note, I'm not really an expert with the Apache incantations, so I've indiscriminately added the B and BNP flags to every RewriteRule.  I think that's necessary and it works, but it might need a basic sanity check.

### DIFF
--- a/public/daap/.htaccess
+++ b/public/daap/.htaccess
@@ -2,5 +2,5 @@
     RewriteEngine On
     RewriteCond %{REQUEST_FILENAME} !-d
     RewriteCond %{REQUEST_FILENAME} !-s
-    RewriteRule ^(.+)$ /index.php?action=$1 [PT,L,QSA]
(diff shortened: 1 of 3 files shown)


lachlan-00:
B and BNP should be okay in all of them.

Depending on config spaces will be escaped to + before being sent to the url but that's probably helpful to make sure it's always escaped to something.

I'll test them out on my home servers first and see how they respond.

lachlan-00:
sorry that took so long to get to but you're our newest contributor!

thanks for looking at this and putting the time into making Ampache better

PR #3543, opened 2023-06-23
Set login_message from config

mitchray:
"login_message" is mentioned on the login page but isn't present in the config, this fixes that

firefox_sNpumgEcld


kuzi-moto:
Nice catch!

PR #3569, opened 2023-08-06
Server would not return correct url if 'local' is true

mitchray:
download/stream API methods were fine on my local dev setup, but on my remote server (Docker with curl installed) they were returning ```http://localhost/play/index.php?ssid...``` due to ```function_exists('curl_version'),``` being ```true```

The default for ```play_url()``` is ```local: false``` so after removing the checks the correct URLs were returned

After this change the only other instances of ```local``` being checked or forced are these places:

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Daap_Api.php#L437

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Subsonic_Api.php#L1396

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Upnp_Api.php#L1713


mitchray:
Ugh git

mitchray:
Apologies I don't quite understand why its including these other ocmmits

PR #3570, opened 2023-08-06
Server would return incorrect url if 'local' was true

mitchray:
download/stream API methods were fine on my local dev setup, but on my remote server (ampache-docker which comes with curl installed) they were returning ```http://localhost/play/index.php?ssid...``` due to ```function_exists('curl_version'),``` being ```true```

The default for ```play_url()``` is ```local: false``` so after removing the checks the correct URLs were returned

After this change the only other instances of ```local``` being checked or forced are these places:

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Daap_Api.php#L437

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Subsonic_Api.php#L1396

https://github.com/ampache/ampache/blob/e9408ed52c214680b650a803f653c29dfb86a753/src/Module/Api/Upnp_Api.php#L1713

### DIFF
--- a/src/Module/Api/Method/Api4/Download4Method.php
+++ b/src/Module/Api/Method/Api4/Download4Method.php
@@ -69,11 +69,11 @@ final class Download4Method
         }
         if ($type == 'song') {
             $media = new Song($fileid);
-            $url   = $media->play_url($params, 'api', function_exists('curl_version'), $user->id, $user->streamtoken);
+            $url   = $media->play_url($params, 'api', false, $user->id, $user->streamtoken);
         }
         if ($type == 'podcast_episode' || $type == 'podcast') {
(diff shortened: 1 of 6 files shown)


lachlan-00:
this will all be holdover stuff from old code but i'm not sure why this would need to be forced with curl.

lachlan-00:
for things like docker should we actually be looking at local_web_path instead?

mitchray:
I don't think so as these methods need to serve a 'public' URL

lachlan-00:
i can't think of a reason why this would be a problem. so lets do it.

Issue #3537, opened 2023-06-12
Embedded Album cover art not loaded


rhauff1:
### Describe the bug

When adding new music to the database, the cover art is not automatically loaded, and now cannot load manually either.

### To reproduce
Add new music (with cover art embedded) to database
View album, no artwork is shown.
Click on tool on image to Edit/Find Art.
Under Cover Art Search, click "Get Art"
Artwork loads, and first image on left which should be embedded art (I believe) shows "Invalid"

### Screenshots

![Edit-Find_art](https://github.com/ampache/ampache/assets/75855702/06bd4c7e-8ff0-4bee-aaea-ccca0d88f572)
![Invalid_art](https://github.com/ampache/ampache/assets/75855702/24501e9b-4a62-4192-990c-680f7dfff25a)

## Environment

* Ampache version:  Ampache 5.6.0-release
* Web server + version:  Apache/2.4.52 (Ubuntu)
* Server operating system:  Ubuntu 22.04.2 LTS
* Client operating system:  Ubuntu 22.04.2 LTS

### Client type

* Web
* Ampache
* SubSonic

## Logs
Snippet of Ampache log that is generated by the above steps:

```2023-06-12T14:11:34-05:00 [ampache] (Ampache\Module\System\Dba) -> Database connection...
2023-06-12T14:11:34-05:00 [ampache] (Ampache\Module\System\Dba) -> Database connection...
2023-06-12T14:11:34-05:00 [roland] (Ampache\Module\System\Session) -> qh5vl3bonm2uht8t3e58chim6d has been extended to Mon, 12 Jun 2023 15:11:34 -0500 extension length 3600
2023-06-12T14:11:34-05:00 [roland] (ajax.server) -> Called for page: {index}
2023-06-12T14:11:34-05:00 [roland] (Ampache\Module\System\Session) -> Writing to qh5vl3bonm2uht8t3e58chim6d with expiration 1686600694
2023-06-12T14:11:54-05:00 [ampache] (Ampache\Module\System\Dba) -> Database connection...
2023-06-12T14:11:54-05:00 [ampache] (Ampache\Module\System\Dba) -> Database connection...
2023-06-12T14:11:54-05:00 [roland] (Ampache\Module\System\Session) -> qh5vl3bonm2uht8t3e58chim6d has been extended to Mon, 12 Jun 2023 15:11:54 -0500 extension length 3600
[...]

PR #3572, opened 2023-08-15
Fix for #3537: Prepend 'ID3 ' to the art title for MP3 tags

scufre:
Prepend the art title for MP3 tags with `ID3 ` as it was missing and then failing to fetch the embedded art from said files.
Also made a minor refactor to simplify the code.

### DIFF
--- a/src/Repository/Model/Art.php
+++ b/src/Repository/Model/Art.php
@@ -981,6 +981,30 @@ class Art extends database_object
     } // generate_thumb
 
     /**
+     * get_image_from_tags
+     * @param array $images
+     * @param string title
+     * @param string type_key
+     * @param array type_keys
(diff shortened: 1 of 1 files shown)


mitchray:
#3572

lachlan-00:
Thanks for taking the time to make a pull @scufre !

I would first up merge $type_key and $type_keys into a single array parameter.

There might also be an easier way to do the foreach in the foreach too seeing as it's only done once but i'll have a look tomorrow when i can run it on my home server.

If you make any updates before i get to it tomorrow just make sure to run `./resources/scripts/tests/codestyle.sh` to clean up that return as well.

lachlan-00:
Boom. Welcome to the contributor club @scufre

PR #3574, opened 2023-08-17
Prefer cover art when gathering art from embedded tags

scufre:
Sort found art in tags by a priority list (album or artist).
Refactor to remove some duplicate code to gather art in tags which also addresses in a different way #3537.

### DIFF
--- a/src/Module/Art/Collector/MetaTagCollectorModule.php
+++ b/src/Module/Art/Collector/MetaTagCollectorModule.php
@@ -29,7 +29,6 @@ use Ampache\Config\AmpConfig;
 use Ampache\Repository\Model\Art;
 use Ampache\Repository\Model\Song;
 use Ampache\Repository\Model\Video;
-use Ampache\Module\System\LegacyLogger;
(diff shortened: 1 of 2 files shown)


lachlan-00:
one thing i think we've done now is make all artists pick up the front cover now?

Having it wide open for all the types means that artist search is going to fill up with front covers

lachlan-00:
would there be different priorities based on the $art->type? (artist/album/song)

scufre:
> would there be different priorities based on the $art->type? (artist/album/song)

Missed this.  Thanks for pointing this out.

Just pushed some updates where I introduced a priority list for artist, also changed a bit to not filter art when looking for artists and instead sorting it using its corresponding priorities and additionally moved some very similar code used to gather art from tags to a common place which ends up solving #3537 in another way (I'd say this is better as I think the issue was caused by having two ways of fetching art from tags for different stages which ended up being slightly different.

lachlan-00:
cleans up good. just moved things and ran php-cs and merging now!

PR #3656, opened 2023-11-13
Fix

fufroma:
Quick fix after Dogmazic upgrade


lachlan-00:
Just fixed up the merge errors. thanks for putting them all here in one go.

fufroma:
Well, let's merge if that's OK for you, I'll make another PR after :)

src/Repository/Model/Tmp_Playlist.php
```diff
@@ -44,6 +44,7 @@ class Tmp_Playlist extends database_object
     public $type;
     public $object_type;
     public $base_playlist;
+    public $created;
```
lachlan-00, line 47: with created where is that property coming from? is that a custom column in your tmpplaylist table?

fufroma: no idea, it was here before my arrival :smile: 

I have the same dumb approach as usual: if there are error in ampache.log or if ampache is slow, I look for a patch.

lachlan-00: next pr try and see where your created porperty is from but it's no biggie for now

Aisyk: > with created where is that property coming from? is that a custom column in your tmpplaylist table?

Hi,
That it is because the playlists are also created by an user and that one can have private or public playlists.

lachlan-00: the tmp_playlist is used for adding songs to the rightbar in a per-session temporary list. it's also used to store votes for democratic lists.

I've had a look on the older releases and from v3-6 the table has always been.
```
CREATE TABLE IF NOT EXISTS `tmp_playlist` (
  `id`,
  `session`,
  `type`,
  `object_type`
)
```

would be good to know what your tmp_playlist table looks like/if it's in any patches you apply/code

fufroma: Found it! It's a remaining of a really old hack. 

`created datetime DEFAULT current_timestamp(),` save _when_ the tmp_playlist was created, and there was a cron job to trash old tmp_playlist (there was a time with thousand of tmp_playlist, it was painful for the server).

I think I can trash it :)

PR #3705, opened 2023-12-03
Bugfix: anti-spam feature for reset password was too strict

fufroma:
Mail for reset password was never sent for user using it for the first time : if there was no key for the "reset_password" in "user_data", the default value was "now", so the anti spam feature stopped the password reset.


lachlan-00:
good one. that's the only one that wasn't falling back to 0 so i've set this get_user_data to allow a default value so it will always send now.

PR #3769, opened 2024-01-16
Use "Status" string rather than "State" which can be confused by Country

cquike:
This PR changes the string "State" by "Status" in the case it is meaning a condition of something. This workarounds some translation issues. Since the text to translate is the same in case "State" means condition or country, transifex shows a single string to translate with the possibility to have only one translated item, which in turn makes some translations wrong.


lachlan-00:
english is great sometimes

PR #3781, opened 2024-01-22
xbmc localplay : add pause status, play/pause/stop now follow localplay standard

lusum:
xbmc localplay : add pause status, play/pause/stop now follow localplay standard


lusum:
I have to change something, i will add a new pull request

PR #3794, opened 2024-01-25
Localplay and jukebox mode improvements ( Vlc and Xbmc )

lusum:
* Correct bugs in localplay VLC and XBMC
* Enable index in every localplay client ( clients have to handle it )
* Add sleep in XBMC function to handle delay in kodi client response


lachlan-00:
Thank you for your continued effort here @lusum

Issue #3833, opened 2024-02-20
Wrong artist name in a multiartist album 


cquike:
## Description

I have an album which has two artists. I have added metadata with picard which results in the following (among others) tags:
```
    comment[2]: ARTIST=Johann Christian Bach
    comment[7]: ALBUMARTISTSORT=Concentus Musicus Wien, Harnoncourt, Nikolaus
    comment[13]: ALBUMARTIST=Concentus Musicus Wien, Nikolaus Harnoncourt
    comment[14]: MUSICBRAINZ_TRACKID=36dafc10-1413-4b3a-8d43-8016f65b6d08
    comment[15]: ARTISTS=Johann Christian Bach
    comment[16]: ARTISTSORT=Bach, Johann Christian
    comment[17]: MUSICBRAINZ_ARTISTID=470967b1-f9ef-4712-82df-6971b517ebdb
    comment[18]: MUSICBRAINZ_RELEASEGROUPID=309c0745-655d-3d25-8998-5db1b5611af9
    comment[20]: MUSICBRAINZ_ALBUMID=04991fc4-3e14-47de-81c6-1a48105a17b7
    comment[21]: MUSICBRAINZ_ALBUMARTISTID=9b891046-35af-4eb0-a058-eba4c9b8d01f
    comment[22]: MUSICBRAINZ_ALBUMARTISTID=98b95966-64db-4631-8b9f-8aa66f32cc98
    comment[23]: MUSICBRAINZ_DISCID=grAa4zTRM7Rs6YC5Bd_jhVccOUc-
    comment[24]: MUSICBRAINZ_RELEASETRACKID=8a330fbc-b19e-337d-b620-4d9b3cf53857

```
As you can see there are two album artists: https://musicbrainz.org/artist/9b891046-35af-4eb0-a058-eba4c9b8d01f and https://musicbrainz.org/artist/98b95966-64db-4631-8b9f-8aa66f32cc98. ampache correctly identifies that there are two artists in the album. However it wrongly sets the name of the first one to the full ALBUMARTIST string, which contains the name of both authors:

![mannheim](https://github.com/ampache/ampache/assets/17937361/f38d5615-0d73-48ff-b02a-ada21b30686a)

That can also be seen in the database:
```
+-------+----------------------------------------------+--------------------------------------+
| id    | name                                         | mbid                                 |
+-------+----------------------------------------------+--------------------------------------+
[...]

PR #3834, opened 2024-02-20
Use the name provided by musicbrainz rather than the ALBUMARTIST tag

cquike:
This PR is a potential solution for #3833 . It basically sets the name that musicbrainz provides when creating a new album artist.

I don't find this solution particularly elegant for a couple of things:

1. It duplicates code that is found in function` check_mbid()`.
2. It assumes that the musicbrainz plugin is enabled. However this is already the case for `check_mbid()` function, so maybe it is the time to promote the plugin to be part of the core, since ampache already relies on musicbrainz metadata extensively for its correct behaviour.

Maybe this can be merged for the time being to fix the bug, but in the long run it is probably better to refactor the logic when creating a new artist to avoid a cleaner code.

BTW, other possibility to solve #3833 is to "properly" parse the `ALBUMARTIST` tag. However I find that prone to error the general case. If no mbid are available then that's the only option, but if mbid do exist then I think the best option is to rely on a proper musicbrainz query.

### DIFF
--- a/src/Repository/Model/Artist.php
+++ b/src/Repository/Model/Artist.php
@@ -706,9 +706,9 @@ class Artist extends database_object implements library_item, GarbageCollectible
 
         // prefer the name of the artist as provided by MusicBrainz
         if (!empty($mbid)) {
-            $plugin = new Plugin('musicbrainz');
+            $plugin      = new Plugin('musicbrainz');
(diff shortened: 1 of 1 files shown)


lachlan-00:
merging into 7 now. we'll make musicbrainz more important in that branch

PR #3835, opened 2024-02-22
Add specific error if the SQL query is empty

fufroma:
Sometimes there is an error about bad syntax of the SQL query, but it's because the query is empty (sometimes it's just space).

Add specific error in the log about that. Will allow further debug later.


lachlan-00:
just moved it out of the try loop so it doesn't do it 4 times

PR #3852, opened 2024-03-04
Move javascript libraries out of composer and into npm packages

mitchray:
- Composer can return to focusing on just php packages
- NPM will take over all JS packages (all packages remain at their current versions except jquery-ui, see below)
- Bumped jquery-UI from 1.12 to 1.13 as they didn't include built dist files until then
- Removed prettyphoto from repo, will pull the prettyphoto fork in automatically (depends on https://github.com/lachlan-00/prettyphoto/pull/1)
- Removed kumailht/responsive-elements as not in use
- jquery.contextMenu map files included
- Fairly certain the case checking for jQueryKnob, jQueryFileUpload and jQueryContextMenu is no longer needed

Tested each one as I went and no problems. Will make it that much easier to remove jquery plugins in future (looking at you jquery-knob which is only used on the upload file screen as a progress bar...)


lachlan-00:
Would it be good to extend the composer update commands to include an npm command? so then people composer update and composer npm

We'll probably need to add an update command into the config file like `composer_binary_path` and then add in the command to src\Module\System\AutoUpdate.php::update_dependencies

then for the switchover how do people update? will it clean the installed paths when they composer install and then the npm will add the missing ones? ideally people would be aware of the changes but autoupdate does the work but i'd also say not everyone would have npm so we might need a message in 6 servers when the release is closer

mitchray:
Sure I have added those update parts in now

Also added composer.lock back to the repo now we're back to a single composer.json

mitchray:
```composer install``` will purge the old files from 'vendor' (not that it would cause problems by remaining)

Definitely would need a message informing autoupdaters of the new npm requirement

lachlan-00:
all looks good, i haven't updated any of my local sites to the 7 branches yet but a warning message will work which i can do later.

when i do a final 6 before 7 i'll put a change message in like i did for 5=>6

lachlan-00:
@usox this will affect you when updating.

usox:
Awww sweet! Thanks @mitchray for removing all that clutter from composer.json :heart:

PR #3854, opened 2024-03-05
New file uploader

mitchray:
- Replaces jquery-file-upload with [Filepond](https://pqina.nl/filepond/)
- Removes jquery-knob
- Folder selection from jstree still works

https://github.com/ampache/ampache/assets/5735900/b5c8dd1a-2a1c-4b76-999f-705ff87915f1


lachlan-00:
now that's some slick looking upload. i like that it actually says it completed too

PR #3868, opened 2024-03-11
Align header elements; improve light/dark form controls

mitchray:
- Uniform height, font size and alignment of header elements
- Form control accent color takes on theme hue (see 'Random' checkbox)
- Dark mode form controls are now dark 🕶

### Light

Before

Screen Shot 2024-03-12 at 08 55 18

After

Screen Shot 2024-03-12 at 08 55 13

### Dark

Before

Screen Shot 2024-03-12 at 09 11 00

After

Screen Shot 2024-03-12 at 09 11 03

### DIFF
--- a/public/themes/reborn/templates/dark.css
+++ b/public/themes/reborn/templates/dark.css
@@ -24,6 +24,10 @@
    General style rules
    -------------------------- */
 
+:root {
+    color-scheme: dark;
+}
+
 body {
     color: #fff;
     background-color: #222;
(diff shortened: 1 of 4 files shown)


lachlan-00:
pretty

public/themes/reborn/templates/dark.css
```diff
@@ -48,12 +52,15 @@ hr {
...
 input[type=url],
 input[type=year],
+select,
 textarea {
-    background-color: #f5f5f5;
-    background-image: linear-gradient(to top, #eee, #fff);
-    border-color: #fff;
+    background-color: #000;
+    background-image: linear-gradient(to top, hsl(0, 0%, 8%), hsl(0, 0%, 4%));
+    border-color: #222;
```
lachlan-00, line 63: should we make this light? / delete it?
image

Not sure what the fallback colour is but boxes can be a bit hard to see with 222 set

mitchray: Give it

```
background-color: hsl(0, 0%, 10%);
background-image: linear-gradient(to top, hsl(0, 0%, 12%), hsl(0, 0%, 8%));
border-color: hsl(0, 0%, 30%);
```

lachlan-00: yeah that looks heaps better, i'll add that

PR #3870, opened 2024-03-12
swap to npm-run-all2

mitchray:


lachlan-00:
error looks the same after an npm update and npm install withing the updated lock/json

mitchray:
Can you delete the node_modules folder and retry?

lachlan-00:
will close cause the path quoting seems to fix this, i'm assuming the expanded contents is so large combining it into a string with "" stops the error

PR #3878, opened 2024-03-13
Allow to display song even if no data in song_data table

fufroma:
Can't display a song if it doesn't have any data in song_data table

Example : if 'select *  FROM `song_data` WHERE `song_id` = 57170;' is empty, the cache isn't populated.

It return false, and it must be an array

This patch allow false and array as return value.


lachlan-00:
to keep with php7.4 i'll keep them as arrays and check in the is_cached to make sure it's not loading false data earlier

PR #3880, opened 2024-03-15
Decode scrubbed entities back into user-facing (but sanitized) text

mitchray:
Perhaps counter-intuitively, we do want to scrub_out $rule[2], then convert any encoded chars back into 'normal' text after it has been sanitized


lachlan-00:
cool, now i'll just need to make sure the search gets the correct data as well.

There seems to be some extra stripping/encoding going on on the input part but i'll fix that up. i'll just have to make sure that input isn't being used as a direct input
image

lachlan-00:
i've had a quick look and removed the stripping it looks okay but i just need to confirm that all $input is parameterised.

PR #3882, opened 2024-03-15
Hide private playlist to non-admin and non-owner people

fufroma:
Private playlist where available to people even if not logged.

I don't think we wanted that.

### DIFF
--- a/public/templates/show_playlists.inc.php
+++ b/public/templates/show_playlists.inc.php
@@ -79,6 +79,11 @@ foreach ($object_ids as $playlist_id) {
     }
     $libitem->format();
 
+    // Don't show playlist if private AND you're not an Admin AND not the owner
+    if ( ! Access::check('interface', 100) && $libitem->type == 'private' && $libitem->get_user_owner() != $user_id ) {
+        continue;
+    }
+
     // Don't show empty playlist if not admin or the owner
     if (Access::check('interface', 100) || $libitem->get_user_owner() == $user_id || $libitem->get_media_count() > 0) { ?>
         <tr id="playlist_row_<?php echo $libitem->id; ?>">


fufroma:
Not sure what to do next...

Either we (my users and I) did not understand the use of private playlist, either we have a bigger problem because private playlist can be accessed without any auth with their playlist_id.

I'll remove the WIP tag, feel free to trash this MR if necessary.

fufroma:
I don't think I'll be able to do some more code for this, I think it may need deep changes to be done in a good way

@usox @lachlan-00 I'll be happy to have your idea for this.

usox:
So I think it would be quite good to implement a method in the playlist class that performs these access controls. I think one could use this in other places (e.g. not only for the list of playlists but also for direct access to a playlist object). So one could then reuse the code...

Something like this in the playlist-class:
```php
    /**
     * Checks if the given user is able to see/access the playlist
     */
    public function accessibleByUser(User $user): bool
    {
        if ($user->has_access(AccessLevelEnum::ADMIN)) {
            return true;
        }

        return !$this->isPrivate() ||
            ($user->getId() === $this->get_user_owner());
    }
```

fufroma:
I tried to add some access check in the constructor, but it mess with lot of GUI... I'm not used to patch the GUI, it's a little out of my depth.

usox:
I'm happy to do it for you. I've just been traveling in this area anyway and I think I know my way around now

fufroma:
> I'm happy to do it for you. I've just been working in this area anyway and I think I know my way around now

Thank you! I'll focus on things I can handle.

lachlan-00:
The browse should do this instead so you don't have to do any checks for it. i'll see what's happening.

lachlan-00:
oh, you mean people can open them with a direct link? that i don't think is set but i'll also check that

fufroma:
> oh, you mean people can open them with a direct link?

exactly :)

lachlan-00:
i've fixed that with 4155bc726 using has_access. now if you share a link to a playlist and don't have access it won't load.

for the dashboard pages there are no checks for privacy. these come from the stats class and i think they need to be reworked a bit to read public/privacy options
get_recent, get_top, get_newest functions need to probably be removed and sorting applied instead to the get_playlists function.

lachlan-00:
59c2a9871856f4063a0eee4e984b8f8684e6511d should close this out. the objects are pulled but not displayed if it fails access and private check.

lachlan-00:
set and releasing soon so i'll close

PR #3906, opened 2024-03-28
Fix #3761 overrided album name

### LINKED ISSUE #3761: Some albums are named after the artist and not the album name in 6.2
### Describe the bug
Some albums show the artist name instead of the actual album names.
I have both 6.1 and 6.2.1 running with the same music files, and this problem is specific to 6.2.
Some albums names are correct and others aren't with no obvious pattern.

### Screenshots

(In the 3rd screenshot, the song edit pop up shows the correct album name, but in the 4th screenshot, it get replaced by the artist instead)

6.1 : 

 

6.2.1 :

Simounet:
Hi there,
The artist name override the album name because of [the assign foreach here](https://github.com/Simounet/ampache/blob/73ce7cfe743e2a921e63fb954c3fd6fb56d39ab2/src/Repository/Model/Album.php#L439). This behavior doesn't seems to be very predictable to me.
I don't know when when the regression came in. I can search for it if you want me to.
Have a nice day.

### DIFF
--- a/src/Repository/Model/Album.php
+++ b/src/Repository/Model/Album.php
@@ -221,13 +221,13 @@ class Album extends database_object implements library_item, CatalogItemInterfac
         }
         $results = array();
         if (empty($this->album_artist) && $this->song_artist_count == 1) {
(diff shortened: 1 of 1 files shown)


usox (approved):

usox:
Thank you. This behavior of the format method - first writing the values directly into the properties and then doing it again in a loop afterwards - is just bad and incomprehensible.

PR #3916, opened 2024-04-04
Fix display license

fufroma:
Bad if


fufroma:
Introduced in Ampache 6.3.0 https://github.com/ampache/ampache/pull/3806

usox (approved):
:+1:

PR #3919, opened 2024-04-04
Some link are generated with &amp; instead of &

fufroma:
Some browser convert successfully, but crawlers aren't


lachlan-00:
is that specific for links? (<a>)
There are 44-80 links that could be affected by just checking with a regex like `<a.*&amp;` or just `&amp;`

lachlan-00:
@fufroma should i update all these <a> links where i've found &amps? or was this specific to these rows?

fufroma:
I think we must update all of them, but I didn't had the time to test it.

fufroma:
Work done in da6978ebf42bcd4a8bde747c586f6cc30de7e01b

PR #3920, opened 2024-04-04
Add some check in media_to_url()

fufroma:
Sometimes parameters are missing, must handle it.

Example : poorly generated url like https://FQDN/stream.php?action=download&amp;song_id=36849

### DIFF
--- a/src/Module/Playback/Stream_Playlist.php
+++ b/src/Module/Playback/Stream_Playlist.php
@@ -217,7 +217,7 @@ class Stream_Playlist
     {
         $type      = $media['object_type'] ?? null;
         $object_id = $media['object_id'] ?? null;
-        if ($type === null || $object_id === null){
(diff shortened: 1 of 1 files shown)


fufroma:
Before this patch  
```
[ampache] (log.lib) -> [Runtime Error] Undefined array key "object_type" in file /src/Module/Playback/Stream_Playlist.php(218)
[ampache] (log.lib) -> [Runtime Error] Undefined array key "object_id" in file /src/Module/Playback/Stream_Playlist.php(219)
[-1] (/src/Module/Util/ObjectTypeToClassNameMapper.php:96) -> Ampache\Module\Util\ObjectTypeToClassNameMapper::map(): Argument #1 ($object_type) must be of type string, null given, called in /src/Module/Playback/Stream_Playlist.php on line 220
```

PR #3930, opened 2024-04-12
Migrate PNG icons to Google Material SVGs

mitchray:
- All PNGs icons are replaced with Google Material SVGs (except ratings which use a CSS spritesheet, will make a new PR for that later)
- UI::_find_icon will prefer SVG over PNG in the fallback handling (SVG was only in the theme file search)
- The now unused PNGs will be removed from repo pending a future audit (some haven't been used since before Reborn I think)

Sidebar

firefox_9YzrVAWjo7
firefox_bP3losKEth

External links

firefox_W5Ft3u7Vlu
firefox_CbU0ViiImg

Player
firefox_2g7lNRm0Gn
firefox_usg9je4ivq

Album/Artist page

firefox_qR52GXgjTi
firefox_xczJPGmC5n


lachlan-00:
some of these icons are older than my kids.

lachlan-00:
I'll merge it without worrying about colour for now. the fill should be okay in most places (e.g. the fill on the upload page doesn't look odd on the buttons.

I think it's more likely that we've stared at these icons for so long the change feels weird.

lachlan-00:
with currentColor is there a way to override that on a parent div? like set it to something else on specific items

mitchray:
Yep easily done if desired

```css
.parentClass .material-symbol {
    fill: red;
}
```

lachlan-00:
looks good for me in light theme now
image

I like the dark theme being bright for all of them though

lachlan-00:
last one to fix is the disable icon doesn't exist. enable does annoyingly

lock button was out of whack a bit too but i think that's all the icons i could find.
image

Missing icons will show up as a question mark
Screenshot 2024-04-18 090708

hover and active focus all set now for dark to match light.
Screenshot 2024-04-18 092705

mitchray:
There is 'block' which I didn't see before actually, I was using 'hide_source' as an alternative

firefox_E1IZQXIgAk

firefox_87qOy15XAp

PR #3939, opened 2024-04-26
Prepare for separation of JS from PHP; header.inc.php done

mitchray:
The purpose of this is to lay the foundation for separating the Javascript/jQuery out of the PHP files, which will hopefully have the secondary benefit of making future PHPTAL migrations more straightforward.

There will be no changes to the JS logic apart from switching out the interspersed JS-from-PHP to be purely JS.

```public/templates/header.inc.php``` has been processed as a proof of concept; cut from 649 lines to 255.

To outline the changes:

- Introduce [vite](https://vitejs.dev/) to bundle multiple JS files (e.g. ajax.js, tools.js, base.js etc) into a single JS file.
  - Will first attempt to load from the vite server if developing (```npm run dev```), otherwise will load the compiled file (```npm run build```) based on a manifest (output is hashed so it can be cached)
- ```src/js``` will be for global JS which is to be bundled, with ```public/lib/javascript``` remaining for page specific JS files such as search.js 
- To preserve the 'legacy' nature of all the JS functions, they are [exported](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#browser_compatibility) and imported back into the ```window``` namespace so there are no changes needed for their use throughout the codebase and will behave as previous
- Every JS-from-PHP initializer will be extracted to ```public/templates/js_globals.php``` to consolidate all the global JS variables, grouped together and named according to their type e.g. AmpConfig, Preference, etc

### CODE REVIEW

**Comment by lachlan-00** [comment]
i'll check out the commit on my server soon, just about to start house transitions when the new net is connected

**Comment by lachlan-00** [comment]
the only difference i can see with no vite manifest is the  var jsAmpConfig header titles don't work. everything else seems okay without the manifest file.

Is that cause it's taken out of the php and doesn't load it cause it's missing?

not a big deal if you're following the process and doesn't look like it breaks anything else on the pr when you ignore it which is always good.

**Comment by mitchray** [comment]
Precisely, the manifest file needs to be present (from ```npm run build```) unless the dev server is running (```npm run dev```)

**Comment by lachlan-00** [comment]
one log error while updating preferences/browsing around.


```
2024-05-02T23:03:13+00:00 [ampache] (log.lib) -> [Runtime Error] file_get_contents(assets/main-8ZsXVU70.js): Failed to open stream: No such file or directory in file /var/www/music/vendor/idleberg/vite-manifest/src/Manifest.php(160)
```

haven't checked it yet but the path is 

public/dist/assets/main-8ZsXVU70.js so it must be missing the dist somewhere

**Comment by lachlan-00** [comment]
json from the build is missing the dist i think

**Comment by mitchray** [comment]
The paths from the manifest are relative to dist, and we output as ```src="<?php echo "{$web_path}/dist/{$entrypoint['url']}"```

I'll see if I can reproduce it

**Comment by mitchray** [comment]
Fixed in https://github.com/ampache/ampache/pull/3940

File: src/js/tools.js [hunk 1]
```diff
@@ -37,14 +37,14 @@ function showFilters(element) {
...
         $("#playlistdialog").dialog("close");
     }
     closeplaylist = 1;
 }
 
-function showPlaylistDialog(e, item_type, item_ids) {
+export function showPlaylistDialog(e, item_type, item_ids) {
     $("#playlistdialog").dialog("close");
 
     var parent = this;
```
lachlan-00, line 50: @mitchray tracked it back to here.
`TypeError: a is undefined`

could this be an issue where the dialog isn't actually there yet?

the div #playlistdialog doesn't exist and gets created by the function.

If i change this to the event (e), the dialog loads.

```
var parent = e;
```

is that wrong or incorrect somehow?

lachlan-00: set with bf69332923df75af25b38940f74e435e1a5b0e43 seems okay but just in case it's wrong

lachlan-00:
another one, the search values for option lists are in value from 0 now.
image

the js for this is not moved over yet but must be looking for something it can't see. it's getting the names okay so it's a bit odd that the value is wrong. 

I checked if i'd changed the specific data in the rule creation but it occurs with all select types now.

src/js/tools.js
mitchray: Ah yeah the meaning of ```this``` changes now and no longer refers to the ```window``` object, I'll have to look out for that and update as I go.

_Technically_ that change works as its just building up some properties for an object to instantiate the dialog, and ```e``` happens to be an object; fine for the time being but I'll touch it up when I make the next pass

src/js/tools.js
```diff
@@ -219,7 +219,7 @@ function handleShareAction(url) {
 var tag_choices;
 var label_choices;
 
-function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_row_prefix, argument_string) {
+export function showEditDialog(edit_type, edit_id, edit_form_id, edit_title, refresh_row_prefix, argument_string) {
```
lachlan-00, line 222: another one i didn't see was object edit.
this one doesn't send the event so it's not going to work.

When i add this or event in the function it's missing the dialog buttons
```
jQuery.Deferred exception: dialog_buttons is not defined showEditDialog
```

PR #3940, opened 2024-05-03
Don't need to calculate the file hash

mitchray:


lachlan-00:
don't need to entrypoint twice there i assume?

PR #3944, opened 2024-05-06
Lessen rfc 3514

mitchray:
Removes the superfluous use of the [evil bit](https://en.wikipedia.org/wiki/Evil_bit), keeping just one unobtrusive reference in the [header](https://github.com/ampache/ampache/blob/31088e92252c61eb4aaa8d8484c9c7fd86a2a870/public/templates/header.inc.php#L107). 

It was also being used for actual output so those have been renamed.

(sorry to whoever implemented this originally)

### DIFF
--- a/public/templates/error_page.inc.php
+++ b/public/templates/error_page.inc.php
@@ -43,8 +43,6 @@ $dir       = $dir ?? 'ltr'; ?>
 <link rel="stylesheet" href="templates/install.css" type="text/css" media="screen" />
 </head>
 <body>
-<!-- rfc3514 implementation -->
-    <div id="rfc3514" style="display: none;">0x0</div>
     <div class="navbar navbar-inverse navbar-fixed-top" role="navigation">
         <div class="container" style="height: 70px;">
             <a class="navbar-brand" href="#">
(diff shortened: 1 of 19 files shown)


lachlan-00:
Seems to break the webplayer popup with this one

mitchray:
Wow, its actually being used for critical operation. I'll revisit this.

Again, wow.

vollmerk:
> Wow, its actually being used for critical operation. I'll revisit this.
> 
> Again, wow.

Hahaha, AMAZING - I added that when "AJAX" was just starting to be a thing, and you had to edit the DOM manually. Sad to see it go, but glad that it survived this long in a way.

mitchray:
Its a testament to your efforts (and the web as a whole) that it still works!

New PR here https://github.com/ampache/ampache/pull/3945

lachlan-00:
Classic Ampache here. (love it)

ajax pages echo xoutput_from_array
image

which defaults to xml
image


(the xoutput request variable only seems to be for next/prev on browses)
image

And then loops back to xoutput_from_array as a callback
image

which seems to be used to avoid putting a header on the string
image

Which is then all echoed out.

at least that makes sense to why the arrays are there.

PR #3951, opened 2024-05-14
Explicity use 'window' as 'this' no longer refers to it

mitchray:
Addresses bugs from #3939


lachlan-00:
this window event parent... JS is a harsh mistress

PR #3957, opened 2024-05-24
Convert remaining MyISAM to InnoDB

### LINKED ISSUE #3776: MyISAM may be deprecated
Hi,

MyISAM look like it's deprecated : https://en.wikipedia.org/wiki/Comparison_of_MySQL_database_engines say it's no longer "Under active development"

Ampache 6.2.1 still have a few MyISAM table : 
```bash
# grep 'ENGINE=MyISAM' ./resources/sql/ampache.sql|wc -l
4
```

The list:
- [ ] album_disk
- [ ] album_map
- [ ] artist_map
- [ ] tmp_browse

Maybe we should migrate to something else?

BR,

fufroma:
Fix #3776

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -16,8 +16,9 @@ This release has made greater use of browses reducing reliance on customizing mi
 * Allow sorting searches by `type`, `limit` and `random` status
 * Ensure `catalog` is available on media browses
 * Add many missing database columns to browse sorts and filters on all browses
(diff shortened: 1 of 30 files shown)


lachlan-00:
I've put this in patch 7 with c3ab60047

I'll use the drop/recreate of tmp_browse so people can convert their DB if they want without failures

I don't want to force a change like that on people when 6 is mostly done so leave it optional

PR #3958, opened 2024-05-26
Rework of get_themes() to let PHP handle most of the work

fufroma:
Instead of using `opendir()`/`readir()` I used `glob()`, so we get only directory with a `theme.cfg.php` file.

If themes dir have some trash in it (like `.htaccess` or anything else), it won't show.

I think I could even get rid of the `ksort()`, but I let it.


lachlan-00:
I was going to change the other functions to send file path but name is fine now

PR #3963, opened 2024-05-30
Add a permalink in the search form

fufroma:
This patch add a permalink, so people can bookmark a search or share it with a link.


fufroma:
And yes, one can argue it's quite a dirty piece of code.

lachlan-00:
that;s a pretty good idea. cause i do searches manually often for sanity checks

lachlan-00:
composer cs:fix will sort you out here but i'm doing it now

lachlan-00:
made some additional changes in 9abf7423f872b019e6a4cac228d6ecfeed1b2023 to allow post and request data checks and will only show the link button when you have run an search action

Issue #4011, opened 2024-08-26
Test of the OpenSearch search (XML) fails for me


lachlan-00:
### Discussed in https://github.com/ampache/ampache/discussions/4010

<div type='discussions-op-text'>

<sup>Originally posted by **DJ-SteffEN** August 27, 2024</sup>
Hello,

I would like to know if the OpenSearch search (XML) still works, or how I can test it. 

![image](https://github.com/user-attachments/assets/af36878e-7c6e-465e-be94-46e91a09d56a)

Kind regards 
icqtoni</div>

lachlan-00:
I've created what i think is valid XML output but i still get an error adding as a search engine, not really sure where firefox will log or report these errors?

the file is public/templates/show_search_descriptor.inc.php
```
use Ampache\Config\AmpConfig;

$web_path = AmpConfig::get_web_path();
$char_set = AmpConfig::get('site_charset', 'UTF-8');

header(sprintf('Content-type: application/opensearchdescription+xml; charset=%s; filename=opensearch.xml', $char_set));

echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" .
    "<OpenSearchDescription  xmlns=\"http://a9.com/-/spec/opensearch/1.1/\">" .
    "<ShortName>" . scrub_out(T_('Ampache')) . "</ShortName>" .
    "<Description>" . scrub_out(AmpConfig::get('site_title')) . "</Description>" .
    "<InputEncoding>" . $char_set . "></InputEncoding>" .
    "<OutputEncoding>" . $char_set . "></OutputEncoding>" .
[...]

PR #4015, opened 2024-08-28
Partial fix of #4011 - Opensearch

fufroma:


lachlan-00:
definitely shouldn't be there

PR #4031, opened 2024-09-22
Fix content-length calculation for MP3 transcode when specific bitrate requested

ruipin:
When FLAC->MP3 transcoding was enabled, some clients (notably supersonic) would incorrectly calculate the duration of the song. Scrubbing also seems to break in these clients, and more importantly also the webplayer.

I looked into it and noticed that Ampache was reporting an incorrect Content-Length header for the media requests e.g.
```
https://ampache.example.com/play/index.php?ssid=foobar&type=song&oid=1234&uid=1&transcode_to=mp3&transcode_to=mp3&bitrate=320000&name=Foobar.mp3
```
And the song duration calculated by the client was exactly `Content-Length / (320*1024)` i.e. the client was relying on the content-length and assuming a fixed bitrate of 320kbps to calculate the song length.
Because the content length returned by Ampache was the original file size, this would return a broken song duration.

Looking at the code that generates this `Content-Length` header, I quickly noticed that the `$maxbitrate` variable is calculated based on the `bitrate` GET parameter to the request captured in `$troptions['bitrate']`:

```php
if (isset($troptions['bitrate'])) {
    $maxbitrate = $troptions['bitrate']; // <---- bits
} else {
    $maxbitrate = (empty($transcode_settings))
        ? $media->bitrate / 1024
        : Stream::get_max_bitrate($media, $transcode_settings);  // <---- kilobits
}
```

Note that the first branch of the `if` does not divide by 1024, but the second branch does. This means that whenever the request contains a `bitrate` parameter (in bits), we will use it as the `$maxbitrate`, but when it doesn't, we will use `$media->bitrate / 1024` (i.e. in kilobits). I've added comments to help illustrate.

This unit mismatch means that the logic to calculate content length breaks. Because the maxbitrate in bits will almost certainly be larger than the media bitrate in kilobits (due to the 1024 factor), it will instead just use the `$media->bitrate` value (second branch of the ternary) when calculating `$stream_rate`:

```php
} elseif ($transcode_to == 'mp3') {
    // mp3 seems to be the only codec that calculates properly
    $stream_rate = ($maxbitrate < floor($media->bitrate / 1024))
        ? $maxbitrate
        : (int)floor($media->bitrate / 1024); // <--- will use the second branch of the ternary
    $stream_size = (int)(($media->time * $stream_rate * 1024) / 8);
} else {
```

Changing this, such that the `$maxbitrate` is always in kilobits, seems to result in a correct content-length estimate.

With this change, supersonic displays the correct song duration. It also seemed to fix scrubbing in both supersonic *and* the webplayer.


ruipin:
I don't think the build failures have anything to do with my change as I barely touched any code.

lachlan-00:
just some failed tests in the master branch.

I have that set in develop, not sure why i didn't backport to 6 but lets do that

ruipin:
Thanks for merging.

After more time using this PR, I have noticed that this does not fix the situation where the media bitrate is lower than the MP3 bitrate.

For instance, my server is configured to transcode everything that is not MP3 into 320kbps MP3. I have a 315kbps m4a file, which even after this fix triggers the second branch of the ternary and miscalculates the content-length as if the MP3 bitrate were 315kbps, rather than 320kbps.

Upon further analysis think the `$stream_rate` ternary makes no sense, and we should just do `$stream_size = (int)(($mediaTime * $maxbitrate) / 8);`

PR #4035, opened 2024-09-23
Further fixes to Content-length calculation

ruipin:
The previous PR#4031 I submitted did not correctly handle the case where the pre-transcode bitrate is lower than the transcode bitrate.

For instance, my server is configured to transcode everything that is not MP3 into 320kbps MP3. I have a 315kbps m4a file, which even after this fix triggers the second branch of the ternary and miscalculates the content-length as if the MP3 bitrate were 315kbps, rather than 320kbps.

Upon further analysis think the $stream_rate ternary makes no sense, and we should just do `$stream_size = (int)(($mediaTime * $maxbitrate * 1024) / 8);`.

This is because if we know we are doing a transcode, then the original media bitrate does not really matter - all that matters is the target bitrate of the transcode.

This PR contains this change, as well as a slight cleanup to make the code less confusing. In particular, it uses `$stream_rate` instead of `$maxbitrate` (since that variable is used elsewhere in the same function with a different meaning), and correctly emulates the logic in `Stream::start_transcode` used to decide the transcode bitrate, to ensure the bitrate used to calculate the Content-Length is the same as sent to the transcode application.


ruipin:
I have noticed that when there is not a transcode because the file is already MP3, but the source file has a large amount of non-music data (e.g. image tags), supersonic displays a longer duration than expected. This seems to be because it assumes the content-length only covers the music stream, rather than the entire file.

This seems to also affect the webplayer too - if I build a MP3 file with 50% of the size being an MP3 stream, this seems to also break scrubbing by essentially adding a 50% "deadzone" to the scrubbing bar as if the song was double the duration.

I'm not sure what would be the best way to handle this. We could replace the content-length calculation by our own when `$is_download` is false, but for this to work properly with all players I believe we'd also need to be able to skip non-music data when handling a HTTP range request. As such, this PR doesn't touch that, and just leave the non-transcode path as-is.

lachlan-00:
Thanks for looking so deeply. it's very confusing here so it's always good to have people look over these areas. This is also one of the most annoying areas of the code to get right.

In my head i assume that the initial size of the file should give us an idea of what junk (non-music) data is on there for a stream

so it would be something like:

* media->size - (media->bitrate / media->time) = Left over data

so then ideally a transcode would usually re add the junk data outside the music stream?

Left over data + expected stream_size = output total size

Does that make sense? I hope i've explained that well i think that might help cover the differences if we put 50% of the left over data on there

ruipin:
I think you misunderstood the second issue - it only affects non-transcodes.

When we are doing a transcode, we send the client only the raw audio stream, so it works well after this PR.

When we are not doing a transcode, we just pass the client the original file. However, in the case where the client doesn't want to download the file, it seems both the Web player and supersonic assume that the only thing being sent over is the audio stream. Since we send them the full file, the scrubbing and song duration calculation is wrong.

To fix this, we would need to know the size of the audio stream portion of the file, and also the offset it starts at. The former is easy to calculate, but the latter would be necessary to properly handle HTTP Range requests.

That issue occurs even without this PR, but since the audio stream is usually most of the file, it is usually not noticeable.

lachlan-00:
thank you for explaining!

PR #4065, opened 2024-11-04
Album name added to songs search

Simounet:
Hi there,
On the search bar, when you have more than one version of a same song (studio album, lives…) I thought it could be handy to display the album name.
I made a CSS adjustment to improve the styling on long line result.

I would also make a PR to reorder CSS loading. Seems weird to me that components CSS loads after ours.

Before:
```css
<link rel="stylesheet" href="https://ampache.simou.net/templates/base.css" type="text/css" media="screen">
<link rel="stylesheet" href="/templates/jquery-ui.custom.css" type="text/css" media="screen">
```

After:
```css
<link rel="stylesheet" href="/templates/jquery-ui.custom.css" type="text/css" media="screen">
<link rel="stylesheet" href="https://ampache.simou.net/templates/base.css" type="text/css" media="screen">
```

It would be easier to override default styling instead of duplicating a class to have an higher specificity, just like here with the `.ui-menu-item.ui-menu-item` class.

### DIFF
--- a/public/templates/base.css
+++ b/public/templates/base.css
@@ -225,10 +225,14 @@ a.tag_size4 { text-decoration: none; }
     float: left;
 }
 
+.ui-menu .ui-menu-item.ui-menu-item a {
+    display: flex;
+    align-items: center;
+}
+
 .searchitemtxt {
     display: inline-block;
     vertical-align: middle;
-    height: 32px;
     color: #222;
 }
 
--- a/public/themes/reborn/templates/default.css
+++ b/public/themes/reborn/templates/default.css
@@ -408,6 +408,7 @@ div#autoupdate > span {
 
 #ui-id-1 {
     min-width: 20%;
+    max-width: 630px;
 }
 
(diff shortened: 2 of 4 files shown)


lachlan-00:
one thing that's odd is there must be some JS to arrange the search results, the element css is regenerated each time i open the results to match the screen size

image

I'm happy to change anyhting css related i think a lot of this is just old / holdover work done which hasn't been touched for a long time

lachlan-00:
getting rid of the height helps a lot. but it's weird, if i open the search bar, visually it looks good and then after each click/reload it reverts to the old style

image

then typing a third letter to regenerate search
image

Simounet:
I fixed it with the less ugly solution in my opinion. I tried both with CSS and JS. It's all about the autocomplete jQuery pluggin. Let me know if everything is ok on your side.

lachlan-00:
cool thank you, i had a look and i got lost. i'll try it now

PR #4067, opened 2024-11-06
Rating icons replaced by material SVG icon

Simounet:
I tried to look at every section containing the rating system but maybe I missed something. I removed the reset icon, you need to click on the current rated icon to remove your note.
I added to translation, not sure how you want me to handle that:
- T_('Reset rating')
- T_('Rate ' . $count)
- T_('Favorite')
- T_('Unfavorite')


lachlan-00:
I'll check it out when I'm free again but I would say clear or remove

The text is stars and flag mostly so

%d star/s
Remove stars
Flag
Remove flag

Simounet:
The "flag" wording looks weird to me because the app use "favorite".

lachlan-00:
yeah that's fine i think i'm thinking more from the backend.

It looks good but i don't think we change the behaviour of clicking the star again to reset the value, i've added in the 0 stars button again. clicking the same icon will function as it did before and reset the star and apply plugins.

image

lachlan-00:
what a banger pull though, brings it all together

PR #4088, opened 2024-12-07
Remove broken rss recently skipped

cquike:
This  removes the currently wrong RSS link to recently skipped that appears in a users profile.
    
This has actually been wrong since a while.  The RSS link to the recently skipped songs is actually pointing to the recently played songs.

I guess this is not a very useful feature, but can be restored (and properly implemented) if someone thinks it is useful.


lachlan-00:
rss for skipped wouldn't be missed i would say

PR #4106, opened 2025-01-18
Display artist in playlist

fufroma:
Add an "artist" column to the playlist page.


lachlan-00:
if you enabled extended playlist links you'll get this.

`action=admin&tab=playlist`

name in english is "Show extended links for playlist media"

lachlan-00:
This is the section in code. get_f_parent_link() is avaialble on all media types as well
image

I'll see if it looks better it would at least allow sorting items by artist as well

lachlan-00:
added preference for it, this will allow people to show/hide based on preferences.

PR #4112, opened 2025-01-25
OpenSubsonic API non compliant with specification

### LINKED ISSUE #4110: OpenSubsonic API non compliant with specification
### Describe the bug

There is a capitalization issue within the OpenSubsonic API getOpenSubsonicExtensions,
discovered this while i tried to connect Ampache to MusicAssistanc using subsonic
https://github.com/music-assistant/support/issues/3496
here is some more details ^

### To reproduce

Use py-opensonic call  conn.getOpenSubsonicExtensions() and look at the output

### Expected behavior
{'status': 'ok', 'version': '1.16.1', 'type': 'ampache', 'serverVersion': '6.6.6', 'openSubsonic': True, 'opensubsonicextension': [{'name': 'formPost', 'versions': [1]}, {'name': 'transcodeOffset', 'versions': [1]}]}
in here 'opensubsonicextension' should be 'openSubsonicExtensions' According to the spec: https://opensubsonic.netlify.app/docs/endpoints/getopensubsonicextensions/

## Environment

* Ampache version: 7.2.0
* Server operating system: Linux

### Client type
* SubSonic

fufroma:
Fix #4110

### DIFF
--- a/src/Module/Api/Subsonic_Xml_Data.php
+++ b/src/Module/Api/Subsonic_Xml_Data.php
@@ -256,7 +256,7 @@ class Subsonic_Xml_Data
      */
     public static function addOpenSubsonicExtensions($xml, $name, $versions): void
     {
-        $xextension = self::addChildToResultXml($xml, 'opensubsonicextension');
+        $xextension = self::addChildToResultXml($xml, 'openSubsonicExtensions');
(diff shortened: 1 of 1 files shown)


lachlan-00:
i'm so bad with case. apologies

PR #4113, opened 2025-01-26
Use id attribute selector as workaround for multiple identical IDs

### LINKED ISSUE #3995: Setting number of stars for a song does not update the webplayer GUI
## Description

In the webplayer UI I usually set the number of stars for a song. The DB is updated but in the UI there is no update, so to the user it looks as if nothing has happened. Interestingly enough, if the song is also currently shown in the main web canvas (as part of a playlist for instance) then the stars do get updated correctly.

### To reproduce

1. In the web UI click select a playlist which has a song that is not yet rated.
2. Play the given song and while playing set he number of stars in the player UI at the bottom.
3. The number of stars is not set in the player UI. However the song stars as listed in the playlist is properly updated

## Environment

* Ampache version: 6.5.0
* Server operating system: Debian 12
* Client operating system: Debian 12

### Client type

* Web

mitchray:
Fixes #3995

### DIFF
--- a/src/Module/Api/Ajax.php
+++ b/src/Module/Api/Ajax.php
@@ -67,7 +67,8 @@ class Ajax
         if (in_array($source, $non_quoted)) {
             $source_txt = $source;
         } else {
-            $source_txt = "'#$source'";
+            // use id attribute selector as workaround for multiple identical IDs (e.g. rating)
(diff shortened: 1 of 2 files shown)


lachlan-00:
this is huge. i'll look when so=ber

PR #4153, opened 2025-04-07
Add plugin to share with Bluesky & Mastodon

fufroma:
Bluesky is pretty straightforward.

Mastodon was tricky: I need to ask for the Mastodon server. But it would not be as generic as the other share plugin. I found a trick: selfhosting a [toopick](https://github.com/Juerd/tootpick) html page. Work perfectly, no external dependency, and the licence is OK with that.


lachlan-00:
i'll check if fosstodon works, that's the one the main account uses

Issue #4162, opened 2025-04-20
[Regression] Catalogue update scans the whole collection, ignoring catalog_verify_by_time option


cquike:
## Description

I have realized that it now takes a long time to update the catalogue, despite the option catalog_verify_by_time= true in the ampache.cfg.php file.
This is actually a regression, since it was working fine before.

### To reproduce

1. Launch the update cli command:
`cli run:updateCatalog -e`
2. Check in the logs that "Reading tags from ..." debug message appears for every single file in the collection.

### Expected behavior
Only the files that have been modified since the last scan are actually read and updated.
 
## Environment

* Ampache version: 7.3.0
* Web server + version: nginx 
* Server operating system: Debian 12 
* Client operating system: Debian 12
## Settings
...
catalog_verify_by_time = "true"
...

PR #4164, opened 2025-04-20
Fix skipping files based on modification time when updating catalogues

cquike:
This PR tries to fix the problem described in #4162 .
There are actually a couple of things:

- I think the problem was introduced in commits d2038e65605e81d025271cac5be1553c1cf710d1 and b8dc35a8a8dfb57985a4101b99a827b31d77d158. For some reason the logic has been changed to check the update time of the database. But that simply does not work if the files are updated externally and the catalog_verify_by_time option is set. I kind of reverted those two commits in this patch. What it is checked now is the last time the song was updated according to the database and the real modification time of the file. 
- Also, the check with the catalogue update does not make much sense to me. What is important is the last update in the file. So the comparison between the last time the catalogue was updated and the filename modification time has been removed from the original patch.
- Related to the previous point, there is no filtering of the songs in the database by the update catalogue time. I actually saw a bug with that. If a file(s) get's updated in a run of `run:updateCatalog -e` then the song update time in the database is newer than the catalogue update (because the timestamp for this one is when the updateCatalog command starts.  If then, the file is updated _again_ on disk, the next time the `run:updateCatalog -e` runs it is ignored because the sql would ignore files which have a modification time in the database larger than the update catalogue timestamp.

I am not sure whether these changes would also be fine for other setups that are not mine, so probably check carefully. Regarding that, I think it would be very useful to have some unit tests that exercise the different options and check for the expectations. I think it is quite easy to break this code without having some safety net applied to each commit. Maybe there is something and I did not see it, but a naive `grep verify_catalog_proc -r tests`  did not show anything. I am afraid I cannot be of help here because I know very little about testing php code.


lachlan-00:
The idea with last_update is to reduce the size of the files being checked.

With lots of files it can take weeks to verify a catalog so the last_update will skip any files that were updated AFTER the last update.

e.g. if i have 1million files
* I update 100k files and then stop the process
* if you take out last_update they will be verified again next time you start meaning you are likely to never finish
* With last_update you only have 900k songs to go and can do this in smaller chunks (with the limit parameter)

The idea is that large catalogs can reduce their verify lists until they are completed and then will start again.

It's not a bug that songs have newer update times than last_update, the last_update value is only updated when you run out of songs updated before that time.

But i did take out song modification time checks and we can look at putting that back, the verify process is a lot faster now as well so it shouldn't affect anything to add the check back

lachlan-00:
release is done now i can do this. i would say that the way forward is to use time OR last_update. i'll make the changes on the PR to allow either option

cquike:
I see your point on the update of large collections. I have modified the PR to accommodate for that. The way I solved it is that even if the file tags are skipped because of the modification time, still, the database records that the files has been updated.
That seems to solve the problem. I have tested that and also the update of the collection in batches using the -l option.
Shall I push the changes to the PR? Or do you prefer to see the changes in a separate branch?
Thanks!

lachlan-00:
just push it all on in here is fine. we're starting again on 7,4,3 now so not a big deal for new changes

lachlan-00:
whatever we get done here will all squash in and it's no biggie for me to test

cquike:
I have updated now to 7.4.2 and the update is failing now with something that seems unrelated:
TypeError Ampache\Repository\Model\Catalog::check_int(): Argument #2 ($max) must be of type int, float given, called in /srv/ampache_musica/src/Module/Util/VaInfo.php on line 656
(thrown in /srv/ampache_musica/src/Repository/Model/Catalog.php:3738)

I just wanted to try my patch on top of 7.4.2.

So I will try to figure out what is going on with this before I rebase this patch.

cquike:
Ok, I think I have solved the last problem and created a separate pull request: #4169 .
Then I have encountered yet one more problem with 7.4.2, which I think is the same as https://github.com/ampache/ampache/issues/4168
I see that https://github.com/ampache/ampache/commit/386730d0915774239920b6dd13c1fe5916f23203 might be a solution, will try out...

lachlan-00:
i'll have some time for this tomorrow so i think with your update i can get it done. for #4164 i think that was done for 32bit int weirdo's like the raspberry pi but it seems to be the only place that actually uses this function as well.

cquike:
So, I think I have now something that works. I have rebased it on top of the last changes here.

I have tested the following:
- If -l option is given with limits, everytime the file chunk is different and eventually the whole thing converges, the last time the command is run 0 files will be updated and the catalogue update time is updated.
- If no options, then the whole catalogue is scanned.
- In both cases, the file modification time is a shortcut to see whether the tags need to be read or not.

What I did not test; 
- The verify_by_album option.
- Catalogues with  podcast.

lachlan-00:
I think this is the way (combined with some develop work)

* catalog_verify_by_time is for file mod time only
* last_update is still kept and is used at all times to reduce catalog load (users with small catalogs will never notice this)
* All table checks except album verify will check mod time (might have to do a check for files inside the album)
* Null update_time has been removed as a possibility from songs, videos
* added update_time to podcast_episodes

Now that albums are the special case i might check the all file mod times in the album and then update or ignore when one file matches.

But i think this is done. i'm going to squash it all and commit

cquike:
Thanks for that last summary.
I think there commit https://github.com/ampache/ampache/pull/4164/commits/ca2c3a862bfd0ed29fae23cb4475fc660ec6e849 needs to be reverted. That causes the bug I mention that the catalogue update bounces every two updates and not all files are considered.

cquike:
Also,
the last commit in devel has the following:
```
+            if ($verify_by_time && $tableName === 'song') {
+                $file_time = filemtime($row['file']);
+                if ($file_time === false) {
+                    debug_event('local.catalog', 'Unable to get file modification time for ' . $row['file'], 3);
+                    continue;
+                }
+                // check the modification time on the file to see if it's worth checking the tags.
+                if ((int)($row['min_update_time'] ?? 0) > $file_time) {
+                    //debug_event('local.catalog', 'verify_by_time: skipping ' . $row['file'], 5);
+                    Song::update_utime($row['id']);
+                    continue;
+                }
+            }
+
+            if ($verify_by_time && $tableName !== 'album') {
+                $file_time = filemtime($row['file']);
+                if ($file_time === false) {
+                    debug_event('local.catalog', 'Unable to get file modification time for ' . $row['file'], 3);
+                    continue;
+                }
+                // check the modification time on the file to see if it's worth checking the tags.
+                if ((int)($row['min_update_time'] ?? 0) > $file_time) {
+                    //debug_event('local.catalog', 'verify_by_time: skipping ' . $row['file'], 5);
+                    continue;
+                }
+            }

```
So, it duplicates the logic for song. Is this intentional or a side effect of the squashing?

lachlan-00:
must be from merging it in from develop. didn't see that in the merge changes

date needs to be there so are you saying no files are scanned every second update? that should revert to total if 0

cquike:
You can try to reproduce the problem with the update_time of the catalogue as follows for a catalogue of 20000 songs:

1. Call run:updateCatalog -e -l 5000 
2. Call run:updateCatalog -e. This will consider only 15000 files
3. Call run:updateCatalog -e. This will consider only 5000 files
4. Call run:updateCatalog -e. This will consider only 15000 files again
5. Call run:updateCatalog -e. This will consider only 5000 files again
6. ....

The logic is simple: In the first call the update time of the 5000 songs is now larger than the catalogue update time and in 2. only the other 15000 are considered. But now, the update time of the catalogue is *less* than the update time of all those 15000 songs because the timestamp is from _before_ all those 15000 are updated. And the same logic repeats itself everytime, creating a bouncing behavior when  updating the whole catalogue.
Setting the catalogue update time to a time _after_ the files have been updated solves the problem (and I hope it does not introduce other problems).

lachlan-00:
just done some testing and i think it's okay after 172382605f1be2e2bb797c5b873a43b4462407b6

set a limit and it will go to 0 results and then redo the whole catalog

Verify and cancel the process, number gets smaller
Finish verify and restart and the total will be pulled

PR #4169, opened 2025-04-26
Fix input to check_int. Use PHP_INT_MAX rather than hardcoded 4294967294

cquike:
After the upgrade to 7.4.2 I get the following exception when upgrading the catalog:
```
sudo -u www-data php -f /var/www/ampache/bin/cli run:updateCatalog -a
Leyendo Catálogo: "My Catalogue"

Empezar a añadir nuevos medios
Reading flac inside /music
TypeError Ampache\Repository\Model\Catalog::check_int(): Argument #2 ($max) must be of type int, float given, called in /var/www/ampache/src/Module/Util/VaInfo.php on line 660
(thrown in /var/www/ampache/src/Repository/Model/Catalog.php:3738)

```
It looks like after introducing type declarations the inputs of check_int in Catalog.php need to be integers. There are a few calls to that function that use 4294967294 as a hard-coded upper limit, which is interpreted as a float. This pull request replaces them by PHP_INT_MAX, which also has the advantage of being platform independent.


lachlan-00:
i think this was done for raspberry pi systems with a 32bit int.

are you okay with the last patch i made casting to float? this function i think is just for videos to truncate large values

9f4258c38d32f7f6e7402aad71465b9388cfa1d2
```
public static function check_int(int|float $my_int, int|float $max, int $min): int
```

cquike:
Well, funny enough I found the issue in a raspberry pi 32 bit :)
For me, in that system, the PHP_INT_MAX solution works fine and seems to me like the most portable one which still takes the benefit from the type declaration, since the check_int function seems to really be thought for integers.
But I am fine if you prefer the float solution.

PR #4174, opened 2025-05-04
Scrutinizer build fix

fufroma:
Scrutinizer build is fix, I saw that you and @usox worked on it in 2024

~~But, even if it build, it doesn't find any issue. I think some tuning is still necessary.~~ Fixed


fufroma:
Scrutinizer fixed by launching "php-scrutinizer-run"

Issue #3966, opened 2024-05-31
[Feature Request] Better signup captcha


fufroma:
Our public instance of Ampache is open to signup, but we have bot creating account everyday.

Is there a way to enforce a better captcha?

Ampache 6.4.0

lachlan-00:
is it reading the image or the math captcha?

We do need to update it but it's been a lot beyond my knowledge.

you can play around with the constants in `src/Module/Util/Captcha/easy_captcha.php` as well.

lachlan-00:
there is a log you can enable which goes to /tmp/captcha/captcha.log by default

The id is also an md5 string with a salt, doesn't seem very random
![image](https://github.com/ampache/ampache/assets/1305249/243eecac-5ddc-4b88-9e34-8b321f5d0026)

lachlan-00:
probably try changing the CAPTCHA_PARAM_ID and CAPTCHA_PARAM_INPUT as well

and you can disable image if it's using OCR  by setting $this->image = false; in the generate() function

lachlan-00:
if it's a bot i would also look at updating the config with random values daily/weekly to try and screw with it. it's probably reading things as they haven't changed in forever

lachlan-00:
this was the last time we did something with it but just to fix up the image https://github.com/ampache/ampache/pull/3740

lachlan-00:
and another one the alt text makes the math probably readable / solveable
![image](https://github.com/ampache/ampache/assets/1305249/d5be69b8-547e-4ebd-bd5f-da69d22c2796)


i haven't looked at alternatives yet but there would have to be new libraries for this that aren't google/etc

fufroma:
> We do need to update it but it's been a lot beyond my knowledge.

Same here.

We are using the img captcha, I'll try the few config you suggest but I think a better captcha will still be a good improvement

fufroma:
https://github.com/altcha-org/altcha may be a solution

fufroma:
Do you have captcha somewhere in the roadmap? Bot are annoying :) 

Worst case i'll look to add a proxy to ampache

fufroma:
I now have dozens of bots to trash daily. Is captcha in your roadmap or should I look into other options?

mitchray:
I'd look into other options if you need an urgent resolution

fufroma:
https://github.com/ampache/ampache/pull/4367#issuecomment-4600571064

lachlan-00:
[...]

PR #4182, opened 2025-05-21
Check some spambot feature in user creation

fufroma:
As mentioned in https://github.com/ampache/ampache/issues/3966 , actual captcha does not protect from actual spambot

I added those lines a few weeks ago and most of spam user creation are now blocked.

I share it, but I would understand you don't want that in the main branch :)


lachlan-00:
I'll play with it but i think either of those is fine to have as options. might make it a config regex so people can allow/customise them

lachlan-00:
regex so then you can allow different registration types, no warning or anything though but should allow different things like disallowing spaces/other custom regex

PR #4196, opened 2025-06-02
Missing user in images enum class

fufroma:
User is allowed in db but not in class.

Caught it because of this error in logs:
`(src/Module/Application/Image/ShowAction.php:94) -> "user" is not a valid backing value for enum Ampache\Repository\Model\LibraryItemEnum`

It fix those links : 

- https://FQDN/image.php?object_type=user&object_id=28496&thumb=4

### DIFF
--- a/src/Repository/Model/LibraryItemEnum.php
+++ b/src/Repository/Model/LibraryItemEnum.php
@@ -43,4 +43,5 @@ enum LibraryItemEnum: string
     case TAG_HIDDEN      = 'tag_hidden';
     case TAG             = 'tag';
     case VIDEO           = 'video';
+    case USER            = 'user';
 }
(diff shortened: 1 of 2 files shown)


lachlan-00:
i've looked in the code and these shouldn't be there.

You should add a rewrite in there to redirect the bad calls. i've added on in https://github.com/ampache/ampache/commit/604a80bcc41cf56324810097a82f3bec24baec1f#diff-4755a08008be4b6e37fa3ef3b468c6b00a773dd371eb5bc81f8a04a35c2a6b0b

lachlan-00:
i think it's also stream_beautiful_url, which may cast it as user

in 90ae7a3bc8e7c772aa52c0893c6cff4436e63c14 that section is skipped entirely for user urls

fufroma:
> where are your links coming from?

I have no idea, when I look at the bug I didn't find any way to reproduce this url. I just have it in the log. As the `image` table got an enum with `user`, I though Art should be able to display user.

I reverted my patch and upgrade to 7.5.2, feel free to drop this pr if you think it's irrelevant :)

Issue #4199, opened 2025-06-04
Ampache 7.5.2 - Can't insert PNG


fufroma:
Ampache 7.5.2

Can't add this art to an album : 
[cash.png](https://forum.musique-libre.org/uploads/editor/5m/kisceqyl0qn6.png)

Error is : 
`Art file failed to insert, check the dimensions are correct.`

I just did:
`convert cash.png cash.jpg`

cash.jpg worked fined

I think new Art class doesn't handle png very well.

fufroma:
hum it may not have been the dimension but the file size, I need to take time to do some tests

PR #4203, opened 2025-06-09
Add error msg for Art upload

fufroma:
Help to find error for Art upload
Cf https://github.com/ampache/ampache/issues/4199


lachlan-00:
would you be able to update the french transifex translations asap if i did this one?

lachlan-00:
I might tweak this one a bit more to return different errors instead of relying on a straight bool return

lachlan-00:
now you will get an explicit error message based on the return. this should make things a bit easier to understand and find the error

PR #4206, opened 2025-06-14
Put back Etag to allow client-side cache

fufroma:


mitchray:
Ideally this needs to have some kind of 'last updated' info, otherwise it will be cached indefinitely

But yeah the etag is needed otherwise every image is going to be sent every time

fufroma:
I though the art id change if you upload new art.

Anyway, cache is for 7 days

https://github.com/ampache/ampache/blob/041335407bb9ee383dfc20b652baa363e7ca41c4/src/Module/Application/Image/AbstractShowAction.php#L200

PR #4208, opened 2025-06-15
Use getId() instead of $this->id

fufroma:
Got error like 
`PHP Fatal error:  Uncaught Error: Typed property Ampache\\Repository\\Model\\Query::$id must not be accessed before initialization in /opt/play.dogmazic.net/src/Repository/Model/Browse.php:521`

id is not always set, and getId() already got a workaround

So I use getId() instead of $this->id , no more errors

### DIFF
--- a/src/Module/Api/OpenSubsonic_Json_Data.php
+++ b/src/Module/Api/OpenSubsonic_Json_Data.php
@@ -852,7 +852,7 @@ class OpenSubsonic_Json_Data
         }
 
         if ($songs) {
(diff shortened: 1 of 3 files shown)


lachlan-00:
this could be on a deleted browse or a public user with no session reloading a link.

image

I would probably just set a 0 id on the class, that would cover all situations
image

fufroma:
I tried to set a default value, seems to work fine, let's do this!

lachlan-00:
set with 9b78de462a173370ed8afb6860eab30a351be8da

PR #4213, opened 2025-06-21
Direct call to image.php

fufroma:
If you call image.php without parameters: 
```
2025-06-21T19:55:30+00:00 [-1] (src/Repository/Model/Art.php:267) -> Typed property Ampache\Repository\Model\Art::$object_type must not be accessed before initialization
```

So, I added a dumb fix to handle dumb bot and not having error in my log.

I'm pretty sure you'll want to handle this another way, feel free to trash this PR

### DIFF
--- a/public/image.php
+++ b/public/image.php
@@ -34,6 +34,13 @@ use Psr\Container\ContainerInterface;
 define('NO_SESSION', '1');
 define('OUTDATED_DATABASE_OK', 1);
 
+if (empty($_REQUEST)) {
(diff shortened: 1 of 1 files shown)


lachlan-00:
you could probably just ignore it at the abstract action.

f2dd28181ab0bcf183e7e20d60414c7bca714369

we can put error codes in if they're not giving up though. The error is actually because the ID isn't found in the art construct.

fufroma:
I have no idea how to do it your way :s

PR #4225, opened 2025-07-06
remove loading of non existent bootstreap theme css

olaulau:
causes 404
/install.php, /index.php ...


olaulau:
so this PR is valid, bootstrap has breaking changes that hasn't been taken into consideration in install process.
https://github.com/ampache/ampache/pull/4228

PR #4263, opened 2025-09-07
Fix "Unable to locate package software-properties-common" when building docker images

moormaster:
Fixes an error message when running `docker compose up`:

```
1.810 E: Unable to locate package software-properties-common
```

Apparently this package is not part of Debian Trixie (the current Debian Stable) anymore:
https://packages.debian.org/search?searchon=sourcenames&keywords=software-properties


lachlan-00:
I found extrepo which seems like a cleaner way of doing it.

I haven't build anything yet for it so it's only in master https://github.com/ampache/ampache-docker/commit/505eab0586e551f507dd5cb23c552c0794488fd8

PR #4285, opened 2025-09-26
Fix

fufroma:
Some fix for error I got in ampache.log


fufroma:
> Did it tell you where the error was in hte logs?

I pasted the error in the commit msg 

`(src/Repository/Model/Art.php:267) -> Typed property Ampache\Repository\Model\Art::$object_type must not be accessed before initialization`

PR #4298, opened 2025-10-26
Add a new option to the run:updateCatalog command to rename a file

cquike:
This PR allows to rename a file in a local catalog. The old nae needs to be present in the database and the new file needs to exist in the local filesystem.
To use it:
`run:updateCatalogFile -r /tmp/new-file.flac some-catalog /tmp/old-file.flac`

My rationale for this feature is the regular use of picard to keep up to date my collection. I configure picard to rename the files following the track and album names, which might potentially change. In that case, while running my periodic database update, ampache detects the old filename as deleted and the new filename as a new song, losing in the process the properties of the old song like rating or play history.

With this PR I can safely rename the files and update the database for that file before running the periodic database update without losing the history of the song.


lachlan-00:
that's a good one, i'll extend this to allow organising the file as well according to catalog patterns

PR #4328, opened 2026-02-04
Fix subsonic return code when scrobble contains invalid id

cquike:
This PR fixes the error code returned by the Subsonic API when the given id to scrobble is an invalid object.

In the current implementation an invalid object id (999999999 in this case) still gives an OK response:

```
$ curl "https://example.org/rest/scrobble.view?id=999999999&time=1688760805000&submission=True&c=curl&v=1.11.0&f=json&u=user&p=enc%9999999999999999999999999999999999999999999999"
{
    "subsonic-response": {
        "status": "ok",
        "version": "1.16.1",
        "type": "ampache",
        "serverVersion": "7.8.0",
        "openSubsonic": true
    }
}%        
```

This follows from the discussion in OpenSubsonic: https://github.com/opensubsonic/open-subsonic-api/discussions/201

### DIFF
--- a/src/Module/Api/OpenSubsonic_Api.php
+++ b/src/Module/Api/OpenSubsonic_Api.php
@@ -3850,15 +3850,21 @@ class OpenSubsonic_Api
         }
 
         // check that all the objects are valid
-        $all_media = [];
+        $valid_media = [];
         foreach ($sub_ids as $sub_id) {
-            $media     = self::getAmpacheObject((string)$sub_id);
-            if (!$media instanceof Media || !isset($media->time) || !isset($media->id)) {
-                self::_errorOutput($input, self::SSERROR_DATA_NOTFOUND, __FUNCTION__);
-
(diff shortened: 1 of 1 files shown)


cquike:
With the code in this PR, an error is given:
```
{
    "subsonic-response": {
        "status": "failed",
        "version": "1.16.1",
        "type": "ampache",
        "serverVersion": "7.8.0",
        "openSubsonic": true,
        "error": {
            "code": 70,
            "message": "The requested data was not found.",
            "helpUrl": "https://ampache.org/api/subsonic"
        }
    }

```

lachlan-00:
there isn't a clear discussion on whether an error should be returned on anything but i'm sure there are unspoken defaults on there.

I'll work on these pulls as a priority next week but my time is tight lately

lachlan-00:
Looks like it's an all or nothing error in the discussion so i'll go with that.

PR #4368, opened 2026-06-03
Fixes js-cookie copying since #4358

rsertelon:
I'm building ampache using ` npm install` in the process, the `js.cookie.js` file couldn't be loaded since 7.9.5 because the file changed places. Which was breaking the web player.

This MR simply fixes the path of the file in newer version of js-cookie.


lachlan-00:
thanks @rsertelon i'll add in a verification step to my zip builds and docker as well

PR #4377, opened 2026-06-10
Restore functionality to broadcast playback; closes #3441

### LINKED ISSUE #3441: [Feature Request] Re-add the broadcast feature
## Describe what you don't like about Ampache, or what it is missing

In an earlier version of Ampache, there used to be a "broadcast" feature that was similar to "Remote group session" on Spotify. There is a flag to add to page to the sidebar, but the feature is completely broken. Adding this feature would allow Ampache to stand out from the rest of the media players online.

## Describe how you would like the feature to work**

Ideally, it would support creating broadcast sessions that use websockets to synchronize songs between the host and all of the clients. 

I was able to get it somewhat working again a couple of months ago by updating some webplayer and request handling code but it was very buggy. For someone more experienced with the codebase and PHP, the scope of this issue shouldn't be very big.

stutsmaniii:
After trying to utilize the broadcast feature myself and finding #3441, I decided to take a look into the problem. It turns out Broadcast does not implement the Media interface, which causes the call to media_object_to_url to be skipped and the playback URL for any Broadcast object to end up empty. 

I got the broadcast feature working for myself by adding an alternative version of the media_object_to_url function which accepts the Broadcast object. I suspect Broadcast should eventually implement the Media interface, but this allows a Broadcast object to be played regardless.

### DIFF
--- a/src/Module/Playback/Stream_Playlist.php
+++ b/src/Module/Playback/Stream_Playlist.php
@@ -35,6 +35,7 @@ use Ampache\Module\System\Session;
 use Ampache\Module\Util\ObjectTypeToClassNameMapper;
 use Ampache\Module\Util\Ui;
 use Ampache\Repository\Model\Art;
+use Ampache\Repository\Model\Broadcast;
 use Ampache\Repository\Model\Democratic;
 use Ampache\Repository\Model\library_item;
 use Ampache\Repository\Model\LibraryItemEnum;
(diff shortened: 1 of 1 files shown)


lachlan-00:
AWESOME! This is something i never fully understood so it's great to see you've picked this up.

Let me check it up and merge. implementing the interface should be easy enough for me to tack on.

Will test and merge asap

lachlan-00:
happy with this as is for now. i'll extend Broadcast to use the media interface after merge

Welcome as a new contributor to Ampache!

PR #4395, opened 2026-07-15
Add play history consolidation into object_count_summary

fufroma:
I had an issue on the Dogmazic instance: the object_count table size was... too much. Really too much for the server.

We could trash the oldest counts, but we don't want to. Stats are important for our community. A solution would have been to create a consolidated table — not really complicated, but it takes some time.

For ze lolz, I asked Claude if it could write a patch... and... well, it did, and frankly it looks clean. x

Here's a PR with this consolidated table option. Feel free to trash it, but I think it can be quite useful.

PS : I didn't run it on production server, I want to stay mainstream to Ampache. If you don't want this kind of feature, or code made a bot, I'll look for another solution

------

On long-running servers the object_count table can grow to several GB. This adds an opt-in consolidation mechanism:

* New object_count_summary table storing aggregated counts per (object_type, object_id, user, count_type) with the covered period
* New stats_consolidate_threshold config option: number of days of detailed play history to keep (0/unset = disabled). config_version bumped to 88
* New CLI command cleanup:consolidateStats [--count-type X] [--execute] (dry-run by default), running INSERT+DELETE inside a transaction so an interruption cannot double count rows
* All count rebuild queries (Stats::clear, Catalog::update_counts, Album/Artist::update_table_counts, Video::update_video_counts and per-object variants) combine both tables so consolidated totals survive maintenance
* All bounded all-time readers extended to include the summary: Stats::get_object_count, User::get_play_size (streamed data size), the RatingMatch plugin play/skip counts
* Hygiene: user garbage collection, playlist deletion, Stats garbage_collection/migrate also handle the summary table
* Database migration 794005; fresh-install schema updated (table + db_version bump so the migration is not replayed on new installs)
* systemd example units for monthly scheduled consolidation

Known limitation (documented in the config option and consolidate()): readers that inspect individual plays only evaluate the retained window after consolidation: period-based statistics (trending, recent, graphs, Last.fm export - windowed by design), all-time top charts (Stats::get_top_sql with a 0 threshold, including the cron cache), smart playlist play-history rules and play count sorting. VALUES() in ON DUPLICATE KEY UPDATE is deprecated on MySQL 8 but the alias replacement is unsupported on MariaDB; VALUES() is kept as the only portable form (consistent with Playlist::update_map).


lachlan-00:
could you update this one for develop8 branch? i'm going to move to that branch after the next release?

also get claude to consider the cache_object_count table as it is semi-used like that and might need some updating.

It was originall used for caching counts and totals for recent history but might need work

fufroma:
Ok, I close this PR and prepare a new one to the good branch with your indication. Thx :)

PR #4402, opened 2026-07-18
Add play history consolidation into object_count_summary

fufroma:
* First exchange about that was in https://github.com/ampache/ampache/pull/4395

Add play history consolidation, archival and restore for object_count (develop8)

On long-running servers the object_count table can grow to several GB.
This adds an opt-in consolidation mechanism with a lossless, reversible
archive:

Consolidation
* New object_count_summary table storing aggregated counts per
  (object_type, object_id, user, count_type) with the covered period
* New stats_consolidate_threshold config option: number of days of
  detailed play history to keep (0/unset = disabledQ). config_version
  bumped to 90
* New CLI command cleanup:consolidateStats [--count-type X] [--execute]
  (dry-run by default). It archives, aggregates then purges inside a
  single transaction so an interruption cannot double count rows
* All count rebuild queries (Stats::clear, Catalog::update_counts,
  Album/Artist::update_table_counts, Video::update_video_counts and
  per-object variants) combine both tables so consolidated totals
  survive maintenance
* All bounded all-time readers extended to include the summary:
  Stats::get_object_count, User::get_play_size, RatingMatch plugin
* cache_object_count: ObjectCache::compute() now merges consolidated
  counts into the threshold 0 (all-time) cache entries, mirroring the
  get_top_sql aggregation shapes (plain types, podcast rollup from
  episodes, album_disk rollup from songs), so cached all-time counts
  and the cron_cache path of get_object_count stay exact
* Fix pre-existing bug in Stats::get_top_sql: the album_disk branch
  hardcoded count_type 'download' in its object_count join, so stream
  and skip album_disk cache entries were computed from download rows

Data resilience & restore
* New object_count_archive table (cold storage, single lookup index,
  ROW_FORMAT=COMPRESSED on InnoDB) retaining the detail rows removed by
  consolidation. Every non-derived (media) detail row is copied here
  before it is deleted, in the same transaction as the consolidation
* New CLI command cleanup:restoreStats [--execute] (dry-run by default).
  It re-inserts the archived detail, rebuilds the derived rows, subtracts
  the restored counts from the summary, drops summary rows that are fully
  accounted for and empties the archive - all in one transaction, then
  runs Catalog::update_counts. INSERT IGNORE throughout, so a repeated
  restore is a no-op rather than a duplicate
* Derived types (album, album_disk, artist, podcast) are not archived;
  they are rebuilt at restore time from the archived media via joins, so
  the parent resolved is the one that is true now (merges / re-tagging
  are respected) and a restore also repairs parent rows missing for any
  other reason
* Reconstruction mirrors Stats::skip_last_play exactly: a skip deletes
  the album/artist/podcast rows but leaves album_disk, so album/artist/
  podcast are rebuilt from 'stream' media only and album_disk from
  'stream'+'skip'. This keeps restore from resurrecting parent plays a
  skip deliberately removed, so a consolidate -> restore cycle is exact
  even for skipped history

Index & hygiene
* Drop redundant object_count indexes that were exact duplicates or
  leading-column prefixes of existing keys, lowering write overhead and
  table size
* User garbage collection, playlist deletion, object_id remap and Stats
  garbage_collection / migrate / clear also handle the summary and
  archive tables
Migrations
* 800013 - object_count_summary
* 800014 - drop the redundant object_count indexes
* 800015 - object_count_archive
  (fresh-install schema includes the tables; db_version left at 794004
  following the develop8 convention, the migrations are idempotent)
* systemd example units for monthly scheduled consolidation

Known limitation (documented in the config option and consolidate()):
readers that inspect individual plays only evaluate the retained
window: period-based statistics (windowed by design), live all-time
top charts computed without cron_cache, smart playlist play-history
rules and play count sorting.
VALUES() in ON DUPLICATE KEY UPDATE is deprecated on MySQL 8 but the
alias replacement is unsupported on MariaDB; VALUES() is kept as the
only portable form (consistent with Playlist::update_map).


lachlan-00:
sorry i was moving things around and updated the db to 13

lachlan-00:
I am working on updating this to archive as well as restore to original state, we want to make sure that you can change back to the old way safely. (Commet here a'la AI as well to make sure it reads how you want it)

## Updated: what actually happens when you enable this

Reworked so that consolidation **cannot lose data**. Enabling it no longer
discards anything — it relocates rows.

### On consolidate

Play rows older than `stats_consolidate_threshold` days are aggregated into
`object_count_summary`, and the detail rows are **moved to a new
`object_count_archive` table** rather than deleted. The archive keeps every
column — exact `date`, `agent`, `geo_*` — for streams, skips *and* downloads.

Parent rows (`album`, `album_disk`, `artist`, `podcast`) are deliberately
**not** archived. They carry no unique information: the `set_played` fan-out
copies them from the media row that triggered them. Storing them would double
the archive for zero gain.

### On restore

`cleanup:restoreStats` moves the archive back into `object_count` and
**regenerates** the parent rows from the restored media rows, resolving through
`song.album`, `album_disk`, `artist_map` and `podcast_episode.podcast`.

Regenerating is deliberately *more* correct than archiving those rows would be.
A stored parent id can be moved by an artist merge (`Stats::migrate`) or a
re-tag, so restoring it verbatim can point at the wrong parent or at nothing.
Resolving at restore time always yields the parent that is true now. The
regeneration reads `object_count`, so a restore also repairs parent rows that
went missing for unrelated reasons.

### Space saving

`object_count` carried **13 indexes**, two of them byte-identical
(`object_count_full_index` duplicated `object_count_UNIQUE_IDX`, the widest key
on the table) and three more redundant prefixes. Those are dropped in `800014`
— no data loss, no query-plan change. The archive is then defined with one
index and `ROW_FORMAT=COMPRESSED`, so moving a row out of a 13-index table into
a 1-index compressed table is where most of the saving actually comes from,
while keeping every byte of the data.

### Bugs found and fixed while doing this

- `skip_last_play` filters on `object_type`, so a skip flips only the media row
  and its `album`/`artist` rows stay `'stream'`. Reconstruction keyed on the
  song row being `'stream'` would have dropped the parent rows for **every
  skipped track**. Now matches `IN ('stream','skip')`.
- The archive was invisible to `Stats::clear()`, `garbage_collection()`,
  `Stats::migrate()`, `Playlist::delete()` and the user-deletion table list, so
  a restore could resurrect deliberately-deleted history or leave rows on a
  merged-away id. All five now maintain it.
- `TRUNCATE` inside the restore transaction would have implicitly committed and
  defeated the rollback — now a `DELETE`.
- Pre-existing PHPStan failures in `consolidate()` (dead `$dbh !== null` guards)
  fixed, so `composer stan` passes.

### Verification

Round-trip tested against a seeded database covering skips with stream parents,
downloads (no fan-out), NULL agents, songs with no album, episodes with no
podcast, non-derived types, and post-cutoff rows. Four incremental
consolidations followed by one restore: **28 → 28 rows, nothing lost, nothing
invented** on a full-column null-safe diff, and a second restore is a no-op.

### Still worth a look

`stats_consolidate_threshold` defaults to `0` (disabled) and the CLI is dry-run
unless `-e` is passed, but the example systemd unit ships with `-e` in
`ExecStart` — copying it verbatim gives a live run on first timer fire. That's
now called out in the unit file's comments.

fufroma:
Asked Opus 4.8 to take a look

Testing the restore path surfaced a blocking bug: `cleanup:restoreStats` always failed with `SQLSTATE[22003] BIGINT UNSIGNED value is out of range` and rolled back, so nothing could ever be restored.

The `subtract` step does `SET count = GREATEST(0, count - restored)`, but `object_count_summary.count` is `INT UNSIGNED`. MySQL evaluates `count - restored` in unsigned arithmetic **before** `GREATEST` runs, so as soon as a restored group has more rows than the summary currently holds (normal after several partial consolidations) the subtraction underflows and the statement aborts. Fixed by casting both operands to `SIGNED` before subtracting:

​```sql
GREATEST(0, CAST(`count` AS SIGNED) - CAST(`restored` AS SIGNED))
​```

I also added the `song.album > 0` guard to the `album_disk` branch of `derivedSources()`, matching the guard the `album` branch already has, so songs with no album don't resolve to a bogus parent during regeneration.

Verified with a full round-trip (four incremental consolidations then a restore) on MySQL 8.0: restore now completes, detail rows come back with no loss or duplication, a second restore is a no-op, and play counters are identical before and after.

Patch to apply on fc857eb

[fix-restore-stats.diff.gz](https://github.com/user-attachments/files/30181319/fix-restore-stats.diff.gz)

lachlan-00:
i'll have another parse over this today but this should be fully reversable and should also increase the performance of the table by getting rid of those keys. i'll need to test it more fully though on a copy of my db first and see how the restore goes

PR #4408, opened 2026-07-22
Fix CVE-2026-13149

fufroma:
brace-expansion: DoS via exponential-time expansion of consecutive non-expanding {} groups

https://github.com/advisories/GHSA-3jxr-9vmj-r5cp


lachlan-00:
had a look it shouldn't affect an install but not biggie to do

PR #4410, opened 2026-07-22
Better played count with enabled cache + add cache to Catalog::get_update_info()

fufroma:
Cf #2587 there was an issue to count played song when cache is enabled.
This PR is a solid way to prevent that, the idea is from https://github.com/ampache/ampache/pull/2589
    
I also added cache to Catalog::get_update_info(). There is no use to request SQL for the same value multiple time, let's cache it.

It you accept this PR you can close https://github.com/ampache/ampache/pull/2589


fufroma:
I checked that the SQL got the right index to stay fast

You can refuse this PR with the argument https://github.com/ampache/ampache/pull/2589#issuecomment-715639740 that is still valid, but the part about the cache of `Catalog::get_update_info()` may still be interesting.

fufroma:
When in doubt, let the user decide: here's an option to turn it on or off.

(again, feel free to trash this PR if you don't like the way it go)

PR #4415, opened 2026-07-23
jPlayer use old img, switch to svg 

fufroma:
On my screen, the player icons looked blurry.

Using bootstrap, here an attempt to have icons in high-resolution.

Before : 
player-old

After :  
player-new


lachlan-00:
wow they look nice

PR #4417, opened 2026-07-24
Remove all bitmap

fufroma:
Everything is now SVG and CSS, no more bitmap assets.

@lachlan-00 I'm not as confident as I was with my PR for the audio player, as there are parts of the video section I wasn't able to fully test.


lachlan-00:
that's okay i should be able to find out any remaining uses. it's been a good process cleaning jplayer up. i think a lot of these might not be used already as they're in the action buttons but i'll check manually

lachlan-00:
I can't see anything referencing these so looks safe to remove them

Issue #1373, opened 2016-08-28
[Feature] Bulk actions / multi select for web app


roba91:
Using the web front-end would be much easier with bulk actions.

Just on example, editing a playlist. I accidentally added a whole artist to that playlist. It is really annoying to click the red x for every track (especially because the web page is being scrolled all the way up automatically).

Better would be: Check boxes or multi select with `Shift+Click` and a select tag or a context menu (right click or so) where I can delete all songs together from the playlist.

Phyks:
I'm pretty sure this is a duplicate, but cannot find the other associated issue.

Anyways, :+1: for this feature.

dotanm:
I've opened #4418 with multi-select on the playlist items list.

To be clear about why: not because this issue is old, but because I think multi-select would help on a lot of Ampache's screens, and the playlist item list seemed like the smallest and least intrusive place to show what it could look like. So please read it as a first implementation and a starting point, not a finished design — if you'd rather the interaction or the placement worked differently, I'd like to hear it before I take the same pattern anywhere else.

Rows get a checkbox, `Ctrl`/`Cmd`+click toggles a row and `Shift`+click selects a range anywhere on the row. A bar shows how many rows are selected along with the actions that apply to them: play, play next, play last, add to the temporary playlist, add to another playlist, remove from the playlist. Removing a selection is one request and one track renumber instead of one of each per track — which is also what stops the list jumping back to the top.

On the technical side it adds no new dependency: no npm or composer package, no config option, no database change. The selection logic is one new module in the existing Vite bundle using the jQuery that's already loaded, and it drives the actions through the same delegated `data-*` click handling `Ajax::button` already uses, so it keeps working after an AJAX browse refresh. The server side mostly needed nothing — `directplay` and `append_item` already accept a comma-separated list of ids; only `delete_track` and the basket action were extended to match. The action bar is its own template, so adding another screen is a checkbox cell and a list of actions.

PR #4418, opened 2026-07-26
Add multi-select with batch actions to the playlist items list

dotanm:
Addresses #1373.

Removing several tracks from a playlist meant clicking the red x once per track, and each click deleted one row, called `regenerate_track_numbers()`, and re-rendered the whole browse — which is what threw the page back to the top, the specific complaint in that issue.

This is meant as a first implementation of multi-select rather than a finished design, and the playlist items list is deliberately the smallest place to try it. If the interaction or the placement should work differently I would rather hear it before the same pattern goes anywhere else.

## What it does

* Rows carry a checkbox, and the header checkbox selects the whole page
* `Ctrl`/`Cmd`+click toggles a row and `Shift`+click selects a range, anywhere on the row that is not itself a button or a link. `Cmd` is honoured because on macOS `Ctrl`+click is a right-click
* A bar shows how many rows are selected with the actions that apply to them: play, play next, play last, add to the temporary playlist, add to another playlist, remove from the playlist
* Removing a selection is one request and one renumber instead of one of each per track
* The bar is `position: sticky` on desktop so it stays in view while scrolling a long list

## No new dependency

No npm or composer package, no config option, no database change. `package.json`, `composer.json`, `ampache.cfg.php.dist`, `CONFIG_VERSION`, the migrations and `resources/sql/` are all untouched.

## How it fits what is already there

* The selection logic is one new module in the existing Vite bundle (`src/js/multiselect.js`) using the jQuery already loaded. Every handler is delegated from `document` and the actions are driven by `data-*` URL templates, matching the delegated `data-ajax` handling `Ajax::button` already uses, so a browse replaced by an AJAX refresh keeps working with no re-init
* Most of the server side needed nothing: `PlayItemAction`/`StreamItemAction` already `explode(',', object_id)` and `append_item` already explodes `item_id`. Only `delete_track` (now a track id list, with a single `regenerate_track_numbers()` at the end) and the basket action (now an id list) were extended
* The action bar is its own template (`show_multiselect_actions.inc.php`) rendered from a `$multiselect_actions` array, so a second screen is a checkbox cell plus a list of actions
* `cel_select` reuses the cell class `show_pvmsgs.inc.php` already had, and every icon already exists under `resources/images/material-symbols`
* Only a real `Playlist` gets the remove action; a `Search` shares `has_collaborate()` but has no `delete_track`
* On mobile the bar stays static: `#header` already owns `top: 0` there and is 120px of auto height, so a second sticky strip would either hide behind it or need an offset that breaks whenever the header rewraps

## Also fixed

`show_disabled_songs.inc.php` has been calling `check_select('song')`, a function defined nowhere in the repository (`release7` included), so that Select toggle silently did nothing. The new module defines it.

## Testing

`composer run-script syntax` and `cs:fix` clean, PHPStan level 8 clean on every changed file with no new baseline entries. The suite has 65 pre-existing errors in 4 classes (OIDC needs `jumbojett`, plus the Rest/Subsonic spec validators); running the same filter on an unmodified `develop8` gives the identical 65.

Driven in a real browser against the dev instance: single click, `Shift` range, `Ctrl`/`Cmd` row toggle, select-all with the indeterminate state, and the exclusions for links and the drag grip. Add to temporary playlist fires **one** request (`id=25,3,4`) and the rightbar really gains 3 items; remove fires **one** request (`track_id=28,29,30`) and the list goes 12 rows to 9 with the count reset. Batch play sends one `directplay` with the comma list — the two follow-up calls it triggers are identical for a single-row play. At 360px `scrollWidth == innerWidth`, so no horizontal overflow.


lachlan-00:
I'll update and merge when i can but this is going to make a lot of people happy i would say

lachlan-00:
I've made view optional through cookies and also made the space where the box would be permanent so it doesn't shift the content weirdly when you turn it on / off

PR #4420, opened 2026-07-30
Add multi-select with batch actions to the song lists

dotanm:
Continues #1373, following #4418.

#4418 said the playlist items list was deliberately the smallest place to try multi-select, and that the same pattern
should not go anywhere else until the interaction looked right. The song lists are the natural next screen: one
template, `show_songs.inc.php`, covers the songs browse as well as the song lists on the album, artist and search
pages, so a single change reaches most of the places a selection is actually useful.

This registers into the `Browse::MULTISELECT_TYPES` list and the `$can_multiselect`/`$show_multiselect` split added
after #4418 was merged, so the `Select` option and the reserved column behave exactly as they do on the other screens.

## What it does

* The same checkbox column, header select-all, `Ctrl`/`Cmd`+click toggle and `Shift`+click range as the playlist items list
* A bar with the actions that apply to a song selection: play, play next, play last, add to the temporary playlist, add to another playlist, download
* Download is the batch zip. A selection cannot stream as one file, so a zip is the only sane meaning for "download several songs"

## One config default changed

Unlike #4418, this one is not config-free: `song` has been added to `allow_zip_types` in `ampache.cfg.php.dist`.

`song` was never a documented value there, so without it `ZipHandler::isZipable('song')` is false and the download
action can never appear, no matter what the admin does. Nothing else moved — no npm or composer package, no new
preference, no database change, `CONFIG_VERSION` and `resources/sql/` untouched. Admins who do not want it can leave
`song` out of their own `allow_zip_types` and the action simply is not rendered.

## How it fits what is already there

* `batch.php` now takes a comma separated `id` list, so one zip can hold the medias of several objects instead of one. This is the same shape of change `delete_track` needed in #4418, and it leaves the single id form byte for byte identical
* The bar comes from the existing `show_multiselect_actions.inc.php` and a `$multiselect_actions` array, which is what #4418 built it for. No new template
* The checkbox cell is echoed from `show_songs.inc.php` rather than added to `resources/templates/song_row.xhtml`. The cell only ever has to lead the row, so the PHPTal row template stays untouched
* The download action needed one new action kind in `multiselect.js`: `link` navigates instead of firing `ajaxPut`, because a zip arrives as one response rather than an AJAX side effect
* A zip carries a single type, so `link` is disabled on a selection spanning several types. The add-to-playlist dialog is untouched — it handles several types itself now
* The download link is gated on `Access::check_function(FUNCTION_BATCH_DOWNLOAD)` and `isZipable('song')`, matching how the album and artist pages already gate their own zip links

## Not fixed here

The `tfoot` in `show_songs.inc.php` omits the `cel_year` `<th>` that `thead` has, so the footer row has been one cell
short of the header for a while (11/11/10 with year shown). The new column was added to both, so the gap is unchanged
rather than made better or worse. Happy to fix it separately if you want it.

## Testing

`composer run-script syntax` clean, `cs:fix` reports 0 of 2553 files changed, PHPStan level 8 clean on every changed
file with no new baseline entries. The suite has 64 pre-existing errors in 4 classes (OIDC needs `jumbojett`, plus the
Rest/Subsonic spec validators); running the identical filter on an unmodified `develop8` gives the same 64.

Driven in a real browser against the dev instance, at Build 035 with the bundle rebuilt:

* With `Select` off: the column is laid out but empty, no checkboxes, no bar, and the other columns sit exactly where they do without the change
* Toggling `Select` on in the View menu: 39 checkboxes, select-all with the indeterminate state, `Shift` range, `Ctrl` row toggle, the selected-row highlight, and the bar hidden by `visibility` at zero
* Reloading with the cookie set restores the checkboxes, and toggling back off removes them and leaves the column reserved
* Add to temporary playlist fires **one** request (`type=song&id=28,29,32`)
* A zip of 3 songs contains 5 entries (the 3 files, the folder and the m3u) at their real sizes, named `<first song> and 2 more items.zip`
* `batch.php?action=album&id=1` returns exactly the same byte count before and after the change; an unknown id inside a list is skipped; an all-unknown list behaves the same as it already does for `action=album`
* At 360px `scrollWidth == innerWidth` in both toggle states, and the bar stays static

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -452,7 +452,7 @@ catalog_prefix_pattern = "The|An|A|Die|Das|Ein|Eine|Les|Le|La"
 ; This setting allows/disallows zip download of specific object types
 ; If empty, all supported object types can be zipped.
 ; Otherwise, only the given object list can be zipped.
-; POSSIBLE VALUES: artist, album, album_disk, playlist, search, tmp_playlist
+; POSSIBLE VALUES: artist, album, album_disk, playlist, search, song, tmp_playlist
 ; DEFAULT: none
 ;allow_zip_types = "album, album_disk"
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,9 @@ You can downgrade to Ampache7 if you try this out and have issues, using the cli
   * A selection can be played, played next, played last, added to the temporary playlist, added to another playlist, or removed from the playlist in a single action, from a bar that stays in view while you scroll the list
   * Removing a selection sends one request and renumbers the list once, instead of one request and one full renumber for every entry
   * A mixed collection groups the selection by type before acting on it, so playing a selection of albums and songs together works from one bar
+  * The same checkboxes, gestures and action bar on the song lists, which covers the songs browse as well as the song lists on album, artist and search pages
+  * A selection of songs can be downloaded as a single zip, when `allow_zip_types` includes `song` and batch download is permitted
+  * `batch.php` accepts a comma separated `id` list, so one zip can hold the medias of several items instead of only one
 * A "Create Playlist" button on the playlists browse. Until now a playlist could only come into existence as a side effect of adding something to one
 * Database
   * New `api_enable_8` preference to enable/disable API v8 responses per user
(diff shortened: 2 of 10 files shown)


lachlan-00:
Fyi develop8 has been renamed and will be deleted.

develop is now the main dev branch again

PR #4421, opened 2026-07-31
Reuse TAL view across browse rows instead of new PHPTAL per row

fufroma:
Move `createTalView()` and `setTemplate()` out of the row loop and reassign contexts each iteration.

This is allowed by PHPTAL https://phptal.org/manual/en/split/execute.html

Building the PHPTAL engine once instead of per row skips redundant object instantiations + config/realpath calls.


lachlan-00:
I haven't done any work on templates as i am not sure about how deep we go into tal but that's a good thing to do.

develop8 merging today!

PR #4427, opened 2026-08-02
Answer a status when a stream cannot be opened

dotanm:
A stream that cannot be opened is the one failure in `PlayAction::run()` that does not set an HTTP status. It logs the
problem and returns, so the framework answers **200 with an empty `text/html` body**, and the client reports something
misleading — Firefox says *HTTP "Content-Type" of "text/html" is not supported* and Chrome
`DEMUXER_ERROR_COULD_NOT_OPEN`. Nothing tells the user, or the person reading their bug report, that the server failed.

## What it does

Two lines, in the two places that made the failure quiet.

* `PlayAction.php:1170` — the `!is_resource($filepointer)` branch sets `header('HTTP/1.1 500 Stream Failed')` before it
  returns. That method already sets an explicit status for eighteen other failure conditions (`400 Nothing To Play`,
  `403 Session Expired`, `404 File unreadable`, and so on); this branch was the omission. `header('HTTP/1.1 NNN <reason>')`
  is what the rest of the method uses, so it follows that rather than introducing `http_response_code()`
* `Stream.php:700` — the missing `transcode_cmd` message moves from level 5 to level 2. Its sibling misconfiguration
  twenty-six lines down (`Target format ... is not properly configured`, `Stream.php:726`) is already level 2 and also
  `return []`s, and the *symptom* of this one is logged at level 2 further along (`Stream.php:918`, `Transcode requested,
  but get_transcode_settings failed`). So the cause was the only part of the chain sitting at the most verbose level

## How you reach it

The usual way is a transcode with no command configured, which the shipped config makes easy to hit:
`ampache.cfg.php.dist` has `transcode_mp3 = "allowed"` uncommented while `transcode_cmd` stays commented, so any mp3
above the user's `transcode_bitrate` asks for a transcode that cannot run. `.ogg` is unaffected because its
`transcode_*` lines stay commented and it streams raw, which is part of why this looks intermittent and file-specific.

I am not proposing a change to those defaults. `transcode_mp3` is uncommented deliberately — mp3 waveforms need a
transcode to wav (`Waveform.php` gates on it), which is exactly why it was enabled in the first place. And falling back
to a raw stream would defeat the `transcode_bitrate` cap, which is the point of the cap. Whatever the config says, a
stream that cannot be opened should not answer 200.

## What this does and does not do

It does not fix anyone's playback. It turns a silent 200 into a real error, so a client can report a server failure and
a log reader can find the cause. #2819 is the class of report I have in mind — "audio sounds warbled" style transcode
complaints where the user has no signal to work from — though this only makes such a case diagnosable, it does not
diagnose it.

I could not find an existing report of the empty-200 specifically, but GitHub search is fuzzy enough that I would not
claim it is unreported.

## Testing

`php -l` clean on both files, PHPStan level 8 clean with no new baseline entries, `php-cs-fixer` reports 0 of 2 changed
files needing fixes.

Driven against a running instance with a 320kbps mp3, requesting `bitrate=128000` to force the transcode, and toggling
the two hunks in and out of the working tree so each row is a real before/after on the same build:

| `transcode_cmd` | request | before | after |
|---|---|---|---|
| commented out (as shipped) | `bitrate=128000` | **200**, `text/html`, 0 bytes | **500 Stream Failed** |
| `ffmpeg` | `bitrate=128000` | 200, chunked, transcoded to 3616863 bytes | unchanged |
| `ffmpeg` | none | 200, `audio/mpeg`, `Content-Length` 9056216 = file size | unchanged |

So the only behaviour that changes is the one that was already broken; both working paths are untouched.

For the log level, with `debug = "true"` and `debug_level = 3` — a plausible production setting — the same failing
request logs the cause zero times before the change and does log it after, while `Transcode requested, but
get_transcode_settings failed` was visible in both. That is the asymmetry the second hunk removes.


fufroma:
Could you ask to be less verbose? You added a http header and changed the loglevel of a message, those are straight forward changes.

lachlan-00:
run over it with these skills. i've spent a bit of time trying to make it fit the styles properly so this should help

[claude.zip](https://github.com/user-attachments/files/30639658/claude.zip)

lachlan-00:
i'll review when i'm done with mpd but those skill will give you a big leg up in making any changes

dotanm:
> Could you ask to be less verbose? You added a http header and changed the loglevel of a message, those are straight forward changes.

@fufroma , you're completely right, and I apologize. I'll take note of that in the future.

dotanm:
> run over it with these skills. i've spent a bit of time trying to make it fit the styles properly so this should help
> 
> [claude.zip](https://github.com/user-attachments/files/30639658/claude.zip)

Thank you, @lachlan-00 - I installed it and will follow it from here.

PR #4428, opened 2026-08-02
Add search user in admin browser-user page

fufroma:
Make it easier to find an user when you have thousand of them :)


lachlan-00:
do you use the filter box? 
image

PR #4433, opened 2026-08-08
Found some function that may be deprecated.

fufroma:
Found some function that may be deprecated. So, added a deprecated warning on it


lachlan-00:
can't see them

PR #4436, opened 2026-08-09
Add connect/read timeouts to UPnP device cURL requests

bunlongheng:
The two active cURL requests in UPnPDevice (sendRequestToDevice and parseDescriptionUrl) never set CURLOPT_TIMEOUT or CURLOPT_CONNECTTIMEOUT, so a UPnP renderer that stops responding mid-transfer hangs the request until the server's own max_execution_time kicks in, tying up a PHP worker. Since the host comes from device discovery and the stored description URL, an unreachable or misbehaving renderer stalls playback control. I added a 5s connect timeout and a 30s overall timeout to both handles, matching the timeouts already used in PlaylistUrlResolver and SubsonicClient. No behavior change for devices that respond normally.


lachlan-00:
Hello new Ampache contributor and thank you!

PR #4440, opened 2026-08-11
Preference speed

fufroma:
Drop redundant user_preference indexes

Stop scanning user_preference for the sharing lists


lachlan-00:
Releasing tomorrow so get any quick wins you want in over the next 12 hours and I'll start when I wake up tomorrow

PR #4442, opened 2026-08-11
Serialize artist/album creation, parallel uploads created duplicates

### LINKED ISSUE #4441: Parallel uploads create duplicate artists (race condition)
_Tested on develop branch b90bf43b08c9ee87b1f3faeeb7190ab9698a3794_

Uploading several files of the same (new) artist creates duplicate artist rows.
 
FilePond uploads files in parallel by default, so two concurrent POST requests each run `Artist::check`, both name lookups miss (the other insert is not committed/visible yet), and both insert. There is no UNIQUE constraint on `artist` to catch it.
 
## To reproduce
 
1. Upload 2+ files tagged with the same artist name, not yet in the database (no artist/album selected in the form, parallel upload).
2. Two identical artist rows are created (same name, empty mbid).
Uploading the same files sequentially creates a single artist.
 
Verified on develop `b90bf43b0`.
 
## Notes
 
- `Album::check` likely has the same race.
- Possible fixes: `maxParallelUploads: 1` in the FilePond options (workaround), or serialize lookup+insert in `Artist::check` with `GET_LOCK`.

fufroma:
Fixes #4441
 
`Artist::check` and `Album::check` looked up then inserted with no serialization: two concurrent requests (parallel uploads) both missed the lookup and both inserted. `artist` and `album` carry no UNIQUE constraint to catch it (multiple NULL mbids are allowed).
 
- New `DbLock`: named MySQL lock (`GET_LOCK`/`RELEASE_LOCK`), key prefixed with the database name, 10s timeout.
- `Artist::check` / `Album::check`: fast lookup unchanged; on miss, acquire the lock, look up again (finds the row a parallel request just created), insert otherwise, release. Readonly path never locks. If the lock cannot be acquired, insert anyway and log.
- The re-lookup also resolves the artist through the mbid back-filled by the first pass.

Tested: two parallel uploads of a new artist created 2 rows before, 1 row after (same instance). Other creators on the upload path (`album_disk`, `tag`) are already covered by UNIQUE constraints.

### DIFF
--- a/src/Module/System/DbLock.php
+++ b/src/Module/System/DbLock.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * vim:set softtabstop=4 shiftwidth=4 expandtab:
+ *
+ * LICENSE: GNU Affero General Public License, version 3 (AGPL-3.0-or-later)
+ * Copyright Ampache.org, 2001-2026
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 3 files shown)


lachlan-00:
sorry just set this with 6570764469ee6feb5fc4713da675ffe2c0eff35a

locking with tags as well considered. build and release time!

PR #4447, opened 2026-08-13
Podcast here we go

fufroma:
Podcast :
- `itunes:image` always emitted, 1400x1400 (`upscale_images` gated), falls back to the placeholder
- Art fallback urls serve the full size image instead of a 128px one
- Cover art uses the item's own type (was hardcoded to `album`)
- `description` / `itunes:summary` always emitted, generated when the item has none
- `itunes:author` added
- `itunes:explicit` added, set with `rss_explicit`
- `language` uses the two-letter code (`fr`, not `fr-FR`)
- Optional Apple sub-category, set with `rss_subcategory`

Fix suffix range and attachements for streams 
- `bytes=0-` and `bytes=0-0` answer 206 instead of 200
- Suffix ranges (`bytes=-500`) resolved against the file size
- Streams sent as `inline`, only explicit downloads stay attachments
- Play counts, transcoding and 416 handling unchanged

### DIFF
--- a/config/ampache.cfg.php.dist
+++ b/config/ampache.cfg.php.dist
@@ -983,6 +983,16 @@ show_footer_statistics = "true"
 ; DEFAULT: "false"
 ;use_rss = "true"
 
+; Set the itunes:explicit flag sent in rss feeds. Podcast directories
+; require the tag, Ampache has no per-item rating to fill it from.
+; DEFAULT: "false"
+;rss_explicit = "true"
+
+; Apple sub-category advertised by podcast feeds, on top of the "Music"
+; category. Valid values: Music Commentary, Music History, Music Interviews
+; DEFAULT: ""
+;rss_subcategory = "Music Commentary"
+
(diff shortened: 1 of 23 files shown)


lachlan-00:
where does an explicit value come from? i see the itunes advisory tag but there doesn't seem to be a common answer for it. I could set it as a flag in the song_edit windows or something but that would then be manual

fufroma:
I used

- https://www.castfeedvalidator.com — Cast Feed Validator (Blubrry)
- https://podba.se/validate/ — « How's my feed? » (podbase)
- https://podcastai.com/fr/validateur-de-flux

And the explicit was listes as a l'équipement. I just set up a default value, but allowing people to set it by song or by album and allowing people to hide explicit song would be nice.

fufroma:
- https://help.apple.com/itc/podcasts_connect/#/itcb54353390 `<itunes:explicit>` is in Required list
- https://podcasters.apple.com/support/829-validate-your-podcast list missing explicit tag as error

PR #4455, opened 2026-08-15
Recently played: stop repeating the same play

fufroma:


fufroma:
Prevent duplicate in /rss.php?type=recently_played

PR #4457, opened 2026-08-17
Replace fopen/fread/fwrite loops that read or write an entire file at once

fufroma:
- Art::_write_to_dir() checked the target directory with is_readable(); use is_dir(), which is what the log message already claimed.
- Art::_write_to_dir() returned true even when the write failed. insert() reads that return value to null the image column, so a failed disk write could drop the blob from both disk and database. It now returns false and logs.
- InstallationHelper::write_config() returned true unconditionally; it now returns the write result.
- Log unreadable art files instead of falling through silently.

Streaming reads (playback, WebDAV, uploads, WAV headers, 32-bit filesize fallback) are left alone.

### DIFF
--- a/src/Module/Art/Art.php
+++ b/src/Module/Art/Art.php
@@ -663,15 +663,12 @@ class Art extends database_object
 
         // Check to see if it's a FILE
         if (isset($data['file'])) {
-            $handle = fopen($data['file'], 'rb');
-            $size   = Core::get_filesize($data['file']);
-            if (
-                $handle
-                && $size > 0
-            ) {
-                $image_data = (string) fread($handle, $size);
-                fclose($handle);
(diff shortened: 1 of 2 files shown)


lachlan-00:
i see you've already updated, album dashboard loading in  0.1856 seconds!

fufroma:
> i see you've already updated, album dashboard loading in 0.1856 seconds!

Yes, 8.0.1 was a happy update because we had lot of local patchs merged upstream! I had to rework a bit the [PoW](#4444) feature but everything went well.

The most impressive perf gain was unexpectedly from the PR [#4456] . I didn't thought the disk cold cache was such an issue, so now I'm looking everywhere around the disk ;)

PR #4465, opened 2026-08-19
Features, fixes and perf

fufroma:
Summary:

_Measured on a copy of the Dogmazic database: 62k songs, 8k albums, 5k artists, 3k playlists, 8k accounts.
Same box, PHP and DB. Only the code and the fun changes._

Not everything is on Dogmazic, but the most significant feature and fix already are. We saw drastic perf increase and server load improvement with those patch.

### Art

Drawn art instead of a blank placeholder for items with no cover.

Preview: [Artist](https://play.dogmazic.net/artists.php?action=show&artist=7102) / [Album](https://play.dogmazic.net/albums.php?action=show_disk&album_disk=2718) / [Playlist](https://play.dogmazic.net/playlist.php?action=show&playlist_id=3777)

artist  album playlist2

settings

It's cheap on both ends:

- nothing is stored => no file on disk, no db, no cache directory. The drawing is dynamically built
- 4645 bytes of SVG whatever size is asked for. Typical 200x200 jpeg on Dogmazic is approx. 10 KB and a
  400x400 approx 25 KB. Light, and stay sharp even in 4k
- one drawing per item for all size => so a grid showing the same album as a thumbnail and as a cover fetches it once.
- ETag + Cache-Control
- two templates (light/dark), four motifs (record, medallion, waveform, tracklist), and a title saying "no cover yet, drawn from XXX" so it wont be mistaken for real artwork
- off by default => and even if an admin enable it, an user can turn it off for himself

### Others features
- Dynamic page titles carrying the artist or album; the player still overwrites it while playing. Optional emoji per page type (user setting) => emoji was lighter than write "Album" or "Song" or "Podcast" in front of title.

- svg favicon, apple-touch-icon and a link preview image =>all customisable in a new Branding section. Let's share beautifully !

- test.php now warns when http compression is off.

## Perf

- API and mobile clients
  - app launch (getPlaylists + getArtists + getAlbumList2 + getRandomSongs + getAlbum): 2.63 s / 8972  queries => 0.40 s / 188 queries
  - ampache API: songs 478 => 54 queries, and query counts no longer grow with page size
- Web pages
  - lighter HTML
  - album page went from 641 queries to 51 ; playlist from 173 to 133
  - global speed improvement on all pages
  - play history search rules had no index: "last played more than N days" 7.8 s => 0.4 s, "played by
  me" on artists 25 s => 2 s
- no longer insert in database for people just browsing
  - no session rewrite per request
  - no `tmp_playlist` for visitors who never queue
  - no stream session row per page view
  - no unused `tmp_browse` from dashboards
  - no session opened to serve a cover
- admin and maintenance
  - preference rebuild after a db update repairs the install once instead of once per user, and the garbage
  collector only rewrites names that changed: 500 users to repair 248 s => 6s
  - server-wide user counters batched: went from 159638 to 19 queries per cron run
- API
  - `user.apikey` now indexed and the hashed-key fallback only reads users who have a key. API auth
  13.4 ms => 0.9 ms

## Fix

- advanced random never answered => Search > Random, submit, F5: the reload is a GET with no `$_POST`,
  so no limit, so the whole library gets queued one insert per song. 62005 queries => 172, 0.19 s
- recently played showed dates up to a year old => it regrouped `object_count`, whose rows are pruned
  as they age, instead of reading the `last_played` column 800029 added and maintains. 1629 albums of
  8079 had a stale date. 540 ms => 95 ms
- saving the server Interface preferences trashed `custom_favicon`, `custom_login_logo` and
  `custom_login_background` => This one is a 8.0 regression, reproduced on a clean develop.
- Migration700005 wrote the playlist id in place of the count; 810007 recomputes it for installs that
  already ran it
- a custom browse base ignored every filter, group and sort
- toggling a browse option emptied a browse that had been handed its ids
- grid view had lost its edit and delete buttons, and the confirm dialog ignored the theme
- boolean preference inputs were settled by a hardcoded name list, so `cron_cache_live_count`,  `broadcast_private` and `show_composer` rendered as text inputs
- a missing cover served the 1400x1400 placeholder whatever size was asked for
- the album page fell back to a blank cover instead of the album artist art <= @lachlan-00 you may want
  to discard this one 552f9e8760981b4d2d96bf15e8dc1d1453cab9bf
- `stats.php?action=show` served a blank page instead of saying access denied
- UPnP answered every browse with a fatal error on PHP 8.5
- Subsonic `getIndexes`/`getMusicDirectory` now browse the real folder tree, so folder based clients work
- the OPML export of podcast subscriptions ignored the caller's catalog filter
- a playlist's total duration left videos and podcast episodes uncounted
- uploading new art didn't replace the image already on the page
- the login redirect carried a literal `&amp;` in the url
- html validity
- a client with a wrong clock pinned itself to the top of recently played, forever (we got one on Dogmazic) => found two more doors: album disks and the play queue. Fixed, added unit test, and added 1 minute of slack for device not using ntp


lachlan-00:
I will be back in 2 weeks, there are some ui template issues in 

760e1a01cce1193f0a416bd5a958b3b4feacc3c1 and 7eb148bd6f08962a6503bf01754252cc93f67a67 that you should apply to the site, they don't close the box correctly on form pages and get art

fufroma:
Almost all those patch are on Dogmazic server. Didn't notice any side effect, performance are great.

People really like the dynamic art.

More work to do on the share part but it conflict with the templates, so I'll wait before working more on that part.

lachlan-00:
for https://github.com/ampache/ampache/commit/552f9e8760981b4d2d96bf15e8dc1d1453cab9bf which pages are already falling back to artist art? i would probably only revert to song art instead of going to artist that way

fufroma:
> for [552f9e8](https://github.com/ampache/ampache/commit/552f9e8760981b4d2d96bf15e8dc1d1453cab9bf) which pages are already falling back to artist art? i would probably only revert to song art instead of going to artist that way

The home "random Albums box" calls it with force = true, so it already falls back to the album artist's art when an album has no cover of its own. Normal album listings fall back too, but only when the album artist already has art.

Found this "bug" because when I clicked on a random item in the Home, the page got a différent art than the home. I first though it was an issue and it didn't open the right file,but it was this fallbacl

Album::display_art() is the one

lachlan-00:
I've taken off some of the extra preferences as the access level is the way to block users making changes

Custom arr also takes precedence over generated but I may reverse that. I'll keep going again tomorrow

fufroma:
I looked at [3d30e07](https://github.com/ampache/ampache/pull/4465/commits/3d30e0783af7bc807f77b29126de0a46e7f0624e) and you are right, I forgot to look other API version. So, two more commit

